PR #412, opened 2014-12-16
Factored out isBrowser check into runtime-util

jg20:


jpolitz:
Sorry that took so long.

PR #444, opened 2015-02-20
Fix typo in Tour docs

gus-massa:
I'm not sure if it is `@pyret{\n}` or `@pyret{\\n}`


jpolitz:
That works great. The Travis failure is unrelated, and I built and the rendering looks good. Will get pushed out soon. Thanks!

PR #445, opened 2015-02-21
Support for nested block comments

blerner:
These commits provide support for nesting `#| ... |#` style comments in Pyret, and improving the editor modes (CM and emacs) to handle indentation for them properly.  (While I was at it, because it was essentially the same technique, I also fixed up indentation for triple-quoted strings.)  The indentation policy is:

* By default, block comments indent their bodies 3 spaces, so that the text lines up after the leading `#|` and space (i.e. three characters).
* By default, a closing `|#` as the first thing on a line will outdent to match the opening `#|`; a closing `|# elsewhere on a line will not cause an outdent.  (This is consistent with the existing behavior of closing parentheses, braces, brackets, and semicolons.)
* If the content of a comment block is indented at all, i.e. to any column other than column 0, then the indenter will *not* re-indent the content.  So for example, merely commenting out a block of code will not cause it to be reindented.  (Minor caveat: commenting out top-level definitions and reindenting the whole file will cause the first lines of those definitions to indent, since they're not otherwise indented...  I don't think that's a big enough problem to try to fix, but if it's really an issue later, we can try to hack in some kind of support for it after this merges.)
* Similarly to comments, triple-quoted strings indent to match whatever column contains the first backtick of the string.  This seemingly-odd choice means that e.g. triple-quoted `doc:` strings will line up perfectly, when the triple-quotes are placed on their own lines.
* Similarly to comments, any customization of the indentation will be respected.


jpolitz:
Nice.  LGTM.

PR #448, opened 2015-02-24
add safe-for-space tail calls with a flag to disable/enable

jpolitz:
(includes options now attached to the last step of the compiler)

The old version of the compiler had this, and I've confirmed that this all works, but it doesn't actually live on master.  This PR makes safe-for-space tail calls the default, and adds a flag that can be used to turn them off.

### DIFF
--- a/src/arr/compiler/anf-loop-compiler.arr
+++ b/src/arr/compiler/anf-loop-compiler.arr
@@ -380,10 +380,11 @@ fun compile-fun-body(l :: Loc, step :: String, fun-name :: String, compiler, arg
         end
     end
   end
-  stack-attach-guard = if compiler.options.proper-tail-calls:
+  stack-attach-guard =
+    if compiler.options.proper-tail-calls:
       j-binop(rt-method("isCont", [list: j-id(e)]),
-                            j-and,
-                            j-parens(j-binop(j-id(step), j-neq, ret-label)))
+        j-and,
+        j-parens(j-binop(j-id(step), j-neq, ret-label)))
     else:
       rt-method("isCont", [list: j-id(e)])
     end
--- a/src/arr/compiler/anf-loop-compiler.arr
+++ b/src/arr/compiler/anf-loop-compiler.arr
@@ -57,6 +57,7 @@ j-try-catch = J.j-try-catch
 j-throw = J.j-throw
 j-expr = J.j-expr
 j-binop = J.j-binop
+j-and = J.j-and
 j-eq = J.j-eq
 j-neq = J.j-neq
 j-geq = J.j-geq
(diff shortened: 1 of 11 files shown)


jpolitz:
Or we could just run CPO with tail calls off by default.  I really want there to be a way to run safe-for-space Pyret programs

jpolitz:
(Meta:  Github really needs to move the "Close and comment" button to a different place than the "Cancel" button that appears in other comment boxes.)

blerner:
+1 for safe for space tail calls. Possibly +1 for turning that off on CPO,
at least by default...

blerner:
Another thought: safe-for-space tail calls can be opted into on a
per-function basis in our compilation model. We should definitely compile
the compiler and the standard libraries with them (especially list
functions like fold or iter...), even if we don't compile student programs
that way. (I view that as no different than eliding source locations
"internal to Pyret", which we already do.)

jpolitz:
Also, the best way to see if this causes a problem is probably to enable it and keep an eye out, or design a study.  I don't think we know one way or another at all what real impact this could have aside from improving performance.  Most stack traces in student programs I see on CPO are well under 50 frames, and students use the first and last few frames to debug (anecdotal, from sitting next to them and helping them code).  Those frames usually contain the site immediately around the error, and the test case the trace started at.

@shriram what does BSL have to say about tail call frames in error messages?

jpolitz:
I also just realized that virtually no student programs will have tail calls in them, because most student programs are annotated, and return annotations turn into an extra checking step that puts any apparent tail calls one step away from the end of the function.  The speed difference between putting `-> Number` at the end of tail-recursive sum becomes very much apparent in this case.  Worth thinking about if there's a way to do that better, because it means that annotated programs won't have safe-for-space behavior.

jpolitz:
Relevant: http://www.cs.princeton.edu/~mg19/papers/popl2015_space.pdf

blerner:
I'd spin this off into a separate issue.  You and I had brainstormed a 
bit about how some (most?) contracts could be checked via safeCall 
without needing a dedicated $step, but that deep contracts still needed 
one, and refinements or other calls into arbitrary Pyret code definitely 
did.  Maybe there's a way we can build better smarts into the annotation 
checker so that it doesn't compile return annotations into an annotated 
let-binding...

jpolitz:
Problem is that safeCall isn't safe-for-space.  Would need to use safeTail, but that won't get the right behavior.  I think we just need a better representation for return annotations, so multiple of the same check can be squashed together (e.g. you need one check to happen, but not N checks).  Mutual recursion will require some more thought, but answers are out there in the literature, we just have to do the work of applying them.

Issue added at https://github.com/brownplt/pyret-lang/issues/450

PR #451, opened 2015-02-24
Small improvements to SublimeText3 syntax highlighting

awan1:
Added highlighting for "cases" and multi-line comments.

Clarified readme, added instructions for Windows users.

This is a tiny tiny increment on the work that @rrshaban did.


rrshaban:
I intentionally removed cases from the top two because it shadows the more detailed cases handling on lines 38-47 -- I'd recommend moving those lines up top if you're going to add general "cases" highlighting

PR #472, opened 2015-03-23
Merge Benchmarking Tools into Main Repo

awstlaur:
Benchmarking tools are aptly located in tools/benchmark. The merge also makes minor additions to the .gitignore and to eval-lib.js.

The benchmark tools have node dependencies which are explicated in tools/benchmark/README.md.


tools/benchmark/tests-code.js
```diff
@@ -0,0 +1,606 @@
...
+  		return flag;
+  	}, 'failure',5000);
+
+  	runs(function(){
+  		expect(passed).toBe(false);
+  	});
+  });
+});
+
+//no longer used
```
jpolitz: Should this whole test block just be removed?

tools/benchmark/tests-code.js
```diff
@@ -0,0 +1,606 @@
...
+  	waitsFor(function(){
+  		return flag;
+  	}, 'failure',5000);
+
+  	runs(function(){
+  		expect(passed).toBe(true);
+  	});
+  });
+
+  xit('passes on a valid (long) program', function() {
```
jpolitz: Lots of `xit` with this same test name.  Should they just be removed?

tools/benchmark/tests-code.js
```diff
@@ -0,0 +1,606 @@
...
+
+describe('createSuite', function(){
+  it('creates an instance of Benchmark.Suite with the correct number of benchmarks', function(){
+    expect(b.test.testCreateSuite()).toBe(true);
+  });
+});
+
+describe('runBenchmarks', function(){
+  it('returns array of objects with correct field types', function() {
+    debugger;
```
jpolitz: Is this a stray debugger statement that should be removed?

tools/benchmark/tests-code.js
awstlaur: Yes it should. Will be gone in a subsequent push.

tools/benchmark/tests-code.js
awstlaur: These were skipped because the 'long' part was making them take a very long time, as one might imagine. I will go over them again and make a decision.

tools/benchmark/tests-code.js
awstlaur: Yes. I've since locally removed all the uncommented debugger statements I could find, which will be reflected in a subsequent push.

Issue #397, opened 2014-12-02
is-List is not imported


sorawee:
> is-List
    The name is-List is used but not defined at
    > lists.is-List
    <function>

I think that is-List should be available. `data Test: test(a);` creates type `Test`, function `test()`, `is-test()`, and `is-Test()`. List should provide the same thing.

Issue #376, opened 2014-11-17
empty-list-set, empty-tree-set, list-to-list-set, list-to-tree-set are not imported


sorawee:
For example

    > empty-list-set
    The name empty-list-set is used but not defined at ...
    > [list-set: ]
    [list-set: ]
    > import sets as S
    > S.empty-list-set
    [list-set: ]

PR #560, opened 2015-09-04
add some missing constants / functions for CS19, fixing #376, #397

sorawee:


jpolitz:
Thanks!  This will be pushed out to CPO soon.

PR #580, opened 2015-10-15
Factor out Particle handling from world.js and change it to work with our core shim.

sstrickl:
In order to simplify the student experience, we plan to have a configurable (through events) shim that can be flashed once to a Particle core, and then configuration of reading/writing pins and the like are done via Pyret.  These commits change the Particle handlers to talk with our shim, and also refactors the Particle-specific code out of world.js (since we end up adding a lot more Particle-specific exports to Pyret).

### DIFF
--- a/src/js/trove/particle.js
+++ b/src/js/trove/particle.js
@@ -1,0 +1,154 @@
+define(["js/runtime-util", "js/ffi-helpers", "trove/world", "trove/world-lib", "trove/particle-shim-structs"], function(util, ffiLib, world, worldLib, pShimStruct) {
+
+  return util.definePyretModule(
+    "particle",
+    [],
+    {
+      values: {
+      },
+      aliases: {
+      },
+      datatypes: {
+      }
+    },
+    function(runtime, namespace) {
+      return runtime.loadJSModules(namespace, [worldLib, ffiLib], function(rawJsworld, ffi) {
+        return runtime.loadModulesNew(namespace, [world, pShimStruct], function(pWorld, pStruct) {
+          var WorldConfigOption = runtime.getField(pWorld, "internal").WorldConfigOption;
+          var adaptWorldFunction = runtime.getField(pWorld, "internal").adaptWorldFunction;
+
+          var sd_to_js = function(sd) {
+            var ret = {};
+            var arr = ffi.toArray(runtime.getField(sd, "keys-list").app());
+            for(var i in arr) {
+              var k = arr[i];
+              var v = runtime.getField(sd, "get-value").app(k);
+              if(typeof(v) === "string") {
+                ret[k] = v;
+              } else {
+                throw new Error('unimplemented value conversion for StringDict');
+              }
+            }
+            return ret;
+          };
+
+          var OnParticle = function(handler, name, sdict) {
+            WorldConfigOption.call(this, 'on-particle');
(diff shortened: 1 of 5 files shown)


src/js/trove/world.js
```diff
@@ -504,75 +460,70 @@ define(["js/runtime-util", "trove/image-lib", "trove/world-lib", "js/ffi-helpers
...
+          "provide-plus-types": makeObject({
+            "values": makeObject({
+              "big-bang": makeFunction(function(init, handlers) {
+                ffi.checkArity(2, arguments, "big-bang");
+                runtime.checkList(handlers);
+                var arr = ffi.toArray(handlers);
+                var initialWorldValue = init;
+                arr.map(function(h) { checkHandler(h); });
+                bigBang(initialWorldValue, arr);
+                ffi.throwMessageException("Internal error in bigBang: stack not properly paused and stored.");
```
jpolitz: This is a fantastic idea.  I wonder if there's an easy way to wrap it in a helper/macro and make it a default thing that should be done for tail-position-pausing calls.

src/js/trove/particle.js
```diff
@@ -0,0 +1,154 @@
...
+            var worldFunction = adaptWorldFunction(that.handler);
+            var eventGen = function(w, k) {
+              worldFunction(w, function(v) {
+                if(ffi.isSome(v)) {
+                  var xhr = new XMLHttpRequest();
+                  xhr.open("POST","https://api.particle.io/v1/devices/events");
+                  xhr.setRequestHeader("Content-type","application/x-www-form-urlencoded");
+                  xhr.send("access_token=" + that.options.acc +
+                           "&name=_event&data=" + that.event + ":" +
+                           runtime.getField(v, "value") + "&private=true&ttl=60");
```
jpolitz: What are the situations where this could fail?  It seems like a failure to POST will be silently ignored (which might be the intended semantics), but that could also be kind of baffling.

jpolitz:
Cool, just waiting for Travis to finish.  A few minor comments, but nothing stopping this from merging that I see.

jpolitz:
Weird – a field-not-found error for on-particle?

https://travis-ci.org/brownplt/pyret-lang#L1285

jpolitz:
Ah I see there are still export specs for the types of `on-particle` and `to-particle` in world.js, which are no longer exported as values, so `include world` wouldn't work.  There some `particle-shim-structs` still in that file; can that all be removed, too?

src/js/trove/particle.js
sstrickl: Events can sadly be lost for other reasons also (such as rate limiting), so even getting back a 200 won't necessarily mean it succeeded, I don't think, but I'll check into it.  There's more work that needs to go into this anyway, but this is intended to be a stable enough checkpoint for others here to try without having to go through the local CPO dance, so more robust error handling is definitely something on the docket.

sstrickl:
Yep, I see that now.  Sorry about that, will push a fix.

src/js/trove/particle.js
jpolitz: Yeah I figured this is tricky.  I think it'll either end up being silent failure, or find some way to check for any kind of failure and just kill the world if anything at all fails, so the user knows something is wrong.

PR #581, opened 2015-10-15
Added Documentation for List Functions

### LINKED ISSUE #568: join-str and friends are missing documentation
http://www.pyret.org/docs/latest/lists.html#%28part._lists_.List_join-str%29

belph:
Minor edit which fixes #568.


docs/written/trove/lists.scrbl
```diff
@@ -1623,13 +1623,104 @@ end
 }
 
 @list-method["member"]
+
+Returns true if the @pyret{elt} is located somewhere in the list.
+
+@examples{
+check:
+  [list: 'a', 'b', 'c'].member('a') is true
```
sorawee: Perhaps `'a' satisfies [list: 'a', 'b', 'c'].member` is better?

docs/written/trove/lists.scrbl
```diff
@@ -1623,13 +1623,104 @@ end
...
+  [list: 'a', 'b', 'c'].member( 1 ) is false
+  [list: ].member('a') is false
+end
+}
+
 @list-method["append"]
+
```
sorawee: Don't know if it's a good idea to mention that `<list> + <list>` is also appending

sorawee: Nevermind. There instead should be a documentation that lists have `_plus` which is appending

docs/written/trove/lists.scrbl
belph: I think that the way it's written is slightly better, since it illustrates an explicit usage of the method. Additionally, it follows the conventions that I'm seeing elsewhere in the documentation.

blerner: Agreed.  Using `satisfies` with a curried method call seems needlessly tricky for an example meant to illustrate how the method works.

docs/written/trove/lists.scrbl
belph: I agree. I'll add that in and add it to the PR.

docs/written/trove/lists.scrbl
sorawee: Yeah, I am actually neutral on this, but recently Shriram suggests that it's a good idea to produce a hint message to use `satisfies` when someone writes `f(x) is true`. If this is going to be implement, I think that we shouldn't have `f(x) is true` in the documentation.

sorawee:
By the way, since you are editing this file, can you change both `.each` and `each` and variants as well? The first argument (which is a function) could return anything (`Any`), not `Nothing`.

docs/written/trove/lists.scrbl
blerner: I think these are two separate things.  `f(x) is true` is stating something about `f`.  `x satisfies f` is stating something about `x`.  If the documentation is about `f`, I'd use the first version.

In general, sure, it's a stylistic nit to say `f(x) is true` when you could say `x satisfies f`, just as much so as `if x: true else y` is better written as `x or y`, etc...

belph:
I think that is something that was generated from [the definition of the each method](https://github.com/brownplt/pyret-lang/blob/master/src/arr/base/lists.arr#L127). While non-`Nothing` returning methods *do* seem to type check, shouldn't the change be made there instead? 

I did see that @jpolitz made some changes involving documentation generation in d3e1869, though. I might be misunderstanding, but does that commit remove the pre-generated documents altogether? Or does it just remove the generated documents as a dependency for building?

In short, where (if anywhere...maybe it's really intended to be `Any -> Nothing`) should this change occur: in `lists.arr`, `lists.scrbl`, or both?

jpolitz:
Thanks!

PR #582, opened 2015-10-15
New JSON library for Pyret

sstrickl:
This adds a JSON manipulation library for Pyret.  There is an initial attempt at docs based off the docs for the `s-exp` library, but the JSON structs also have two methods defined on them, `native` and `serialize`, that convert the JSON structs to "native" Pyret values and to a JSON string, respectively.  I'll be glad to add docs for those as well before this is merged if I'm pointed to a good example of the right way to do them within the docs directory.

### DIFF
--- a/src/arr/trove/json-structs.arr
+++ b/src/arr/trove/json-structs.arr
@@ -20,7 +20,7 @@ data JSON:
       end
       "{" + l.join-str(", ") + "}"
     end
-  | j-arr(l :: List) with:
+  | j-arr(l :: List<JSON>) with:
     native(self):
       self.l.map(lam(x): x.native() end)
     end,
--- a/src/arr/trove/json-structs.arr
+++ b/src/arr/trove/json-structs.arr
@@ -1,0 +1,58 @@
+provide *
+provide-types *
+
+import string-dict as SD
+
+data JSON:
+  | j-obj(dict :: SD.StringDict<JSON>) with:
+    native(self):
+      d = self.dict
+      ret = [SD.mutable-string-dict:]
+      for map(s from d.keys().to-list()):
+        ret.set-now(s, d.get-value(s).native())
+      end
+      ret.freeze()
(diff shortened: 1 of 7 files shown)


src/arr/trove/json-structs.arr
```diff
@@ -0,0 +1,58 @@
...
+      ret.freeze()
+    end,
+    serialize(self):
+      d = self.dict
+      l = for map(s from d.keys().to-list()):
+        '"' + s + '": ' + d.get-value(s).serialize()
+      end
+      "{" + l.join-str(", ") + "}"
+    end
+  | j-arr(l :: List) with:
```
blerner: List of whats?  (`List` on its own is a bare type...)

sstrickl: Ack, good catch.  List<JSON>.  I'll fix.

sstrickl:
Just realized that I don't think I added the test file to the appropriate suite. Will do once I finish commute home.

sstrickl:
No, I was wrong, I just checked and it *is* included in the right place, so we're set (on the test front, at least).

sstrickl:
Actually, just realized that there's one more thing I want to add before this gets merged, so going to temporarily close.

sstrickl:
Hmm, I just realized I'm unsure whether there's a way to do what I want to do.  @blerner, @jpolitz, @ds26gte, is there a way to tell whether an arbitrary value is a `string-dict`?  I don't see `is-string-dict?` or the like exported from `string-dict.js`.

(For context, I'm trying to write a `tojson` function that takes arbitrary Pyret values and turns them into a `JSON` value (if possible).)

ds26gte:
`is-string-dict` and `is-mutable-string-dict` are now in `pyret-lang#horizon`.

sstrickl:
Due to rebasing, I'll have to reopen this.

PR #583, opened 2015-10-16
New JSON library for Pyret

sstrickl:
This adds a JSON manipulation library for Pyret. There is an initial attempt at docs based off the docs for the s-exp library, but the JSON structs also have two methods defined on them, native and serialize, that convert the JSON structs to "native" Pyret values and to a JSON string, respectively. I'll be glad to add docs for those as well before this is merged if I'm pointed to a good example of the right way to do them within the docs directory.

(For earlier discussions, see #582.)

### DIFF
--- a/src/arr/trove/json-structs.arr
+++ b/src/arr/trove/json-structs.arr
@@ -20,7 +20,7 @@ data JSON:
       end
       "{" + l.join-str(", ") + "}"
     end
-  | j-arr(l :: List) with:
+  | j-arr(l :: List<JSON>) with:
     native(self):
       self.l.map(lam(x): x.native() end)
     end,
--- a/src/arr/trove/json-structs.arr
+++ b/src/arr/trove/json-structs.arr
@@ -1,0 +1,58 @@
+provide *
+provide-types *
+
+import string-dict as SD
+
+data JSON:
+  | j-obj(dict :: SD.StringDict<JSON>) with:
+    native(self):
+      d = self.dict
+      ret = [SD.mutable-string-dict:]
+      for map(s from d.keys().to-list()):
+        ret.set-now(s, d.get-value(s).native())
+      end
+      ret.freeze()
+    end,
+    serialize(self):
+      d = self.dict
+      l = for map(s from d.keys().to-list()):
+        '"' + s + '": ' + d.get-value(s).serialize()
+      end
+      "{" + l.join-str(", ") + "}"
+    end
+  | j-arr(l :: List) with:
+    native(self):
+      self.l.map(lam(x): x.native() end)
+    end,
+    serialize(self):
+      "[" + self.l.map(lam(x): x.serialize() end).join-str(", ") + "]"
+    end
+  | j-num(n :: Number) with:
+    native(self):
+      self.n
+    end,
+    serialize(self):
+      tostring(self.n)
+    end
+  | j-str(s :: String) with:
+    native(self):
+      self.s
+    end,
+    serialize(self):
+      '"' + self.s + '"'
+    end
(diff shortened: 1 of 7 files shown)


src/arr/trove/json-structs.arr
```diff
@@ -0,0 +1,102 @@
...
+      "[" + self.l.map(lam(x): x.serialize() end).join-str(", ") + "]"
+    end
+  | j-num(n :: Number) with:
+    native(self):
+      self.n
+    end,
+    serialize(self):
+      # This feels like a pretty big hack.  All I want is to convert a
+      # roughnum (or a rational) as a JavaScript number, which just
+      # means cutting off the initial "~".
```
jpolitz: Yeah this suggests we should have a "num-format" for such situations.  Though with rationals, it's actually worth thinking through each situation, so...

Note that this means that going to JSON and back from a Pyret value is lossy.

tests/pyret/tests/test-json.arr
```diff
@@ -0,0 +1,89 @@
...
+  n("0") satisfies is-number
+  n('"a"') satisfies is-string
+  n("[]") satisfies is-empty
+  # n("{}") satisfies SD.is-string-dict # apparently there's no is-string-dict?
+  n("{}") is [SD.string-dict:]
+  n("true") satisfies is-boolean
+  n("false") satisfies is-boolean
+  n("null") satisfies is-nothing
+
+  n("5.1") is%(within-abs(0.000001)) ~5.1
```
jpolitz: Is this test saying we expect this to be a roughnum, and not exactly 5.1?  I'm pretty sure this will return a Rational, per https://github.com/brownplt/pyret-lang/blob/master/src/js/base/js-numbers.js#L160.

I guess the JSON spec itself doesn't mandate a precision, but it's essentially going to be JavaScript's (double) in every reasonable application.

src/arr/trove/json-structs.arr
sstrickl: Yep, it is definitely lossy.  But I also didn't want to restrict the nums this was allowed to work with to be fixnums (which doesn't include the JS doubles), so I had to figure out some way of representing non-integers.  Changing rationals to roughnums at least guarantees they'll fit into a JS value if something that's not Pyret uses the resulting JSON.

I also considered removing access to the constructors so that the only only way from Pyret to JSON is through `tojson` and from JSON to Pyret via `native`, so that I could control exactly what numbers are in j-nums.  However, I think that'd be bad both for using this in a pedagogical setting and for typing (since the "native" encoding is not typeable, at least in any type system Pyret is likely to use).

tests/pyret/tests/test-json.arr
sstrickl: Originally I thought it would return a Roughnum, but when I wrote another test that actually checked that, I found out that, just as you say, it returns a Rational.  I figured I'd leave the test in it as it is, though, in case we ever had a way of forcing Roughnums when converting JS nums to regular numbers and this library was changed to use that.  (After all, it's still true now with Rationals!)

jpolitz:
Cool.  I don't think the issues I pointed out should block this, they're just worth thinking about.  I suspect we won't be able to predict all the odd number cases that come up, and something will surprise us.

It is a little unfortunate that JSON isn't a good serialization format for _Pyret_ values, because there will always be annoyances with rationals, but I don't see a simple way to avoid that and produce reasonable valid JSON at the same time.

We could have an option to produce "Pyret" JSON, that has some special structure in it for rationals (like an object with a special field), that one could use to keep track of exactness better.  But that's not for this PR.

PR #594, opened 2015-10-26
Adds support for core-specific commands and raw events.

sstrickl:
These commits add support for the `"core"` and `"raw"` option keys, and also has `configure-core` take in the same options hash (so that it can be shared between all calls if desired).  The `"core"` key should be mapped to a string that is the short name of the desired core (*not* the hexadecimal ID).  The `"raw"` key should be set to either `true` or `false`, and if true, sends a regular event on the Particle stream, not a core stub-compatible `_event` wrapped event.  This makes it possible to send JSON-encoded data using the Particle stream between the same world or different worlds, or to interact with non-stub-using Particle cores.

### DIFF
--- a/src/js/trove/particle.js
+++ b/src/js/trove/particle.js
@@ -28,6 +28,10 @@ define(["js/runtime-util", "js/ffi-helpers", "trove/json", "trove/world", "trove
               var v = runtime.getField(sd, "get-value").app(k);
               if(typeof(v) === "string") {
                 ret[k] = v;
+              } else if(v === runtime.pyretTrue) {
+                ret[k] = true;
+              } else if(v === runtime.pyretFalse) {
+                ret[k] = false;
               } else {
                 throw new Error('unimplemented value conversion for StringDict');
               }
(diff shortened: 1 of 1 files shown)


src/js/trove/particle.js
```diff
@@ -70,12 +67,38 @@ define(["js/runtime-util", "js/ffi-helpers", "trove/json", "trove/world", "trove
 
 
           //////////////////////////////////////////////////////////////////////
-          
-          var ToParticle = function(handler, ename, sdict) {
+
+          var sendEvent = function(ename, data, options) {
+            var xhr = new XMLHttpRequest();
```
jpolitz: Meta-comment: I wonder if we should just allow ourselves a hard dependency on JQuery or not.  XMLHttpRequest is kind of a pain IMO – how annoying do you find raw XMLHttpRequest?

jpolitz:
Cool, will merge when Travis finishes.  I like the refactoring of moving `sd_to_js` to where you did.

src/js/trove/particle.js
sstrickl: It hasn't been a problem for me, but then again, I haven't been doing anything ultra-fancy with it.

sstrickl:
Is this waiting on anything else?  (Not sure if the discussion over Slack re: checkISD/checkMSD and being exported for JS use from string-dict means you'd like to see this on this branch or separately.)

jpolitz:
Nope.  Travis just took long enough that I forgot to do this on Monday, sorry about that.  Feel free to click "Merge" on these if it seems like everything is reviewed and resolved; it seems like once another pair of eyes go over these PRs they are working out fine.

PR #607, opened 2015-11-27
fix #604 (empty hash-comment causes next line to be skipped)

### LINKED ISSUE #604: empty hash-comment causes next line to be skipped
Running this program
```
fun f(x):
  x + 10
where:
  f(10) is 20#foo
  f(20) is 30
end
```
says both tests passed, but changing the `#foo` to `#`, so the line reads
```
  f(10) is 20#
```
says only one test passed (the second test, in this case the one producing 30, is the one skipped).

The problem is not specific to tests. If I change the body of the function to
```
  x + 10#
```
I get an error saying I can't use `is` outside a `check` or `where` block.

Credit to Felix Merk for finding this problem.

sorawee:


blerner:
Can you confirm that this negative-lookahead is supported in all the browsers we care about?  Otherwise we can't use it...

sorawee:
Yup, will test.

sorawee:
Works on Chrome, Firefox, and Safari. Don't have IE or Microsoft Edge to test (we don't care these, right?).

PR #609, opened 2015-12-14
Fix typo in CLI arg retrieval; forward tail-call arg to compiler

btlachance:
I'm not sure if the --build option is even used much (compared to --compile-module-js), but it gave me trouble without these changes.

Also, it looks like pyret.arr and a few other places should be using `default-compile-options` from compiler-structs -- if that's the case, and if it'd be helpful, I can re-submit or make a new PR.

### DIFF
--- a/src/arr/compiler/pyret.arr
+++ b/src/arr/compiler/pyret.arr
@@ -107,13 +107,14 @@ fun main(args):
         end
       else:
         if r.has-key("build"):
-          result = CLI.compile(r.get-value("compile"),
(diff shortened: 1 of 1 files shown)


jpolitz:
Thanks!  `compile-module-js` and `build` are two ways of doing roughly the same thing; soon the goal is to switch to the `compile-lib.arr` definitions of compile rather than those in `compile.arr`.

I'll merge this, if you want to hunt down `default-compile-options` spots separately, please do and I'd be happy to merge that as well.

Thanks!

PR #613, opened 2015-12-27
Minify horizon

awstlaur:
Pulling these changes will minify the blob that is pyret, on horizon. This commit also includes the newly boostrapped pyret.js compiler, which passes all Selenium tests on cpo's horizon branch, and builds a phase1 which passes the local tests. We also have *-no-minify targets, so that the original, non-minified standalones can be made for dev purposes.

This minification saves a little over 7 M on the uncompressed side, and about 0.9 M on the gzipped side.

Running this code on the command line was only slightly faster, i.e., the speed up is not significant.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,6 +119,15 @@ standalone2: phase2 $(PHASE2)/pyret.js
 .PHONY : standalone3
 standalone3: phase3 $(PHASE3)/pyret.js
 
+.PHONY : standalone1-no-minify
+standalone1-no-minify: phase1 $(PHASE1)/pyret-no-minify.js
+
+.PHONY : standalone2-no-minify
+standalone2-no-minify: phase2 $(PHASE2)/pyret-no-minify.js
+
+.PHONY : standalone3-no-minify
+standalone3-no-minify: phase3 $(PHASE3)/pyret-no-minify.js
+
 $(PHASE1):
 	@$(call MKDIR,$(PHASE1_DIRS))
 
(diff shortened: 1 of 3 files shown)


awstlaur:
Note: I rebased this branch to keep with `horizon` just to make sure everything is smooth. In particular, I have _not_ updated the phase0 build since the initial pull request.

blerner:
I will happily merge this PR once Joe and I land the new module system -- since among other things, it'll affect how the bootstrapping process works.  Once that lands, I'll ping you with instructions on how the new system works, you can update this PR, and we'll merge.

awstlaur:
Sounds perfect, thanks!

awstlaur:
@blerner Now that the new module system appears to have landed safely, does the spirit of this diff translate straightforwardly?

blerner:
@awstlaur Can you rebase this on memo-jsast?  Take the phase0/pyret.js from the memo-jsast branch since that's a lot more recent than whichever phase0 you have in this PR.  I took a look at the changes in the makefile, but there are a lot of them and I don't understand exactly what you were intending, so I don't want to inadvertently mess it up...  Once this is rebased, I'll land it.

awstlaur:
@blerner This PR predates the "new module system", so it's probably best to start this from scratch. I'll make a note to investigate this off of memo-jsast.

jpolitz:
I'm going to close this for now.  Our reticence to commit to minifying ended up making sense because now we actually have some constraints from doing source-maps.  In particular, we can't get good reverse source-mapping without relying on some of the newlines that we put around function calls, which would need to get conserved.

It's really good to have these commits logged, but I don't think this PR in this form will get merged.

PR #622, opened 2016-01-13
[WIP] TCO

sorawee:


src/arr/compiler/anf-loop-compiler.arr
```diff
@@ -652,22 +652,42 @@ fun compile-split-method-app(l, compiler, opt-dest, obj, methname, args, opt-bod
...
-    j-block([clist:
-        # Update step before the call, so that if it runs out of gas, the resumer goes to the right step
-        j-expr(j-assign(step, after-app-label)),
-        j-expr(j-assign(compiler.cur-apploc, compiler.get-loc(l))),
-        check-fun(j-id(compiler.cur-apploc), compiled-f),
-        j-expr(j-assign(ans, app(compiler.get-loc(l), compiled-f, compiled-args))),
-        j-break]),
-    new-cases)
+  if identical(compiler.ret-label, after-app-label) and
+     J.is-j-dot(compiled-f) and
```
sorawee: `J.is-j-dot(compiled-f) and J.is-j-id(compiled-f.obj)` is unnecessary.

src/arr/compiler/anf.arr
```diff
@@ -21,27 +21,29 @@ end
...
-          name = mk-id(l, "cont_tail_app")
-          N.a-let(l, e-name.id-b, expr,
-            N.a-lettable(l, N.a-app(l, N.a-id(l, self.name), [list: e-name.id-e])))
-      end
+end
+
+data Scope:
+  | scope-c(bindings :: List<A.LetrecBind>, current :: Option<A.Name>) with:
+    setup-bindings(self, bindings :: List<A.LetrecBind>) -> Scope:
+      scope-c(bindings + self.bindings.filter(lam(b :: A.LetrecBind): A.is-s-lam(b.value) end), self.current)
```
sorawee: I put `.filter` at a wrong identifier. Will change.

jpolitz:
My overall complaint is that I wish this hadn't added to the ANF algorithm; it would be nice if the `scope` stuff were all in a separate visitor.  This is because I would like ANF to stand on its own; we might want to do various kinds of inlining, constant propagation, etc, after ANF and before detecting recursive calls.  This is in part because those other passes might find _more_ cases of recursive calls.

So I would like to see the recursion detection written as a visitor, rather than as another argument to anf.

Also, I was thinking that calls would have two pieces of information on them: recursive-ness and tail-call-ness.  Tail-ness could also be a separate, post-ANF pass that would tag some calls as tail-recursive if appropriate.

(Note that if these passes are too expensive memory-wise, we could make `is-recursive` and `is-tail-call` be mutable fields, and just walk the AST and update them, rather than creating whole new ASTs each time.)

It looks like tail-ness is discovered implicitly by the last step of the compiler.  Now that I see the code, that does seem like a reasonable place to put it since the compiler knows a lot about structure and can make good decisions.  In fact, it had _better not_ see an `a-app` with `is-tail-call` true except in the positions that it currently passes `true` to the application compiler.

I think that adding `is-recursive` as a flag is the worst of three options.  The other two are:

1.  Define `a-rec-app` that defines recursive applications
2.  Define a field `app-info` that describes extra details about the application, and is an object.  This makes things much more extensible if we want to make tail calls an attribute of an application expression, or, in the future, other information like knowledge that the application _cannot_ trigger stack management, etc.

blerner:
I like the last option, of an `app-info` field, the best.  I'm curious why you want tail-ness to be discovered after ANFing; it seems like you have precisely the right information available during ANFing itself to record tail-ness, no?

sorawee:
I will summarize a discussion about this PR with Joe here:

We decide to create a new constructor called `s-app-enriched` which contains a field `app-info`. This `app-info` will record the information whether it's `is-recursive` and `is-tail`. Right before transforming the program to ANF, we will transform all `s-app` to `s-app-enriched`. The reason we will do it in this stage is that

1. it still have rich information about "function". After ANF'ing, we will have a lot of ANF variables instead, and it's hard to track what is function and what not.
2. we will have a precise definition of "tail call" for Pyret.

blerner:
Sounds good to me; looking forward to seeing the next revision of this PR :)

sorawee:
Will add some tests in the next PR.

jpolitz:
I'm working on reviving this.  What are the open questions that make it WIP, or was it left in a ready-to-go state?

sorawee:
@jpolitz It's WIP because of two reasons:

1) we have discussed that we want the loc displayed in error message should be faithful. Consider:

      fun foo(t):
        cases (Something) t:
          | b1(x) => foo(x)
          | b2(x) => foo(x)
          | b3(x) => foo(x)
          | b4 => 1
        end
      end

  on `t = b1(b2(b3(b4)))`. The current elided frame mechanism in this PR completely ignore these locs and only indicate that foo (question: which foo?) is called with some frames elided. 

2) I think our mechanism to see if a refinement doesn't have a side effect is still not quite correct.

belph:
Just popping into this PR (haven't taken a look at the code), but, regarding point (1), what about somehow allowing for a stack frame to have multiple locations (corresponding to a stack trace)? I understand this is a nontrivial change, but it could remove limitations such as the one found here.

jpolitz:
@belph: this is straightforward to do, actually.

Each tail-calling application currently increments a counter for the number of "elided frames".  It could relatively easily push onto a stack the location information of the call.

The issues is that

(a) It's not safe-for-space, which is a goal of tail calls (and a fundamental tension point we're butting up against here)
(b) It adds overhead and allocation to the case we're trying to make fast, which is looping around instead of doing something expensive, like making another call

Neither of those is totally damning against the idea, though.

blerner:
"Safe for space" is a slippery notion here.  It uses some heap space, but doesn't use any stack space...and then again, our functions use heap space too to be (call-)stack-safe.  To my mind, the explanatory win of having accurate stack traces probably makes up for the extra heap space they use.  And my hunch is the speed hit is likely to be small/in the noise -- I suspect that avoiding the function call (and try/catch blocks that go with them) will pay for this fairly easily.  I worry a little about 10K-element arrays, though, for truly deep stacks: I don't know whether Array.push stays amortized-constant time when arrays get really big...

belph:
Additionally, I feel as though trying to be safe for space while still preserving stack traces is a non-sequitur; Ultimately, the information about stack traces is going to have to go _somewhere_ (either a location stack or in multiple activation records). Given that, it would make more sense (IMO) to perform tail-call optimization and grow a stack of locations than to only perform the optimization where consecutive activation records have the same location.

jpolitz:
> our functions use heap space too to be (call-)stack-safe

We're actually quite careful to _not_ append stack information when collecting frames for a tail call, making them actually safe-for-space.

You can run

```
f = lam(f): f(f) end
f(f)
```

and watch the task manager in Chrome not grow in memory usage.  Compare to something like

```
f = lam(f): f(f) + 1 end
f(f)
```

which grows at about 50MB/sec due to stored stack frames.

This is literally the _only_ way to write a long-running program in Pyret that doesn't eventually crash with an out-of-memory error.  That's a pretty serious language weakness if we don't support safe-for-space.

> Additionally, I feel as though trying to be safe for space while still preserving stack traces is a non-sequitur

This is why I feel that something like explicit syntax for tail calls is the way to go.

blerner:
Devil's advocate questions: is writing long running Pyret programs something we really need to be optimizing *the language* for? "Surely" for any serious usage, we'd have a compiler flag, rather than a language keyword, that discards all debugging information (srclocs, stack traces, doc strings, possibly even stack safety or annotation checks?) and compiles for faster leaner code.

Also, having another keyword pushes up the language complexity budget, but this time in a fairly abstract spot. "Wait, so, I write this keyword, and Pyret can tell if it's in the right spot or not, and my code uses less memory but something about stacks? My code will never need that (I think...) so why can't Pyret Just Do It For Me?" I have the same nebulous problem with this keyword that Shriram does with our current explanation of `shadow`: if the language can implicitly figure out when it's needed, why is it needed explicitly? It's just another thing to have to remember.

blerner:
Devil's advocate questions: is writing long running Pyret programs something we really need to be optimizing *the language* for? "Surely" for any serious usage, we'd have a compiler flag, rather than a language keyword, that discards all debugging information (srclocs, stack traces, doc strings, possibly even stack safety or annotation checks?) and compiles for faster leaner code.

Also, having another keyword pushes up the language complexity budget, but this time in a fairly abstract spot. "Wait, so, I write this keyword, and Pyret can tell if it's in the right spot or not, and my code uses less memory but something about stacks? My code will never need that (I think...) so why can't Pyret Just Do It For Me?" I have the same nebulous problem with this keyword that Shriram does with our current explanation of `shadow`: if the language can implicitly figure out when it's needed, why is it needed explicitly? It's just another thing to have to remember.

jpolitz:
> Devil's advocate questions: is writing long running Pyret programs something we really need to be optimizing the language for?

There's a difference between _optimizing_ and _making possible_.

> Also, having another keyword pushes up the language complexity budget, but this time in a fairly abstract spot. "Wait, so, I write this keyword, and Pyret can tell if it's in the right spot or not, and my code uses less memory but something about stacks?

This is why it would come with pedagogy for tail calls, which is something a student who asks that question could benefit quite a bit from.

> My code will never need that (I think...) so why can't Pyret Just Do It For Me?"

Because @blerner said you should get better debugging information in that spot :->

blerner:
Those answers miss the point of my question - why does this need to be a language keyword, rather than a compilation setting? Having to flag every tail call as such seems onerous, error prone and boring. 

The same pedagogical benefit can be obtained by running the exact same program with and without TCO, and with and without stack information in the TCO case. Only the lattermost combo results in safe for space programs, and does so without the student having to comb through their entire program looking for tail positions. 

(Which, come to think of it, are harder to spot than we might like, thanks to return annotations on functions...:-/ It'll be really annoying to try to teach how to recognize trail positions by their appearance inside a function body, and then have to walk back the advice we give about annotating programs well... Yes, it can be done, but it shouldn't have to.)

If the best argument to be made for keeping the information is, because I advocated for it, then I'm advocating for a different solution for *not* keeping it, too :-)

jpolitz:
I haven't thought a lot about the pedagogy of language _tools_.  The optional type-checker is our first, and other command-line flags, like this one for TCO, would also go in that "drop-down."  What do we know about teaching students build systems and compile options that are extra-lingual?

It's certainly important.  I don't think students retain much about what I teach them about Makefiles or flags to gcc, except maybe in my compilers course (even then I provide build infrastructure).  It's interesting to think about what it means to teach this explicitly.

belph:
More devil's advocacy:

+1 for making SFS a compiler flag. We could enable the "less aggressive" version without the flag, which would cover a large portion of primitive (and non-mutual) recursive definitions (as often seen in early CS courses), since the consecutive stack frames would have the same location. Then, for more advanced usage, a compiler flag could disable disable the stack traces in favor of being safe for space. I am not a fan of the new keyword; it leaves a similar taste in my mouth to usages of `register` in C. Additionally, for _our_ usage, non-stack-preserving TCO should be something easily disabled in order to debug code; the last thing I want is to choose between disabling _all_ optimizations (using `standaloneX.json`) or modifying all of my code to not use the keyword. The keyword seems to be the equivalent of sprinkling many copies of what should be one compiler flag across one's code.

jpolitz:
A question relevant for decisions like this discussion: Is there a principle that separates this case—making SFS tail calls a compiler option, rather than syntax—from the `block:` keyword change, which added a keyword rather than making it a compile option?

Put a different way, was `block:` a mistake?  Should there be a checkbox in the run dropdown for "allow multiple expressions in a block?"

belph:
My two cents: I don't think that falls under the same category as this issue. `block:` serves a useful purpose as a means of interacting with the well-formedness checker; if you forgot it, _you're likely making a mistake_ (or just absent-mindedly threw in a debugging call to `print`). It has a similar use as the `begin` keyword in OCaml, with the added benefit of not requiring an extra level of indentation. So no; I disagree with the idea that `block:` was a mistake. 

The difference here is that the impetus for using a TCO keyword is _solely_ from wanting the Pyret runtime to make your code faster, a purpose which is orthogonal to the semantic meaning of your program. `block:` expresses intent with respect to _what_ you want your program to evaluate, whereas a TCO keyword expresses intent with respect to _how_ you want the Pyret runtime (specifically!) to evaluate it.

jpolitz:
I do want to make the point that TCO is not just about speed.  It changes the _asymptotic space complexity_ of the program in a fundamental way that makes some programs fit into finite machines that wouldn't otherwise.

So saying it's _solely_ from wanting the Pyret runtime to make your code faster doesn't capture the whole story.

I also don't understand why `block:` is the right way of interacting with the well-formedness checker – why isn't a checkbox in the Run dropdown/a compiler flag the right way of interacting with the WF checker?

I'm pushing on this point because it feels like there's some design principle underlying this that hasn't been clearly stated.

blerner:
Blocks can independently be single- or multiple-expressions, and checking that on a per-block basis is a worthwhile conceptual effort for a programmer.  Put another way, I may want the safety net of checking blockiness independently throughout my program.  Turning tail-calls on or off rarely makes sense on a per-function basis: either I want the (potentially-)asymptotic performance improvements, or I don't need them.  Gaining their benefit is most likely a whole-program property, not a per-function property.

jpolitz:
@blerner: I would love to merge this, and just leave TCO off by default.  It has the `RUNGAS` support as a part of it, which would be good to integrate throughout the system (and not deal with a separate merge).

Of the two issues @sorawee mentioned:

    1) we have discussed that we want the loc displayed in error message to be faithful.
    2) I think our mechanism to see if a refinement doesn't have a side effect is still not quite correct.

For (1), the information issue may be moot if we're having a checkbox that explicitly turns on PTC and loses information by design.

For (2), @sorawee is right; this is going to require some of the work from https://github.com/brownplt/pyret-lang/pull/915 to be correct, where annotations can export whether they contain refinements or not (along with other static information).  However, if PTC is _off_ by default, then turning it on just becomes a slightly-unsafe optimization, which will become totally safe once we have that static information.

We have a lot of branches floating around, and I'd like to do some consolidation work, is another reason I'm pushing this a bit.

What do you think?

blerner:
ok

src/arr/locker/~tables.arr
```diff
@@ -0,0 +1,86 @@
+#lang pyret/library
+
```
cartazio: why is this file in this pr? i presume ~foo style files are some editor induced file?

sorawee: This PR is created last year. A lot of things have changed and when we came back to it recently we had a bad merge. So, yes, the file you mentioned is not intended to be included in this PR. It must be cleaned before merging.

blerner:
@sorawee , @jpolitz : is this PR worth closing and re-starting, to rebase it and clean up its extraneous accumulated cruft?  Ain't no way the actual TCO work affected 110 files over 250 commits...

jpolitz:
This is all merged into a different branch now – lots of these commits were just merging horizon _into_ this PR.  But merging it onto return-stack seems went fine, and required all the delicate work anyway.  So once that branch gets settled, I think this PR can be closed.  It's useful to have as a placeholder for all the discussion about stack frames and errors above, but the code itself has drifted somewhat.

jpolitz:
Closing per the above comment, all of this is in horizon as of https://github.com/brownplt/pyret-lang/pull/1042

PR #624, opened 2016-01-18
Use mutable string dicts for efficiency.

awstlaur:
### What changes?
* A `.merge-now` method added to `string-dict.js`, which mutates the object, and returns `Nothing`.
* Tests of `merge-now` added to test-string-dict.arr
* Four (4) files in `src/arr/compiler/` overhauled to use mutable string dicts in contexts involving collecting sets of identifiers. In particular, many uses of the `.merge` method for immutable string-dicts are replaced by the new `.merge-now`.

### Why merge this PR?
Efficiency. In larger programs, time spent allocating all the new objects is a waste. To show the speedup, below present the results from `ast.arr` and `anf-loop-compiler.arr` (as currently on horizon), using [these profiling tools](https://github.com/awstlaur/pyret-lang/tree/horizon/tools/profiling). In summary, `ast.arr` saw a roughly 50 (fifty) percent speedup, `anf-loop-compiler` saw a roughly 30 (thirty) percent speedup at the calls to `runEvalPyret`. Smaller, more trivial programs saw no significant speedup.

### Data
These files are in [callgrind format](http://valgrind.org/docs/manual/cl-format.html), and can thus be (visually) inspected with a tool like [kcachegrind](http://kcachegrind.sourceforge.net/). For more details on how these data were obtained, see the README [in the profiling tools](https://github.com/awstlaur/pyret-lang/tree/horizon/tools/profiling)
* [ast.arr: before the changes](http://cs.brown.edu/~awstlaur/pub/ast.BEFORE.profile)
* [ast.arr: after the changes](http://cs.brown.edu/~awstlaur/pub/ast.AFTER.profile)
* [anf-loop-compiler: before](http://cs.brown.edu/~awstlaur/pub/anf-loop-compiler.BEFORE.profile)
* [anf-loop-compiler: after](http://cs.brown.edu/~awstlaur/pub/anf-loop-compiler.AFTER.profile)

### Sanity check
* `make test` passes
* `make no-diff-standalone` passes

### DIFF
--- a/src/arr/compiler/anf-loop-compiler.arr
+++ b/src/arr/compiler/anf-loop-compiler.arr
@@ -1441,6 +1441,7 @@ end
 fun import-key(i): AU.import-to-dep-anf(i).key() end
 
 fun compile-program(self, l, imports-in, prog, freevars, env):
+  shadow freevars = freevars.unfreeze()
   fun inst(id): j-app(j-id(id), [clist: RUNTIME, NAMESPACE]);
   imports = imports-in.sort-by(
       lam(i1, i2): import-key(i1.import-type) < import-key(i2.import-type)  end,
--- a/src/arr/compiler/anf-loop-compiler.arr
+++ b/src/arr/compiler/anf-loop-compiler.arr
@@ -14,6 +14,7 @@ import srcloc as SL
 import sets as S
 
 string-dict = D.string-dict
+mutable-string-dict = D.mutable-string-dict
 
 type Loc = SL.Srcloc
 type CList = CL.ConcatList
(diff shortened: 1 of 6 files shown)


blerner:
I'm curious why you chose to make `merge-now` return nothing, instead of the invocant?  It makes the code less convenient this way.  Was there a precedent elsewhere in the code that you were following?

src/arr/compiler/anf-loop-compiler.arr
```diff
@@ -276,61 +277,137 @@ fun arity-check(loc-expr, arity :: Number):
...
+    ans = init.visit(self)
+    ans.merge-now(cond.visit(self))
+    ans.merge-now(update.visit(self))
+    ans.merge-now(body.visit(self))
+    ans
   end
 }
 
+fun copy-mutable-dict(s :: D.MutableStringDict<A>) -> D.MutableStringDict<A>:
+  s.freeze().unfreeze()
```
blerner: Clever :)

awstlaur:
I was following the precedent of the other mutable string dict methods, which all return nothing. I do agree it makes this use case less convenient.

PR #647, opened 2016-02-14
s-array optimization

sorawee:
Given that the parser is efficient (cherry pick https://github.com/brownplt/pyret-lang/commit/feb01d3847072c576c46d3e0ad23538221fb10b3). The following program:

```
fun f(x): x end
[list: f(1), f(2), ..., f(3000)]
```

take 5 minutes to compile. Worse, it's eventually crash. This pull request fixes the problem, and the compilation takes only 20 seconds.

@blerner


sorawee:
yay, finally! :)

blerner:
@jpolitz please double-check that this does still work.  IIRC, it certainly *_should_*, but when I tried merging it myself a week ago, I got some weird performance regressions.  Please double check that `make phaseB` actually goes faster, and then that the initial example compiles faster, too.  If so, hooray!

jpolitz:
So, I didn't manually merge this.  It came along for the ride with `new-horizons`, and has evidently been running happily on that branch for some time.  I haven't been timing, though things don't seem way off.

blerner:
Hah!  Ok, well, let's keep it then :)  It works on my machine...

jpolitz:
The build oh `phaseA` was within 10s (out of about 8 minutes) after I bootstrapped with this included.

PR #648, opened 2016-02-17
Use SVG asset for CI build status

liangricha:


liangricha:
Hmm I can't seem to download Racket locally either (getting a 403 as well). Out of curiosity, which tests depend on Racket?

liangricha:
The Racket DL issue seems to be resolved, and triggering a rebuild resulted in passing tests.

jpolitz:
Cool, thanks!

FWIW, Racket is used to build the documentation, which use https://docs.racket-lang.org/scribble/.

Happy to merge this; I like SVG.  Is there a particular reason to prefer it in this case?

liangricha:
The image will look nice in all screen resolutions :stuck_out_tongue_winking_eye: (it's currently kinda blurry on my MBP 15-inch retina). Travis CI seems to also be using SVG icons instead of PNGs for their status images. Not a huge deal in particular.

PR #688, opened 2016-03-29
Added Matrix Library

belph:
Adds support for matrices. This has been sitting around for a little while, but, from what I recall, this should implement most everything that Racket's `math/matrix` library does.

### DIFF
--- a/docs/written/libraries.scrbl
+++ b/docs/written/libraries.scrbl
@@ -20,6 +20,7 @@ This section contains information on libraries that come with Pyret.
 @include-section["trove/world.js.rkt"]
 @;@include-section["trove/plot.js.rkt"]
 @;@include-section["trove/graph.js.rkt"]
+@include-section["trove/matrices.scrbl"]
 
 @include-section["trove/srcloc.js.rkt"]
 @include-section["trove/pprint.js.rkt"]
--- a/docs/written/trove/matrices.scrbl
+++ b/docs/written/trove/matrices.scrbl
@@ -1,0 +1,1183 @@
+#lang scribble/base
+@(require "../../scribble-api.rkt"
+          (except-in "../abbrevs.rkt" L-of))
+@(require (only-in scribble/core delayed-block)
+          (only-in scribble/manual math))
+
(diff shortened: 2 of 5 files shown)


blerner:
@belph could you update this to the latest memo-jsast branch?  While you're at it, a thought for fixing submatrix above:  Let's split it into two methods, `submatrix : (int loRow, int hiRow, int loCol, int hiCol) -> Matrix` and `extract : (List<int> rows, List<int> cols) -> Matrix`.

belph:
Sure, I will do that.

jpolitz:
We don't have any lessons or applications that use matrices, is probably why this never got merged. This seems like a reasonable design. However, I kind of want to see an _application_ that uses the feature before throwing it in, to make sure we know what use cases we're solving.

Is there some clever reactor program we could write that uses matrices to calculate locations of characters, maybe including a z-index? Is there a cool existing open source example of doing matrix operations in Python somewhere that we could adapt to this interface to demonstrate what kinds of applications we want it for?

schanzer:

ds26gte:
I'm looking to add multiple-variable linear regression to the statistics library. Estimating the coefficients of the predictor function could use matrix methods (+, *, transpose, inverse). I could write my own just for this one use-case, but a well tested library would be nice.

schanzer:
@ds26gte [like this one](https://github.com/brownplt/pyret-lang/pull/1599)?

schanzer:
Would you like me to close that one, and reopen this PR?

ds26gte:
If a genuinely general-purpose and bulletproof matrix library requires more cogitation, I don't need to wait on it. I can write just-the-bare-minimum-I-need inside the statistics library itself, using no new data types. A matrix would simply be a list of lists. No matrix I use will be visible to the end user.

shriram:
Let's not wait for a bulletproof library. It's fine for now to have a library written in JS that only handles matrices of floats that resides internally to Dorai's system. (Since he's adding this to `statistics.arr`, it would effectively end up internal to there, which means it could be useful in other contexts too.)

PR #696, opened 2016-04-08
Error Overhaul

jswrenn:
https://github.com/brownplt/code.pyret.org/pull/57


src/arr/base/contracts.arr
```diff
@@ -33,13 +40,17 @@ data FieldFailure:
   | missing-field(loc, field) with:
     render-reason(self, loc, from-fail-arg):
       [ED.error:
-        [ED.para: ED.text("Missing field"), ED.code(ED.text(self.field)),
-          ED.text("is required at"), draw-and-highlight(self.loc)]]
+        [ED.para: ED.text("Missing field "), ED.code(ED.text(self.field)),
```
blerner: `ED.text` within an `ED.para` shouldn't need trailing whitespace.  Either use `ED.para-nospace` or remove the whitespace, here and everywhere.

jswrenn: I would rather nix `ED.para-nospace` altogether and follow a rule of what-you-write-is-what-you-get. `ED.para-nospace` seems like a needless complication. Let's say I want to write a helper function that takes some argument and depending on its value outputs one string or another. If we keep the `ED.para-nospace` and `ED.para` distinction, I cannot trust that that function will behave how I intended; I need to specialize it for the space-injecting and no-space scenarios and remember to call the right one.

blerner: The alternative, I think, is worse: we've had _so_ many bug reports about "missing space in error message X", and it's a pain to fix them one at a time.  I'd rather have the common case be "we'll stick space in for you".  Another possibility is to eliminate `ED.para-nospace`, and instead create a `ED.nospace` constant that suppresses auto-inserted space in an `ED.para`.  Again, the common case is spaces are needed.  As it stands, you can't write the helper function the way you wish, since it sounds like you want a splicing context (i.e., have a helper function splice a bunch of content into some outer content), and since `ED.para` doesn't take a list argument (it's its own constructor), that helper can't be written directly.  So in any case you'll need to know what you're splicing into, and so you'll know whether there's space around it or not.

jswrenn: Given that going through each error message one-at-a-time is precisely what I have been doing for the past few months and will continue to do for the foreseeable future, fixes relating to missing spaces come essentially for free.

One change in my branch is I've been moving towards a more semantically meaningful `ErrorDisplay`. `ED.para` no longer expands to an `h-sequence` in the errors branch but rather an actual `paragraph` [variant](https://github.com/brownplt/pyret-lang/pull/696/files#diff-b3dee8ff07b1bb3b67f4ea9f3ca07dceR7) of `ErrorDisplay`. This really simplifies things when it's time to render because it means that `para` actually corresponds to paragraphs.

Horizon's `ED.para` conflates two commonly used but distinct things: "paragraphs" and space-separated text. It sounds like what you want is a shortcut for denoting sequences of space-separated text. What if I added a `spaced` list constructor to `error-display` that would denote a sequence of space-separated `ErrorDisplay` instances. A use would look something like:
```pyret
ED.paragraph([ED.spaced:
    ED.text("The"),
    ED.text("quick"),
    ED.highlight([ED.spaced: ED.text("brown"), ED.text("fox.")], someloc, somecolor)])
```

blerner: I was hoping to spare you that level of nitpicky fixes :)  If you're willing to take that on, ok...  So in your branch, what does the `ErrorDisplay` now mean?  Looks like you have `paragraph`s, `v-sequence`s and `h-sequence`s: what do they correspond to?  I'm all for improving the `ErrorDisplay` data type; it was a first-draft implementation on my part, and I'm not wedded to its current design (or, by the same reasoning, to `ValueSkeleton`).

I actually wasn't looking for an explicit `ED.spaced` constructor, but was looking for a default where weneverwould wind up with missing spaces unless we opted in to that.  IMHO having extra spaces in English text is less annoying than having too few...

schanzer:
First off, I have to say how incredibly awesome this is. I immediately called Emma and jumped up and down telling her to check it out. Really, _really_ nice job on this.

I'm curious about the "highlight this" button that's being added to error messages. Every bit of complexity is a place where users can screw up, so I just want to poke at this for a moment. Is there a use-case that the button solves, which couldn't already be solved by having the highlight happen when an error message gets moused-over, and get rid of the button altogether?

(And if this discussion is better had in a different forum, let me know)

PR #707, opened 2016-04-29
Polymorphic newtype

mkolosick:
Added type variables on type alias definitions.


jpolitz:
Cool!  The environment changes all seem reasonable.

I don't see tests that use more than one type variable, or any "bad" tests – those seem like important cases to cover to explain when this can go wrong.

For example, should it be the case now that this fails:

```
type MyList = List
```

Also, it would be good to make sure the that type-arity-mistmatches work as expected.

blerner:
Yes, tests with multiple type parameters are essential for making sure any processing of the parameters (pretty printing, instantiation, inference, whatever) doesn't reverse the parameter list anywhere by mistake... Been there, done that :-)

Also, if we now have to deal with kind errors, we should get @jswrenn involved sooner than later.

mkolosick:
I don't think it makes sense to require the number of parameters to line up, as it is unclear what this would even mean. For example `type Parser<T> = (S-Exp -> Option<T>)` is the motivator for this feature. In complex types it isn't clear where the checks for number of parameters should come from.

I've also added in tests of with multiple type parameters, so I think this is worth merging now. Errors will come from trying to use polymorphic types like `Option` without parameters, but that is a separate design and programming concern.

Issue #719, opened 2016-05-06
Bugs in num-to-string-digit


sorawee:
```
> num-to-string-digits(1/3, 1000)
"NaN.0000000000000000000000000..."
```

Also, finding string-digits of a rough num ought to be possible.

```
> num-to-string-digits(~1, 10)
roughnums cannot be compared for equality
```

blerner:
These are pretty funny:
```
> num-to-string-digits(1/3, 10)
"0.3333333333"
> num-to-string-digits(1/3, 100)
"0.3333333333333333000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
> num-to-string-digits(1/3, 305)
"0.33333333333333337000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
> num-to-string-digits(1/3, 306)
"0.333333333333333300000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
> num-to-string-digits(1/3, 308)
"0.33333333333333330000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
> num-to-string-digits(1/3, 309)
"NaN.000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
```
The last digit before failing over to zeros flips between 3 and 7, and the cutoff for NaN printing seems to be between 308 and 309.

sorawee:
I fixed these already for my plotting lib :)

blerner:
Where and how?  (Can you link to a line number in a commit?)  I don't see a change in that PR that looks like it fixes it...

sorawee:
Ah, my bad. I forgot to commit. It should go to the `sql` branch by next hour. The fix is by using jsnums' `toRepeatingDecimal`

schanzer:
[...]

PR #722, opened 2016-05-06
new plot lib

sorawee:
@jpolitz @blerner FYI


sorawee:
Hopefully after the new module system is merged, all these changes could be done in the CPO repo instead. Seems that Shriram wants this to be done really soon though, so we might need to merge this PR to pyret-lang regardless.

sorawee:
Known bugs:

- When using roughnum as boundaries, things go off (because `num-to-string-digits` is buggy, see #719)
- Need to change rational repr to decimal repr
- `plot-xy` should not consume the win-option (use default value instead)

src/js/trove/plot-lib.js
```diff
@@ -0,0 +1,775 @@
...
+      }).attr('size', '8').addClass('yMax'));
+      controller.append($('<input/>', {
+        type: "text",
+        placeholder: "num-samples",
+        value: rt.num_to_string(gf(windowOptions, "num-samples")),
+        style: "left: 90px; top: 240px",
+      }).attr('size', '8').addClass('numSamples'));
+      
+      function getNewWindow() {
+        var elem1 = $('.xMin').last();
```
sorawee: These names are bad. Will change.

src/js/trove/plot-lib.js
```diff
@@ -0,0 +1,775 @@
...
+                          none: function() {
+                            elem5.addClass('error-bg');
+                            elem5.removeClass('ok-bg');
+                            return null;
+                          },
+                          some: function(numSamples_val) {
+                            elem5.removeClass('error-bg');
+                            elem5.addClass('ok-bg');
+                            
+                            if (rt.isPyretFalse(rt.num_is_integer(numSamples_val))) {
```
sorawee: Should check that it's >= 1

src/js/trove/plot-lib.js
```diff
@@ -0,0 +1,775 @@
...
+        height = dimension.height;
+
+      var detached = createDiv();
+      
+      var divSvg = detached
+              .append('div')
+              .attr('class', 'divsvg'),
+          canvas = divSvg
+              .append("svg")
+              .attr("width", 601)
```
sorawee: Get rid of these hard coded numbers

sorawee:
Need to do documentation

PR #732, opened 2016-06-07
New world-lib

### LINKED ISSUE #483: big bang box should be cleaned after it is closed
Every time I call big bang, the box will be appended to the DOM. However, it is not removed when the box is closed despite the fact that it will not be able to be referenced anymore.

This is not that serious for the normal big bang, as the element inside the the box is just a canvas. However, for the project that I am doing, which invokes big bang box via `runtime.getParam("current-animation-port")`, there could be a lot of SVG elements. It therefore would not be a good idea to leave these elements.

Currently, my workaround is to remove elements manually:

    var terminate = function () {  d3.selectAll(".maind3").remove(); }; 
    // maind3 is the outmost div that elements I am working with lie in.
            
    // simulate dialogclose
    d3.selectAll(".ui-dialog-titlebar-close").on("click", terminate);
    // this is actually bad. there might be a lot of '.ui-dialog-titlebar-close' which are irrelevent
    d3.select("body").on("keyup", function(e) {
        if (d3.event.keyCode == 27) { terminate(); } // hit esc key
    });

It however should be Pyret itself that handles this issue.

### LINKED ISSUE #543: big-bang doesn't nest properly
```
include world
include image
import image-structs as IS

big-bang(0, [list: 
    on-tick(lam(n):
        if num-modulo(n, 10) == 1:
          big-bang(n,
            [list:
              on-tick(lam(k): k - 1 end),
              stop-when(lam(k): k < 0 end),
              to-draw(lam(k): text(num-to-string(k), 20, IS.blue) end)
            ])
          n + 1
        else:
          n + 1 
        end
      end),
    to-draw(lam(n): text(num-to-string(n), 20, IS.violet-red) end),
    stop-when(lam(n): n > 100 end)])
```

sorawee:
Changes:

1. Resume computation when closing a big-bang window correctly
Prior this PR, when we have nested big-bangs and close the topmost big-bang window,
the computation of the previous window will not resume. Instead, the world value
of the bottommost window will be returned. Thix PR fixes this by
resuming the computation in the previous window properly.
2. Change changingWorld to a stack
Prior this PR, if the first big-bang has on-tick (which fires very frequently),
`changingWorld` could be set to true while the the second big-bang is loaded.
This shared changingWorld would cause the second big-bang to wait for the variable
to be false, but it will never be. Thus, the second big-bang will be in non-responsive state.
   
   Setting `changingWorld` to false when resuming the computation to the previous window is also
dangerous. It's possible that the new window is populated by something like:
   
        on-tick(lam(_):
            call-new-big-bang()
            wait-for-an-hour()
          end)
   
   Then, setting `changingWorld` to false could lead the callback to be called again while the old
one is still active. This could lead to an error like "Internal: run called while already running"
Thus, the best way to do this is to keep a stack of changingWorld instead.
   
   This fixes #543
   
3. Add `close-when-stop` per a discussion with @joe
A new WorldConfig which could be used to close the big-bang window when `stop-when` returns true.

4. Enforce modal dialog correctly
When a modal dialog appears, it should grey out everything else, including other dialog.
This is the correct functionality of JQuery UI modal dialog.
Prior this PR manual changes to z-index in Pyret make all dialogs accessible. This PR fixes
this issue.

   Note that one nice consequence is that we do not need to worry whether
the current computation / window should be halted / closed if users close the previous
window because the correct modal dialog will force users to interact with only the topmost window.

5. Prevent computation resuming until the dialog is closed
It doesn't make sense to resume the computation while the dialog is open
(even when the world has stoped) because the computation could then trigger a new big-bang
which will prevent viewing the old window which is the only reason we do not close the window
right away when the big-bang stops.

6. Clean big-bang box when closed
This fixes #483.

This PR should be merged along with https://github.com/brownplt/code.pyret.org/pull/77

With this PR, the following program should be able to run properly:
https://code.pyret.org/editor#share=0Bxr4FfLa3goOREVhM1hRMUJtWW8

### DIFF
--- a/src/js/trove/world-lib.js
+++ b/src/js/trove/world-lib.js
@@ -2,6 +2,13 @@
 define(["js/runtime-util"], function(util) {
 
   return util.memoModule("world-lib", function(runtime, namespace) {
+    var _worldIndex = 0;
+
+    var getNewWorldIndex = function() {
+        _worldIndex++;
+        return _worldIndex;
+    }
+
     var rawJsworld = {};
 
     // Stuff here is copy-and-pasted from Chris King's JSWorld.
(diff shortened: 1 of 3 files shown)


src/js/trove/world-lib.js
```diff
@@ -96,6 +142,23 @@ define(["js/runtime-util"], function(util) {
         clear_running_state();
     };
 
+    // Close all world computations.
```
jpolitz: This comment seems to contradict the name of the function `shutdownSingle`.

src/js/trove/world-lib.js
```diff
@@ -96,6 +142,23 @@ define(["js/runtime-util"], function(util) {
...
+            var currentRecord = runningBigBangs.pop();
+            currentRecord.pause();
+            if (options.cleanShutdown) {
+                currentRecord.success(world);
+            }
+            if (options.errorShutdown) {
+                currentRecord.fail(options.errorShutdown);
+            }
+        }
+        resume_running_state();
```
jpolitz: I feel like the three lines below belong in `resume_running_state` as well, unless there are other planned uses for `resume_running_state` other than this call (this is the only call to it I see).  Since it has `resume` in the name, it seems like it should be the one to call `start`.

src/js/trove/world-lib.js
```diff
@@ -604,8 +667,13 @@ define(["js/runtime-util"], function(util) {
         this.fail = fail;
     }
 
-    BigBangRecord.prototype.restart = function() {
-        bigBang(this.top, this.world, this.handlerCreators, this.attribs);
+    BigBangRecord.prototype.start = function() {
```
jpolitz: I actually like the name `restart` better for this, FWIW, because it is called multiple times per object, and complements calls to `pause`.

jpolitz:
This looks good.  It will need to be applied to CPO rather than to `pyret-lang`, since `world` has moved there.  But it should apply cleanly there.

I can create that PR in a few days, or you can copy these over into a horizon checkout of CPO and make this change there if you want, and I'll merge it from there.  Making the change on CPO is useful because there are some world tests that run on Selenium, also, so it will have some light automated testing applied to it.

sorawee:
Move to CPO (see https://github.com/brownplt/code.pyret.org/pull/93)

PR #734, opened 2016-06-10
new plot + table integration

sorawee:


src/js/base/runtime-anf.js
```diff
@@ -4088,25 +4088,13 @@ function isMethod(obj) { return obj instanceof PMethod; }
...
-      else if (jsnums.equals(n, Number.NEGATIVE_INFINITY)) { return thisRuntime.makeString("-inf"); }
-      else if (jsnums.isReal(n)) {
-        n = jsnums.divide(jsnums.round(jsnums.multiply(n, tenDigits)), tenDigits)
-        var s = jsnums.toFixnum(n).toString().split(".");
-        s[1] = (s[1] || "").substring(0, d);
-        for (var i = s[1].length; i < d; i++)
-          s[1] += "0";
-        return thisRuntime.makeString(s[0] + "." + s[1]);
-      }
+      return thisRuntime.makeString(jsnums.toStringDigits(n, digits));
```
sorawee: @blerner here.

src/js/base/ffi-helpers.js
```diff
@@ -456,6 +456,8 @@ define(["js/runtime-util", "trove/lists", "trove/sets", "trove/option", "trove/e
 
         makeList: makeList,
         makeTreeSet: makeTreeSet,
+
+        isOption: runtime.getField(O, "is-Option"),
```
belph: Why is this not
```js
isOption: function(v) { return runtime.getField(O, "is-Option").app(v); }
```
(for consistency's sake)?

sorawee: Correct. You can click the unfold button under the line number, and you will see `isEither: runtime.getField(E, "is-Either"),`.

blerner: I think then the `isEither` line is wrong too...  And, I guess we never call it at all :)

sorawee: Fixed ;)

docs/written/builtin/numbers.scrbl
```diff
@@ -855,10 +855,11 @@ decimal.
 
 @examples{
 check:
-  num-to-string-digits(5432.1234, 2) is "5432.12"
-  num-to-string-digits(0.123456789, 2) is "0.12"
```
jpolitz: Why not have these two tests?

sorawee: `num-to-string-digits(2/3, 3) is "0.667"` basically shows the same thing. Since I include several more corner case examples, I feel that keeping those two lines will make the documentation too verbose.

jpolitz: Oh duh, thanks. I thought I was reading tests.

jpolitz:
I think overall this is fine to give a shot at merging; I think we'll have more comments from using it that we should process rather than doing a long review up front.

src/js/base/ffi-helpers.js
sorawee: @blerner @belph I know the reason why it shouldn't call `.app` now: so that it can be used in `ffi.cases`, because `ffi.cases` will automatically call `.app` for us.

blerner: That's not an acceptable reason, in the long term -- it feels like a type error to me, since all other functions are raw JS functions...  (I also don't know which came first, that `ffi.cases` used `app` for us, or the annotations were changed to not include the `app`.)  @jpolitz opinions which way to resolve this?

sorawee: Agree that all other functions are raw JS functions, so this seems to be inconsistent. However, most values that you would `cases` on would be constructed from `data` which normally would come from Pyret. I guess it is also awkward to write `.app` everytime you write `ffi.case`.

PR #748, opened 2016-06-27
Threading an error-callbacks object through all the js-numbers methods

### LINKED ISSUE #737: js-numbers needs a different way to throw exceptions in multi-runtime runs
It’s useful for there to only be one copy of js-numbers per JS VM, so that all runtimes can share number representations.  But this makes the `setThrowRuntimeError` method of js-numbers broken in the face of multiple runtimes, which can override this method and confuse error reporting.

I think the right thing to do is to simply not rely on js-numbers throwing Pyret exceptions, and include the logic to check and/or catch non-Pyret exceptions in places that use those functions.  I don’t think we want to instantiate js-nums on a per-runtime basis, because when we do write programs that rely on having different runtimes within the same JS VM, we want numbers to count as primitives that can be passed back and forth.

Another possible fix is to define a `NumberException` in js-numbers, throw it from js-numbers failure cases, and turn it into a runtime-specific `PyretException` as soon as it’s seen in an error case in a catch block in Pyret.  This would also give us a better opportunity to make number errors richer without entangling `js-numbers` with the runtime, since right now all the number errors are `messageException`s.

blerner:
This both gives more precise error message possibilities, and avoids having generative-error-structure problems when sharing the js-numbers module between multiple runtimes.

Ideally this fixes #737.

### DIFF
--- a/src/js/base/js-numbers.js
+++ b/src/js/base/js-numbers.js
@@ -446,23 +446,23 @@ define(function() {
       }
     },
     function(x, y, errbacks) {
-      if (equalsAnyZero(y)) {
+      if (equalsAnyZero(y, errbacks)) {
         errbacks.throwDivByZero('/: division by zero, ' + x + ' ' + y);
       }
       return x.divide(y, errbacks);
     },
     {
       isXSpecialCase: function(x, errbacks) {
-        return equalsAnyZero(x);
+        return equalsAnyZero(x, errbacks);
       },
       onXSpecialCase: function(x, y, errbacks) {
-        if (equalsAnyZero(y)) {
+        if (equalsAnyZero(y, errbacks)) {
           errbacks.throwDivByZero("/: division by zero, " + x + ' ' + y);
         }
         return 0;
       },
       isYSpecialCase: function(y, errbacks) {
-        return equalsAnyZero(y);
+        return equalsAnyZero(y, errbacks);
       },
       onYSpecialCase: function(x, y, errbacks) {
         errbacks.throwDivByZero("/: division by zero, " + x + ' ' + y);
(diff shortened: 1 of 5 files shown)


jpolitz:
Paging @ds26gte to see if he has any memory of particular things we should be concerned with here.

src/js/base/js-numbers.js
```diff
@@ -550,96 +534,96 @@ define(function() {
...
-      return lessThanOrEqual(err, absDelta)
+      return lessThanOrEqual(err, absDelta, errbacks)
     } else {
-      var errRatio = divide(err, abs(ratTv))
+      var errRatio = divide(err, abs(ratTv, errbacks), errbacks)
 
       if (deltaIsRough && delta.n === Number.MIN_VALUE) {
         if (argNumsAreRough && Math.abs(toRoughnum(errRatio).n) === Number.MIN_VALUE) {
-          throwRuntimeError('roughnum tolerance too small for meaningful comparison, ' +
+          errbacks.throwRelToleranceError('roughnum tolerance too small for meaningful comparison, ' +
```
jpolitz: I could not figure out how to write a test that would reach this line.  The `toRational` conversions and back aren't quite the identity, so it's hard to probe for exactly these values.

src/js/base/js-numbers.js
```diff
@@ -704,195 +688,195 @@ define(function() {
...
+  var sqr = function(x, errbacks) {
     return multiply(x, x);
   };
 
   // integerSqrt: pyretnum -> pyretnum
-  var integerSqrt = function(x) {
+  var integerSqrt = function(x, errbacks) {
     if (! isInteger(x)) {
-      throwRuntimeError('integer-sqrt: the argument ' + x.toString() +
+      errbacks.throwDomainError('integer-sqrt: the argument ' + x.toString() +
```
jpolitz: Not exposed in Pyret, so tough to test.

src/js/base/js-numbers.js
```diff
@@ -704,195 +688,195 @@ define(function() {
...
         return Math.floor(Math.sqrt(x));
       }
     }
-    return x.integerSqrt();
+    return x.integerSqrt(errbacks);
   };
 
   // gcd: pyretnum [pyretnum ...] -> pyretnum
-  var gcd = function(first, rest) {
+  var gcd = function(first, rest, errbacks) {
```
jpolitz: Not exposed in Pyret (?) so tough to test.

jpolitz:
There are several functions here (`remainder`, `gcd`, `integerSqrt`, and a few others) that aren't exposed to Pyret, and are tough to test from `test-numbers.arr`.

blerner:
For the non-exposed functions, either expose them (remainder, gcd, perhaps) or have them throw internal errors blaring "how'd you manage to get this to happen?!"

Issue #179, opened 2014-05-27
Add a display hook so that rich values can be printed


melloc:
At the moment, rich values like images cannot be printed to the console. For example, something like the following code should print an image:

``` pyret
  import image as i

  my-circle = i.circle(30, "solid", "red")
  print(my-circle)
```

jpolitz:
This works on horizon now naturally with print()

blerner:
Um, no, it doesn't.  Print always produces a string output, so this produces the string `"<image (60x60)>"`.  But evaluating `my-circle` in the repl does render the circle.  I don't think we want `print` to produce rich output, though we may want another, new function that does.  (There are other issues filed that address similar things.)

schanzer:
Currently on horizon:
![image](https://cloud.githubusercontent.com/assets/77126/16839523/82b17582-499d-11e6-90f5-c91bf0ad6b52.png)

blerner:
This is expected behavior, to a point. If your program were `print("Hello")`, you'd see `Hello` followed by `"Hello"`.  The first one is the `tostring`-form of the value you're printing, and the second is the traced result of the return value of `print`, which is closer to `torepr` in behavior.  That's why you see `<internal value>` followed by a pretty picture: it's _not_ because `print` now knows how to rich-render pictures.

As for why it's `<internal value>` now instead of `<image (60x60)>`, we made that change when we moved the image library from pyret-lang into CPO...but I'm not sure whether I like the result.  In any case, the _original_ motivation of this bug report is still valid: there's no function you can use from Pyret that will rich-render values into the REPL.

jpolitz:
Yup, this is what `spy` is for, which got pushed for this release since it wasn't necessary for any summer users.  Marking unfixed and leaving open.

blerner:
Close now that spies have merged?

jpolitz:
Hooray for closing 3yo bugs! Let's leave it horizon-fixed until it's on master.

schanzer:
Closing now that this is on CPO

PR #844, opened 2016-08-23
Spies

blerner:
This introduces support for `spy`, which looks like

```
spy <expr-title>:
  name,
  name: expr
  ...
end
```

This will render as a spyblock, with the `<expr-title>` as a name for the block.  The contents of the block will render as Each `name: expr` clause will rich-render the expression, and label it with name.  `name` clauses are shorthand for `name: name`, i.e., the identifier is eponymous with its value.


jpolitz:
I think we should not include 72caa5b1a607081b1040fdcf635f9082c814d9f2, because it will be surprising since we didn't announce it earlier.

jpolitz:
This is a gratifyingly small PR.  I think this is a great first run to try out.

src/arr/trove/ast.arr
```diff
@@ -1166,10 +1166,37 @@ data TableRow:
     method tosource(self):
       PP.flow([list: str-rowcolon,
           PP.flow-map(PP.commabreak, _.tosource(), self.elems)])
+  | s-spy-block(l :: Loc, message :: Option<Expr>, contents :: List<SpyField>) with:
```
jpolitz: Pretty sure this isn't a TableRow.

blerner:
I really think we should include the 72caa5b change, since there are likely not to be many people yet relying on it, and it'll be easier to change now as we introduce the intended replacement, than if we let it sit another year.

shriram:
PLEASE do NOT mess with how `print` works. I actually use it in class. My students depend on it. (Also, some of my students think it's the coolest thing ever and wonder why other languages don't do the same thing.) I do not want this to ever change in Pyret.

shriram:
Is the <expr-title> part optional? Otherwise this looks awfully heavy-weight.

blerner:
What the...?  We _*never*_ agreed that print should continue to return its value; it was always a hack (you explicitly called it a hack) that was the whole reason we talked about introducing spy in the first place.  This isn't the right venue to rehash the entire design discussion, though.

Yes, the name is optional.

shriram:
It may have been a hack, but I've come to see it as a pretty useful feature. `spy` does not at all replace its value. And I agree, this isn't the place to discuss this. We should just leave `print` alone for now. I see no harm in doing so.

schanzer:
Linking this to #179

jpolitz:
@blerner should we merge this without https://github.com/brownplt/pyret-lang/commit/72caa5b1a607081b1040fdcf635f9082c814d9f2?  It seems worth getting some progress made on it, and we keep wanting it.

src/js/base/runtime.js
```diff
@@ -5049,6 +5050,79 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
...
+    function toReprArray(vals, reprMethod) {
+      // vals should be an array of {name: String, val: PyretValue, [method: ReprMethod]} objects
+      // return is an array of either {name: String, val: rendered} objects (on success)
+      // or {name: String, exn: exn} objects (on failure)
+      // but all will run.
+      // Each item can specify its own reprmethod, or the default one can be used
+      var results = [];
+      return thisRuntime.safeCall(function() {
+        return thisRuntime.eachLoop(makeFunction(function(i) {
+          thisRuntime.pauseStack(function(restarter) {
```
jpolitz: This needs a `return`

PR #851, opened 2016-08-27
[SD] Implements and tests linear regression with 2 variables

SDooman:
This is missing treatment of the typing, and I'm not sure if this naming matches desired Pyret conventions.

### DIFF
--- a/tests/pyret/tests/test-lists.arr
+++ b/tests/pyret/tests/test-lists.arr
@@ -253,11 +253,11 @@ check "linear regression":
   lists.lin-reg-2V([list: 0], [list: ]) raises "lists must have equal lengths"
   lists.lin-reg-2V([list: 0], [list: 1]) raises "lists must have at least 2 elements each"
 
-  f = lin-reg-2V([list: 0, 1, 2], [list: -5,- 3,- 1])
+  f = lin-reg-2V([list: 0, 1, 2], [list: -5, -3, -1])
   f(0) is -5
   f(2.5) is 0
 
-  g = lin-reg-2V([list: -3, 1, 2, 4], [list: ~4, ~1.5, ~, -2.2])
+  g = lin-reg-2V([list: -3, 1, 2, 4], [list: ~4, ~1.5, ~0, -2.2])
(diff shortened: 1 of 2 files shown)


SDooman:
@blerner I'm inclined to agree returning a lambda is not the best format.  Returning some object like you're describing could include other relevant information about the line of best fit like R^2 value etc.  Figuring out exactly what this return value should include necessitates design conversations, but I agree with @sorawee that linear regression, the `stdev` function & friends should be in some sort of `statistics` module.  What are your thoughts @shriram?

shriram:
I agree, these should go into a new `statistics` module. I really like @sorawee's design criterion that `list` is primarily for polymorphic list functions. Besides, this also improves discoverability: someone wondering “Is there a way to compute standard deviations?” is more likely to say “aha” when they see a library entry called `statistics` than when they see one called `lists`. Besides, we may offer stddev over other datatypes later, and `statistics` would be the place to unify all of them and make them coherent.

tests/pyret/tests/test-lists.arr
```diff
@@ -248,3 +248,17 @@ check "numeric helpers":
   lists.distinct([list: ~1, ~2, ~3]) is-roughly [list: ~1, ~2, ~3]
 
 end
+
+check "linear regression":
+  lists.lin-reg-2V([list: 0], [list: ]) raises "lists must have equal lengths"
+  lists.lin-reg-2V([list: 0], [list: 1]) raises "lists must have at least 2 elements each"
+
+  f = lin-reg-2V([list: 0, 1, 2], [list: -5, -3, -1])
```
sorawee: It seems you need to prefix this with `lists.` ;)

tests/pyret/tests/test-lists.arr
```diff
@@ -248,3 +248,17 @@ check "numeric helpers":
...
+  f(0) is -5
+  f(2.5) is 0
+
+  g = lin-reg-2V([list: -3, 1, 2, 4], [list: ~4, ~1.5, ~0, -2.2])
```
sorawee: Ditto.

PR #857, opened 2016-09-01
[SD] Initialize statistics module (pre-linear-regression)

SDooman:
This is the beginning of transferring the numerical operations into a more appropriate Statistics Module for Pyret.  I came across a bug with the compiler:  When running `make test`, the tests in the file `/tests/pyret/tests/test-statistics.arr` do not compile because nothing from the Lists module is defined.  Shouldn't everything from `src/arr/trove/lists.arr` be included automatically @blerner @jpolitz?


src/arr/trove/statistics.arr
```diff
@@ -0,0 +1,94 @@
+#lang pyret/library
+
+provide *
+provide-types *
+import global as _
+import option as O
```
sorawee: I think you need `include lists` here so that you can use all list functions

jpolitz:
Sorry for not following up sooner.  This isn't a bug with the compiler.  It's just inconvenient, I'm afraid.

Only a subset of the `list` functions are included by default; I think we're leaning towards _none_ of them being included by default in the future.  The right thing to do is to write `lists.<fun-name>` for the functions from the lists library you want to use.  `include lists` won't work, because it will cause name clashes with some of the default includes.

jpolitz:
A bit more refined answer:  for the module being added in `trove`, you'll need to add a line like `include lists as lists` to the top.  Implementing a library in `trove` is a little bit special, because it is where we define new special builtins that can be imported by default by other modules.  There are _no_ default imports in `trove` modules to avoid cycles.  So _no_ list functions (nor the `List` annotation) are available in that file.

In the _test_ file, the comment I made about only some list functions being available is accurate.  The list module is imported by default (as `lists`), so you don't need to add an import line there.

SDooman:
@jpolitz 

adding `include lists as lists` gives the following error

```
Compiling worklist...There were 0 potential parses.
Parse failed, next token is ('AS "as") at builtin://statistics, 10:14-10:16
The run ended in error:
Pyret didn't understand your program around builtin://statistics:10:14-10:16 You may need to add or remove some text to fix your program. Look carefully before the highlighted text. Is there a missing colon (`:`), comma (`,`), string marker (`"`), or keyword? Is there something there that shouldn’t be?
```

I see that many other files in `trove` have this line:

`import lists as lists`

However, when that's added to statistics, the tests still fail for the same reasons as the original PR.  Could it be that I need to add `src/arr/trove/statistics.arr` to some path that it's not a part of?

jpolitz:
Typo, meant `import`, sorry.

jpolitz:
I just pushed a fix that seems to work for me; it was just adding the import to `statistics`, and then the tests ran correctly.

PR #864, opened 2016-09-06
[SD] implements simple linear regression with tests

SDooman:
Defines a new data type, StatModel.  This will be the data type for any statistical model (regressions) we choose to implement.  For now, the only regression feature that has been requested is simple least squares regression (1 independent & 1 dependent variable).

The naming conventions for methods of the `simple-linear-model` comes from [this](https://en.wikipedia.org/wiki/Linear_predictor_function) article.

### DIFF
--- a/src/arr/trove/statistics.arr
+++ b/src/arr/trove/statistics.arr
@@ -17,7 +17,30 @@ is-empty = lists.is-empty
 fold = lists.fold
 length = lists.length
 member3 = lists.member3
+map = lists.map
+map2 = lists.map2
 
+# A StatModel is a data type for representing
+# statistical models.  We create this type so that
+# users can extract, coefficients and meta-data from
+# models, as well as apply the model.
+#
+# This could be extended to include general linear 
+# regression, exponential and logistic regression, etc.
+
+data StatModel:
+  | simple-linear-model(alpha :: Number, beta :: Number) with:
+
+    method predictor(self :: StatModel) -> (Number -> Number):
+      doc: "the function predicting the value of a dependent variable"
+      lam(X): (self.beta * X) + self.alpha end
+    end,
+
+    method apply(self :: StatModel, l :: List<Number>) -> List<Number>:
+      doc: "applys the predictor function to a sample set of the independent variable"
+      lists.map(self.predictor(), l)
+    end
+end
 
 fun sum(l :: List<Number>) -> Number:
 	lists.fold(lam(a,b): a + b end, 0, l)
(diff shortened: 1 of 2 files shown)


jpolitz:
@shriram, you should have a look at this; it relies somewhat on methods, which we've been discussing as a not-so-great way to expose things to students.

shriram:
Good point, but for this particular interface it's really not clear how to do it much  more elegantly. I think @SDooman's original interface was more functional but @blerner, if I remember right, suggested switching it to this. Since this isn't a "beginner" function anyway, maybe we let it slide for now until we have a more coherent story overall?

blerner:
I didn't, quite -- I said I was unhappy that the earlier version returned a lambda, and not the (alpha, beta) parameters themselves (as an object or tuple, was my initial thought).  I'm not wedded to having `predictor` and `apply` be methods on the model, in fact I'm almost perfectly unbiased towards or against them in this setting.

shriram:
My sense is the way this is going to work, there's going to be a "thing" anyway. Whether we have functions on that thing or methods on them may to some extent on what people expect. For instance, if these methods are traditionally thought of as functions, then as methods they'd have lost one argument and that would be bad since they wouldn't match convention. But my sense is that the object here is ambient state that isn't made explicit in math (correct, @SDooman?), in which case the method version is actually better.

SDooman:
I'm assuming we agree that there should be a "thing" (the StatModel data type) to represent a model.  In that case we could either have `predictor` and `apply` be functions that consume a StatModel, or have `predictor` and `apply` be methods.  It's important to note that 

One advantage to having these be methods of a particular kind of StatModel (ex: linear regression with 1 independent and 1 dependent variable) is that we know the exact signature of the predictor.

Consider a general `predictor` function that would consume a StatModel and pattern match, returning the predictor function:

```
fun predictor(sm :: StatModel):
  cases(StatModel) sm:
    | simple-linear-model(alpha, beta) => lam(X): (sm.alpha * X) + sm.beta end
    .
    .
    .
  end
end
```

Since different variants of StatModel will be predicting values of dependent variables in different spaces, these predictors will have different signatures.  If that's acceptable, then defining `predictor` in this way would be more consistent with Pyret's goals (if we're shying away from methods).

However, another consideration is that the coefficients of a predictor function aren't the only attributes of models (to use Pyret terms, they shouldn't necessarily be the only members of StatModel variants).  For instance, consider [R^2 values](https://en.wikipedia.org/wiki/Coefficient_of_determination).  To follow the same pattern of having a general `predictor` function, we should also have some function `determination-coefficient`.  I don't have a criticism of this particular design right now, it's just worth considering that there may be attributes that some variants have, and some don't have, and the behavior of a similar function to `determination-coefficient` when consuming a StatModel without the appropriate member is worth thinking about.

IMO, treating `predictor` and `apply` as methods instead of functions will lead to clearer documentation due to the method signature ambiguity discussed above.  A teacher who wants her/his class to use simple linear regression, may have an easier time understanding what code needs to be written to accomplish this.

jpolitz:
I'm convinced that this is a fine start.

PR #878, opened 2016-09-30
[SD] implemented R Squared for linear regression

SDooman:
Includes the member `r-sqr` which represents the coefficient of correlation (measure of goodness of fit) for a model


jpolitz:
Looks good.

PR #897, opened 2016-10-06
Use short names for runtime methods

puppyofkosh:


jpolitz:
@blerner – this seems like a clear win (with opportunities for hitting more stuff in the future).  Any objections to merging?

blerner:
No objections, though I would like a command-line switch for readability, so that when we're compiling individual modules for testing purposes, we can still read the output code.

PR #900, opened 2016-10-13
[SD] implement chi squared correlation calculation

SDooman:


sorawee:
Do we need this branch around? Or the code was merged by other commits already?

PR #902, opened 2016-10-17
Register Allocation

belph:
This branch extends `js-dag-utils` to use a linear register allocator to reduce the sizes of the lexical scopes of our compiled functions (which is nice, given our ample usage of closures).

As an example of how much this save us, I noticed the reduction of one block from 1003 variables down to 340 variables.

More profiling/discussion can take place here or in `#register-allocation` on Slack; but this is working in its current state.


jpolitz:
I'm closing this (it's not like we lose the code/branch) as stale, because we never really demonstrated a win with it. Definitely worth revisiting at some point, but I think we'll lift the ideas into a new PR, not merge this one.

Issue #678, opened 2016-03-18
Improve loc / error message for auto-imported libs


sorawee:
```
a = 1
list = 1
```

when running, i get the following error:

```
It looks like you've defined the name list twice, at
definitions: line 1, column 0 and
definitions: line 2, column 0.
You need to pick a different name for one of them.
```

where `definitions: line 1, column 0` spans the whole code (from line 1 to line 2), and `definitions: line 2, column 0` spans only over the word `list`.

First, I didn't _defined the name list twice_. I did it only once. Second, the location of `definitions: line 1, column 0` is just weird.

Would be nice to have an error message like `raw-array`:

```
The name raw-array is already defined. You need to pick a different name for raw-array at definitions: xxx
```

though I understand that this might be difficult because built-in functions are not the same as functions from libs.

blerner:
The problem appears to be in resolve-scope, and more generally with how modules export values.  I haven't traced all the details yet, but basically we assign "the entire module" as the srcloc of ambient values.  And fixing it is tricky, because our exported information doesn't provide srclocs of where values came from.  (This will come up again, I suspect, when handling include statements...)

Note that the message you see is in fact the same error structure for both list and raw-array.  It's just that the rendering of it checks whether the initial srcloc is a builtin or not, and chooses which message to render as a result.  Which means once we get builtin locations for the ambient values, this will probably go away.

blerner:
The commit mentioned above fixes the initial example, but the following program still fails badly:
```
include world

big-bang = 5
```
gives
```
It looks like you've defined the name big-bang twice, at
definitions: line 1, column 0 and
definitions: line 3, column 0.
You need to pick a different name for one of them.
```
where the first location spans the entire document.  I wonder whether we shouldn't have a third kind of srcloc, an `included`, that highlights the location of the `include` line that brought the name into scope.  Neither `srcloc` nor `builtin` quite have the right implication here: if I include a file that I've written, then a duplicate name is neither builtin nor defined twice in this file.  @jpolitz thoughts?

sorawee:
Related:

```
a = 1
lists = 3
```

fails like the original program posted above.

schanzer:
@sorawee and @blerner The error is a lot better now. OK to close?

blerner:
No.  The world/big-bang error still occurs on #horizon, and that's a problem.

schanzer:
@blerner the world/big-bang error seems much improved:
[...]

Issue #366, opened 2014-11-04
Pyret does not check if case variants are valid.


Skylion007:
This does not throw an exception. Rather, it only fails when I actually pass an mt argument in which case it says no cases met. I may be misunderstanding the parsing of cases but doesn't (BHeap) mean that only variants of BHeap should work? After all, if I specify node as simply node (without parameters) then it throws a compiling error, but it allows empty to be a case for some reason despite the fact that it's not even the write type. I might just be misunderstanding the Pyret syntax though.

data BHeap:
  | mt
  | node(value :: Number, left :: BHeap, right :: BHeap)
end

fun is-balanced(my-tree :: BHeap) -> Boolean:
  doc: "Determines if a tree is balanced (not actual implementation)"
  cases (BHeap) my-tree:
    | empty => true
    | node(v, l ,r) => false
  end
end

This is the error message return instead when I try to pass mt into is-balanced.

"No cases matched in the cases expression at definitions: line 14, column 2 for the value:

mt"

schanzer:
This appears to be fixed, now that we have typed Pyret. Is the intention that untyped Pyret should still  catch this at compilation time?

blerner:
More or less, yes.  Data types should have some static information available, that they don't provide quite yet, that should be exploited in the compilation of `cases` expressions to help in this situation.

jswrenn:
> More or less, yes.

Some of the challenges of this are noted in #757.

jpolitz:
Closed as dup of https://github.com/brownplt/pyret-lang/issues/983#issuecomment-280455865, which is a superset of this issue.

PR #915, opened 2016-10-25
Start building up better data structures for static cross-module info

### LINKED ISSUE #288: Repl can't tell the difference between identifiers and variables
At the REPL:

> var foo = 5
> foo := 5

Will report that foo was not defined as a variable.  This is because the repl exports don't distinguish correctly, so this is being statically flagged as an error before running the code at the repl.  The fix is to add more information in CompileEnv bindings so the compiler allows variables in its initial scope, and then have repl-support and repl-lib correctly add variables vs. identifiers to the environment.

jpolitz:
(@blerner – requesting review; I would actually like to push this forward if possible, since it fixes a longstanding issue with variables at the REPL, while also working with data structures required for flatness analysis.)

Most directly, this commit fixes #288.

It also gives a place to start hanging more static information off of module
exports.  The information is all "there" but needs to be attached to particular
bindings in different phases.

This should eventually help with statically checking cases (#366) and attaching
cross-module locations to bindings, or associating bindings with import lines
(#678).  The `ValueExport` type also captures the flatness analysis information
that @puppyofkosh is working on.

I'm not sure that the `ValueExport` datatype is in its final form, and how many
variants it will eventually have.  But it's a starting point that makes
variables work, and demonstrated how to get more than just type information
propagated around across modules.  In particular, it maybe should be unified
with the the `ScopeBinding` type, since they reflect similar information, just
cross- vs intra-module.

This commit also uses the "defined-values" field instead of the object-shaped
"values" in "provide-plus-types".  This is because, with variables exported,
it's impossible to represent an exported variable in a Pyret object, since a
Pyret object's field could only hold the value of the variable.  Here, the cell
itself is exported.

I think this is actually a surprisingly expressive feature for a module system
to offer, though the alternative is to rework how the REPL works entirely to
not use module imports and exports as the semantics of binding between entries.


src/arr/compiler/compile-structs.arr
```diff
@@ -190,10 +210,9 @@ fun provides-from-raw-provides(uri, raw):
   values = raw.values
   vdict = for fold(vdict from SD.make-string-dict(), v from raw.values):
     if is-string(v) block:
-      vdict.set(v, t-top)
+      vdict.set(v, v-just-type(t-top))
     else:
-      #print("\n\nYOOOOOOOOOOOOOOOOOOOOOOO: " + tostring(v.name))
```
jpolitz: hi

shriram:
Improving hanging static information off exports will also be relevant to
Ian Boros as he conveys information like call depth. We talked today about
maybe doing a more general abstract interpretation framework for lots of
other little optimizations.

jpolitz:
cca25044213b8e5efa2ea1658a24a39bdaab1f64 had some changes to the exports of `sets`, because this commit is the last step in truly disconnecting `provide`'s semantics from "create an arbitrary Pyret value and shove it across the module boundary." Instead, `provide` can (currently) _only_ make available names that are defined at the top level of a module.  Their values will be whatever the value of those names are in scope at the very end of the module, hence the new uses of `shadow` at the end of `sets.arr`.

The way forward here is to enrich `provide` syntax with primitives like Racket's `rename-out`, `prefix-out`, etc, to control these names reasonably.  But getting away from "just make a value" is the first step, since now Pyret is developing a really good model for cross-module bindings in a first-order way, something that's much harder to reconstruct from an object value.

jpolitz:
Also, I think this is ready to merge with this commit once tests pass, modulo any cleanup.  There are no "dangling" parts of the implementation.  Variable exports work for both serialized and in-memory modules, and at the REPL, which is a complete implementation of static/dynamic information that isn't just vanilla values and types, so we can build on this going forward.

It also makes variables work at the REPL, closing an issue from 2014 that doesn't likely have a simpler resolution :-)

PR #916, opened 2016-10-25
Switched over to immutable-js string-dict implementation

mkolosick:
Uses the optimized version of string-dict for immutable string dicts.

### DIFF
--- a/src/js/trove/string-dict.js
+++ b/src/js/trove/string-dict.js
@@ -87,28 +87,607 @@
       return get(brand, "test").app(val);
     }
 
-    // Prepend a space to avoid conflicting with built-in names
-    function internalKey(s) {
-      return " " + s;
-    }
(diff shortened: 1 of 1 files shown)


jpolitz:
From the failed build:

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

I wonder if this implementation has some bad performance characteristics on existing code?

mkolosick:
I will look into that.

jpolitz:
It actually looks like it might be taking longer on type-check tests, which is really odd (though I actually don't know).

Those lines of output at the bottom print during the general language/lib tests, and type tests run after.  Compare to this build output, where the type tests are enumerated afterwards:

https://travis-ci.org/brownplt/pyret-lang/builds/170315903

mkolosick:
There appears to be a memory leak. I will continue to investigate.

PR #917, opened 2016-10-25
Fixed list shuffling

gregthegeek:
Previously, an element could not be shuffled to its original spot. This meant lists of length 2 were effectively not shuffled, just swapped.


jpolitz:
Thanks!  I added a test to check this – does that jive with what you expect?

jpolitz:
Test at https://github.com/brownplt/pyret-lang/commit/6003dd5abe9abc2e4c9351faf51e98a66066d251

gregthegeek:
It does, thanks!

Issue #921, opened 2016-10-26
Median function does not work on even-numbered lists


les-bres:
When you use the built-in median function from statistics, it does not perform any averaging given an even-numbered list.  For example:
`median([list: 1, 2, 3, 4])`
outputs 3 rather than 2.5

schanzer:
Now that this is merged, can we close the issue?

schanzer:
pinging @SDooman

schanzer:
Now it's also fixed on CPO. Close?

schanzer:
Now also fixed on CPO. OK to close?

PR #922, opened 2016-10-26
Median functionality fix

SDooman:
- Tests and fixes issue documented [here](https://github.com/brownplt/pyret-lang/issues/921)
 - Includes more thorough testing of standard deviation

### DIFF
--- a/src/arr/trove/statistics.arr
+++ b/src/arr/trove/statistics.arr
@@ -48,11 +48,22 @@ end
 
 fun median(l :: L.List):
   doc: "returns the median element of the list"
+
+  fun is-odd(n :: Number) -> Boolean:
+    num-modulo(n, 2) == 1
+  end
+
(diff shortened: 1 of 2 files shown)


shriram:
Median is well-defined for any orderable entity: order it and find the
middle one. But averaging the two middle elements assumes division, which
is a much stronger expectation (basically, numbers). Furthermore, it has
the property that the median is no longer an element of the collection.
Does it make more sense to have multiple median operators, one of which
does averaging but two more that provide the "left middle" and "right
middle" in the case of even numbers of elements?

jpolitz:
I almost gave that response to the issue – that `median` was _a_ perfectly well-defined function as-is (and I never asked @SDooman which one he meant originally).  I just assumed we wanted this spec since I'm happy with the decisions that have been made with this library so far.

Up to @SDooman, IMO.

Issue #914, opened 2016-10-24
type-inference considered unusable


shriram:
Below is a 69-line program.

Running it consistently takes about three seconds. (That's really long, but it's at least usable.)

Type-checking it consistently takes over 38 seconds. (That's 38 seconds on second and third runs; the first one, without a warm cache, is longer!)

That's a 35-second type-checking overhead over 69 lines (of which about half are some combination of blank, `end`, `check`, etc.). So, about a second a line.

I was curious whether adding return type annotations (there are none below) would make a difference, so I added them everywhere. No difference. I added annotations on the `rec` lines. Also no difference.

I commented out the three `check` blocks. Now we were down to 8-12 seconds. That's a LOT better. I mean, not usable (at least 5-6 seconds to type-check about 45 lines of code), but almost usable in a small class demo.

I went back and removed all the `rec` and return type annotations. Still 9-11 seconds. So that's still not the source of the problem.

@mkolosick, I don't see any possibility of writing any papers about inference when it's literally impossible to use on even trivial programs.

```
data Stream<T>:
  | lz-link(h :: T, t :: ( -> Stream<T>))
end

fun lz-first<T>(s :: Stream<T>):
  cases (Stream) s:
    | lz-link(h, t) => h
  end
end

fun lz-rest<T>(s :: Stream<T>):
  cases (Stream) s:
    | lz-link(h, t) => t()
  end
end

rec ones = lz-link(1, {(): ones})

fun nats-from(n :: Number):
  lz-link(n, {(): nats-from(n + 1)})
end
nats = nats-from(0)

fun take<T>(n :: Number, s :: Stream<T>):
  ask:
    | n == 0 then: empty
    | n > 0  then:
      link(lz-first(s), take(n - 1, lz-rest(s)))
  end
end

check:
  take(5, ones) is [list: 1, 1, 1, 1, 1]
  take(5, ones) is map({(_): 1}, range(0, 5))
  take(5, nats) is [list: 0, 1, 2, 3, 4]
end

fun lz-map<A, B>(f :: (A -> B), s :: Stream<A>):
  lz-link(
    f(lz-first(s)),
    {(): lz-map(f, lz-rest(s))})
end

fun lz-map2<A, B, C>(
    f :: (A, B -> C), 
    s1 :: Stream<A>,
    s2 :: Stream<B>):
  lz-link(
    f(lz-first(s1), lz-first(s2)),
[...]

PR #931, opened 2016-10-31
Optimized string dict

mkolosick:
This fixes several issues.

- issue #914: replaces string-dict with the optimized version
- The issue in the previous pull request was an accidental dependency on the ordering of the previous string-dict implementation. This has been remedied.
- Additionally, a type checker bug with polymorphic functions with unannotated return types has been fixed. For now, polymorphic functions require annotations on return types. This restriction will possibly be removed later.


jpolitz:
Looks good to me.

I don't understand yet why tests are slightly slower with this.

In some simple benchmarks, this is _obviously_ faster.

Making a note to myself to try creating many small dictionaries, to see if the initial construction cost is slightly higher with this version.

PR #932, opened 2016-11-02
Server

jpolitz:
I'd like to merge this in – it doesn't take away anything (we can change the `Makefile` back to how it was and everything will still work that way, too).

It nicely speeds up rebuilding phaseA by the amount of time it takes phase0 to start up, and also enables speedups when developing in CPO, which can use a server to save a bunch of time when recompiling the standalones there.

I think it's good enough to use and start getting feedback on/figure out where it breaks.

@blerner, what do you think?


blerner:
I'm fine with this, but have one extra-lingual concern: this creates a server.pid file to keep track of where the server is running, yes?  What happens if say my laptop crashes (as it sometimes/frequently feels like doing), so that the process is gone but the pid file remains?  Is node smart enough to handle that case?

jpolitz:
On restart, it'll complain that a server is running when it isn't, and there's some manual cleanup.  `make clean-servers` should do it for many cases, but we can keep improving that to handle other situations that come up.  We should add a `pyret --clean` option in addition to `pyret --shutdown` to handle the restart case, which is a good one to think of.

I think for a few rounds we'll have some issues like this to iron out (attaching to existing servers from other directories, etc).

blerner:
Just to make sure I understand, that means that after a crash, when I next try to say "make foobar.jarr", it'll complain about a missing server process, instead of auto-restarting it?  This seems like the sort of thing I'd consistently forget how to cleanup myself, so it'd be great to automate it.  Not a dealbreaker to merge this as it stands, but it should definitely be a short-term feature request if it doesn't exist yet.

jpolitz:
Yeah, I will have my ears open for quick fixes like this.

PR #941, opened 2016-11-27
Type checker fixes

mkolosick:
This makes some correctness fixes and also has a few optimizations.

- Letrec's are now reconstructed in the correct order
- Loose check blocks are ignored by the type checker as they cause a massive code blowup
- A number of other optimizations


src/arr/compiler/type-check-structs.arr
```diff
@@ -829,8 +829,10 @@ fun solve-helper-examples(system :: ConstraintSystem, solution :: ConstraintSolu
   cases(ConstraintSystem) system:
     | no-constraints => fold-result({system; solution}, context)
     | constraint-system(variables, constraints, refinement-constraints, field-constraints, example-types, next-system) =>
-      foldr-fold-result(lam(existential-key, shadow context, {shadow system; shadow solution}):
+      foldr-fold-result(lam(existential-key, shadow context, {shadow system; shadow solution}) block:
         {existential; _; fun-examples; checking-fun} = example-types.get-value(existential-key)
+        print("existential: " + tostring(existential) + "\n")
+        print("fun-examples: " + tostring(fun-examples) + "\n")
```
jswrenn: On CPO, won't these `print` calls result in those messages being printed to the REPL?

mkolosick: Yes, I thought I'd removed all of them but apparently I missed two.

mkolosick: Wait, where are you finding those? They seemed to have been removed in 6f2614b

jswrenn: Nope, you're right. The perils of viewing-by-commit.

PR #951, opened 2016-12-15
[SD] Mode tests and implementation

SDooman:
Adds two functions:  `mode` and `modes`.  
 - `mode` will return a single number representing the mode from a list of numbers (if there are   multiple modes in the list, there is no guarantee which mode this function will return.
 - `modes` returns a list of numbers containing every and all modes in the list.

NOTE:  Originally the name of `modes` was `-modes`, but Pyret does not allow function definition with `-` as the first character.

EDIT:  This needs to be rewritten without the use of string-dicts.  Please ignore for the time being

### DIFF
--- a/src/arr/trove/statistics.arr
+++ b/src/arr/trove/statistics.arr
@@ -9,6 +9,7 @@ import equality as equality
 import valueskeleton as VS
 import lists as L
 import math as math
+import string-dict as sd
 
 # A StatModel is a data type for representing
 # statistical models.  We create this type so that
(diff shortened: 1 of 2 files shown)


src/arr/trove/statistics.arr
```diff
@@ -67,6 +68,80 @@ fun median(l :: L.List):
...
+          prev
+        end
+      end
+      
+      m-list = L.fold(aggregate-modes, [L.list: first], rest)
+      L.map(force-unwrap-string, m-list).reverse()
+  end
+end
+
+fun mode(lst :: L.List<Number>):
```
jswrenn: Can you add return type annotation to this?

src/arr/trove/statistics.arr
```diff
@@ -67,6 +68,80 @@ fun median(l :: L.List):
...
+  doc: "Gives the mode for a list of numbers.  If multiple modes exist in list, may return any of them"
+  count = num-counts(lst)
+  numbers = count.keys().to-list()
+  
+  cases (L.List) numbers:
+    | empty => raise("List is empty")
```
jswrenn: Can you document this failure condition in the `doc` string?

src/arr/trove/statistics.arr
```diff
@@ -67,6 +68,80 @@ fun median(l :: L.List):
...
+end
+
+fun force-unwrap-string(s :: String) -> Number:
+  cases (O.Option) string-to-number(s):
+    | none => raise("Mode does not appear in list") # Mode counts construction prevents this 
+    | some(v) => v
+  end
+end
+
+fun modes(lst :: L.List<Number>) -> L.List<Number>:
```
jswrenn: Why is the return type `List` instead of `Set`?

src/arr/trove/statistics.arr
jswrenn: Rather than raise an exception (which cannot be caught), does it make more sense to return an `Option<Number>`?

src/arr/trove/statistics.arr
SDooman: @jswrenn `Set` can't contain rough-nums, so even though all of the elements will be distinct, I think it needs to be `List`

jpolitz: > `Set` can't contain rough-nums

Huh.  Interesting.  More effects of equality being inextricably linked with some datatypes.

This could be particularly baffling for a user, as sets seem like they ought to be nicely parametric, but have this lurking dynamic equality error that can happen.

Things like this make me wonder if `roughly-equal` is a good default for more contexts.

src/arr/trove/statistics.arr
SDooman: @jswrenn Yes this does make sense!  There were issues building `horizon` related tuples that were resolved after this branch was created, so I incorporated your suggested changes (aside from the `Set` vs. `List` change) here: https://github.com/brownplt/pyret-lang/pull/959

blerner:
@SDooman do we need to keep this branch around any more for anything, or have you migrated the code here to other pull requests?

sorawee:
Ping @SDooman

PR #959, opened 2016-12-30
Mode and opts

SDooman:
- Adds and tests `mode` and `modes` to statistics library.  These functions consume `List<Number>` and return the most commonly occurring element/elements from the input, respectively.
 - `mode` returns an `Option<Number>`, instead or raising an error on empty inputs.  If everyone agrees with this change, I'm going to modify the `mean`, `median` and `stdev` functions & documentation to have this same signature.
 - `modes` returns a `List`, rather than a `Set`, because the input may contain rough-nums, and the built in `Set` structure cannot contain rough-nums.

### DIFF
--- a/src/arr/trove/statistics.arr
+++ b/src/arr/trove/statistics.arr
@@ -46,7 +46,7 @@ fun mean(l :: L.List<Number>) -> Number:
   end
 end
 
-fun median(l :: L.List):
+fun median(l :: L.List) -> Number:
   doc: "returns the median element of the list"
 
   fun is-odd(n :: Number) -> Boolean:
--- a/src/arr/trove/statistics.arr
+++ b/src/arr/trove/statistics.arr
@@ -67,6 +67,69 @@ fun median(l :: L.List) -> Number:
   end
 end
 
+fun modes(l :: L.List) -> L.List<Number>:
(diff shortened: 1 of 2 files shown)


jpolitz:
I want to register an argument for error on empty as better than `Option`, sorry I missed this discussion earlier.  Using `Option` complicates the common case to accommodate an exceptional case that, in practice, I think we will regret.

The ergonomics of `Option` add friction for beginners – I don't think we have a great story for what to do if you end up with a bunch of `Option`s in a table column, for instance.  _I_ know what to do, but we don't discuss `Option` much until the ugrad _CS_ level.

If we are going to push pedagogy for `Option` earlier, then this becomes much more palatable.

shriram:
I agree. If we find a need for it, we can later offer -option versions of
the same functions that return an option instead (just like we have a
three-valued equality primitive for cases when people don't want to halt on
error). But the default design should be simple return types.

SDooman:
@jpolitz ping on this, I reverted the return types back to `Number`.  What's the protocol for when I've made changes & everything is good.  I shouldn't be the one merging, right?  Do I need to update docs before this gets merged?  If so I'll do that today.

jpolitz:
Updating docs as well would be good, but this is ready to merge.  Thanks.

PR #961, opened 2017-01-05
Fix the --run flag

### LINKED ISSUE #957: --run flag is broken
```
node build/phase0/pyret.jarr --run tools/benchmark/auto-report-programs/0_empty.arr
```

```
The run ended in error:
Evaluating the field lookup expression at file:///Users/joe/src/pyret-lang/src/arr/compiler/cli-module-loader.arr:312:60-312:71 errored. The left side was an object that did not have a field named `js-ast`:right({js-ast: <redacted: too big>})

Pyret stack:
  file:///Users/joe/src/pyret-lang/src/arr/compiler/cli-module-loader.arr: line 312, column 60
  file:///Users/joe/src/pyret-lang/src/arr/compiler/pyret.arr: line 175, column 4
```

Currently working on a fix: it appears that the `run` function in `cli-module-loader.arr` needs to handle the `Either`ness of the output of `build-program`.

awstlaur:
- Fix #957
- Add `phaseX` scripts to abstract over the functionality (they all symlink to `run-phase` which takes care of things)

### DIFF
--- a/src/js/base/handalone.js
+++ b/src/js/base/handalone.js
@@ -20,7 +20,10 @@ require(["pyret-base/js/runtime", "program"], function(runtimeLib, program) {
   var EXIT_ERROR_JS = 5;
   var EXIT_ERROR_UNKNOWN = 6;
 
-  runtime.setParam("command-line-arguments", process.argv.slice(1));
+  var commandLineArguments = process.argv.slice(1);
+  var runMode = (commandLineArguments.indexOf("--run") > 0);
+
+  runtime.setParam("command-line-arguments", commandLineArguments);
 
   var postLoadHooks = {
     "builtin://srcloc": function(srcloc) {
@@ -147,35 +150,39 @@ require(["pyret-base/js/runtime", "program"], function(runtimeLib, program) {
       runtime.setParam("current-checker", currentChecker);
     }
   };
-  postLoadHooks[main] = function(answer) {
-    var checkerLib = runtime.modules["builtin://checker"];
-    var checker = runtime.getField(runtime.getField(checkerLib, "provide-plus-types"), "values");
-    var getStack = function(err) {
-      console.error("The error is: ", err);
-      var locArray = err.val.pyretStack.map(runtime.makeSrcloc);
-      var locList = runtime.ffi.makeList(locArray);
-      return locList;
-    };
-    var getStackP = runtime.makeFunction(getStack, "get-stack");
-    var toCall = runtime.getField(checker, "render-check-results-stack");
-    var checks = runtime.getField(answer, "checks");
-    runtime.safeCall(function() {
-      return toCall.app(checks, getStackP);
-    }, function(summary) {
-      if(runtime.isObject(summary)) {
-        process.stdout.write(runtime.getField(summary, "message"));
-        process.stdout.write("\n");
-        var errs = runtime.getField(summary, "errored");
-        var failed = runtime.getField(summary, "failed");
-        if(errs !== 0 || failed !== 0) {
-          process.exit(EXIT_ERROR_CHECK_FAILURES);
-        }
-        else {
-          process.exit(EXIT_SUCCESS);
-        }
-      }
-    });
-  }
+  postLoadHooks[main] = runMode ?
+    (function(answer) {
+      //console.dir(answer);
+    }) :
+    (function(answer) {
+        var checkerLib = runtime.modules["builtin://checker"];
+        var checker = runtime.getField(runtime.getField(checkerLib, "provide-plus-types"), "values");
+        var getStack = function(err) {
+          console.error("The error is: ", err);
+          var locArray = err.val.pyretStack.map(runtime.makeSrcloc);
+          var locList = runtime.ffi.makeList(locArray);
+          return locList;
+        };
+        var getStackP = runtime.makeFunction(getStack, "get-stack");
+        var toCall = runtime.getField(checker, "render-check-results-stack");
+        var checks = runtime.getField(answer, "checks");
+        runtime.safeCall(function() {
+          return toCall.app(checks, getStackP);
+        }, function(summary) {
+          if(runtime.isObject(summary)) {
+            process.stdout.write(runtime.getField(summary, "message"));
+            process.stdout.write("\n");
+            var errs = runtime.getField(summary, "errored");
+            var failed = runtime.getField(summary, "failed");
+            if(errs !== 0 || failed !== 0) {
+              process.exit(EXIT_ERROR_CHECK_FAILURES);
+            }
+            else {
+              process.exit(EXIT_SUCCESS);
+            }
+          }
+        });
+      });
 
   function renderErrorMessageAndExit(execRt, res) {
     if (execRt.isPyretException(res.exn)) {
--- a/src/js/base/handalone.js
+++ b/src/js/base/handalone.js
@@ -20,10 +20,7 @@ require(["pyret-base/js/runtime", "program"], function(runtimeLib, program) {
   var EXIT_ERROR_JS = 5;
   var EXIT_ERROR_UNKNOWN = 6;
 
-  var commandLineArguments = process.argv.slice(1);
-  var runMode = (commandLineArguments.indexOf("--run") > 0);
-
-  runtime.setParam("command-line-arguments", commandLineArguments);
+  runtime.setParam("command-line-arguments", process.argv.slice(1));
 
   var postLoadHooks = {
     "builtin://srcloc": function(srcloc) {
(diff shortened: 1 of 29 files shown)


awstlaur:
Hold off on this for now -- I discovered that this is broken if running a file which contains check blocks which end in error. There is some problem with the values produced by the getStack in load-lib.js

src/js/base/handalone.js
```diff
@@ -147,34 +150,45 @@ require(["pyret-base/js/runtime", "program"], function(runtimeLib, program) {
...
-    var checker = runtime.getField(runtime.getField(checkerLib, "provide-plus-types"), "values");
-    var getStack = function(err) {
-      console.error("The error is: ", err);
-      var locArray = err.val.pyretStack.map(runtime.makeSrcloc);
-      var locList = runtime.ffi.makeList(locArray);
-      return locList;
+  if (runMode) {
+    postLoadHooks[main] = function(answer) {
+      var definedValues = runtime.getField(answer, "defined-values");
+      var definedExitCode = definedValues["exit-code"];
```
jpolitz: So this is a new implicit interface that if a module provides an exit-code field, and it's the main module of a run, that value is used as the exit code for the whole run?

I don't know that this is bad, but it seems somewhat unique.

How do other languages do this?  Can we borrow a different design?  What makes us think this is the right design?

src/scripts/run-phase
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
```
jpolitz: There should be some (brief) prose added to the README about `run-phaseX` and how to use them.

jpolitz:
This looks good, aside from my comment about quitting with an exit code.  Lots of languages use a `system.exit()` kind of mechanism for doing this.  I don't know that piggybacking on the module system is necessarily the right way to do it (or not).  It doesn't strike me as obvious, and seems like a non-inevitable design.

Maybe it should be in all caps, or start with an underscore, or be called out in some other way if it's going to be done by provided value.

Another design would be to re-use `raise`, with a special kind of error that comes with a signal number.  So, for example:

```
raise(builtins.exit(0))
```

Could signal an exit code of 0 programmatically.  Catching and checking for that error would have to be built-in to the various standalone handlers, which would call the underlying `process.exit` after doing any cleanup.

src/js/base/handalone.js
blerner: Ooh, I don't like this very much...  I *think* (first-draft speculation ahead) I'd rather a design where we had a runtime-provided function `set-exit-code : Number -> Nothing` that sets the eventual exit code of the program, or even `exit-now-with-code` which terminates the program now with the given exit code.  But having a specially-named identifier that could be provided by any module?  That seems too precious and unique to be a memorable way to exit the program...

jpolitz:
Note that we also have `user-break` defined, which is a special kind of exception handled correctly by the checker library (via `run-task`) so that it correctly kills things and isn't caught by `raises`.  We could do the same thing with another special error, and if we don't want to awkwardly re-use `raise`, only expose the special error through a runtime call.

TL;DR, the infrastructure for a special control flow value exists.  Any reason to not do that?

awstlaur:
Having something like `user-break` which takes an exit code would be nice. For the sake of this PR, as long pyret.arr can specify an exit code, that's good enough.

Of course, this is all assuming we actually want non-zero exit codes when pyret.jarr is used as an interpreter on a file which has failing tests.

jpolitz:
> Of course, this is all assuming we actually want non-zero exit codes when pyret.jarr is used as an interpreter on a file which has failing tests.

I think that's a good idea; in general I think `--run` might be the best default option for calling the `pyret` command:

```
$ pyret failing.arr
1 test passed, 3 tests ran, 2 tests failed
$ echo $?
1
$ pyret passing.arr
Looks shipshape, your test passed, mate!
$ echo $?
0
```

awstlaur:
@jpolitz @blerner  Could y'all check out the `exit-code` branch? I added a variant to the runtime errors and a simple conditional in `handalone.js` which tests for said error.  I also opted for using an exit code in `pyret.arr`, because it feels more appropriate than raising an error when the user messes up a command-line flag or something. Also, raising `exit` has the side-effect of skipping the postLoadHook which just ends up printing "The program did not define any tests".

If it looks good, I'll add a simple test and submit a pull request, then update this branch accordingly after exit-code is merged into horizon.

blerner:
Hmm.  Haven't had a chance to look yet, but I feel uncomfortable _expecting_ that a postLoadHook could get skipped...

awstlaur:
> Hmm. Haven't had a chance to look yet, but I feel uncomfortable expecting that a postLoadHook could get skipped...

I think I'd agree in general, but this particular postLoadHook only renders check results. The question then becomes: do we want an `exit` that skips rendering check results? I had assumed yes.

blerner:
I'd rather have support for some kind of `postLoadFailedHook`, so to speak.  The intent in the `postLoadHook`s is to allow the overall standalone to have full interposition after each module gets to load.  Expecting a module-load to be bypassed seems very odd to me.

awstlaur:
What if we just move the "is this an exit" check to the postLoadHook itself? That would be functionally the same, assuming it runs before the `onComplete` in the standalone.

awstlaur:
Er, wait, scratch that - the postLoadHook wouldn't run if something was raised

awstlaur:
From @jpolitz, via direct instant messaging:

> I think if exit-code signals are coming through exceptions, we should treat them uniformly, except in deciding what to pass to System.exit

> So I think that makes the question of a `postLoadFailedHook` orthogonal

> Because it should be called if the user writes `raise("I want to be done now pls")` just as much as it should as if they used `exit-with-code(1)`

Yeah - at the moment, the `postLoadHook` would get skipped anyway. The only difference with an exit code is that nothing would be written to stdout or stderr - it would just exit. Unless we want an exit-code mechanism that will would a message like "`Program exited with code X`" ? I've been assuming we don't.

jpolitz:
I think it's actually nice if the default says "Exited with code `n`".  There could be another exception constructor, with a name like `exit-code-quiet` that could be used for traditional scripting that would suppress the message, and just set the code in the process.

awstlaur:
> I think it's actually nice if the default says "Exited with code `n`".  There could be another exception constructor, with a name like `exit-code-quiet` that could be used for traditional scripting that would suppress the message, and just set the code in the process.

I like this idea!

I think implementing the support of `postLoadFailedHook` is a bit beyond the scope of this. It may be desirable as a separate venture, however.

So, I think I will go ahead and add a "quiet" variant to what is currently on the `exit-code` branch, and open a PR for that, unless there are further points to discuss.

jpolitz:
SGTM

awstlaur:
@jpolitz @blerner I'd like to discuss the design of how to handle the scenario wherein the pyret file passed via `--run` itself calls `system.{exit, exit-quiet}`.

As of right now, on this branch, the `exit` and `exit-quiet` variants are not distinguished from other errors. In particular, running something like
```
src/scripts/phaseA examples/exit-code/success.arr
```
will render the message (`"Exited with code 0"`, plus a stack trace), then exit with code `1`, as specified in `load-lib`'s `render-error-message` function.

I'm thinking we could wrap `cli-module-loader.run`'s call to `load-lib.run-program` in a `run-task`. Then, if we get back an `exit` or `exit-quiet` error, simply call `system.{exit, exit-quiet}` from there. This would have the advantage of keeping the rendering and exiting logic inside `handalone.js`, where it currently lives.

There are some other options, including but not limited to:
- porting handlone's exiting logic to load-lib, including the call to process.exit
- making load-lib's render-error-message return the appropriate message and exit code.

Both of these would introduce repeated code/logic, so I'm not a huge fan of either.

Let me know what y'all think of the first option (a wrapper around `run-task`) and share thoughts, opinions, etc.

jpolitz:
I like the `run-task` idea.  This is precisely the scenario it was designed for.

Your suggestion of forwarding the exit code, and its silent-ness, I also agree with.

Incidentally, while we settled on this design through our own discussion, it pretty much exactly matches what Python does:

https://docs.python.org/3/library/sys.html#sys.exit

awstlaur:
@jpolitz  I realized the a few things about load-lib's `run-program` function:
- it depends on taking in a separate runtime
- it already behaves like `run-task`, in that it catches errors
- it returns a POpaque, requiring extraction of values to happen inside load-lib

So, the options are now:

1. change `run-program` to return something non-opaque, that the requesting runtime can work with, and/or change it to use the requesting runtime's stack. this seems like it would be a relatively hefty change.
2. add `exn-unwrap` and `get-exn-field` to load-lib, so that propagating an exit code would look like the function/usage below. this would be a simpler change, but would involve getting a field name dynamically
```pyret
import load-lib as L
import runtime-lib as R

fun propagate-exit(result) block:
  v = L.exn-unwrap(result)
  when ERR.is-exit(v):
    code = L.exn-get-field(result, "code")
    SYS.exit(code)
  end
  when ERR.is-exit-quiet(v):
    code = L.exn-get-field(result, "code")
    SYS.exit-quiet(code)
  end
end

result = L.run-program(R.make-runtime(), L.empty-realm(), program.js-ast.to-ugly-source(), options)
if L.is-success-result(result):
  L.render-check-results(result)
else:
  _ = propagate-exit(result)
  L.render-error-message(result)
end
```
3. something like option 2, except write `propagate-exit` entirely inside `load-lib`.

jpolitz:
I do feel like the right design is for `load-lib` to provide an API that knows how to check for exit codes thrown from its runtime.  As long as it's `load-lib` that's doing it, it should work within the separated-runtimes design, since that library is the one that gets references to both runtimes.

Then the job of the main script is just to use that API to inspect the value from `run-program` with this blessed API.

awstlaur:
That sounds good! Then `propagate-exit` would look like:

```pyret
fun propagate-exit(result) block:
  when L.is-exit(result):
    code = L.get-exit-code(result)
    SYS.exit(code)
  end
  when L.is-exit-quiet(result):
    code = L.get-exit-code(result)
    SYS.exit-quiet(code)
  end
end
```

awstlaur:
@jpolitz I recall us discussing a potential test plan for `--run` that would iterate through the files in the `examples/` directory and run them via `src/scripts/phaseA` (the wrapper for `node build/phaseA/pyret.jarr --run ...`). However, some of the examples are broken. E.g.:
- `examples/heap.arr` uses `order` as an identifier (which is now a keyword),
- `examples/loop.arr` has several unqualified blocks (which now need `block:`).

Furthermore, a thorough test plan of `--run` would involve checking that parse errors, compile errors, runtime errors, and errored/failed checks are reported properly, for which I don't believe we have canonical tests (as their own `*.arr` files).

To avoid prolonging / bloating this pull request further, I'd rather have this meged, and then submit a PR for a _separate_ set of changes. The future PR would include
- fix examples
- add examples which demonstate various erroroneous scenarios (see previous paragraph)
- add tests which use the `--run` flag to run the examples, either asserting success or failure as needed.

(In the meantime, I'll document the phaseX runner scripts in the README)

PR #966, opened 2017-01-13
Fix #962

### LINKED ISSUE #962: keywords allowed in table names…sometimes
I can write
```
cs = load-table: timestamp, name, class
```
but not
```
sieve cs using class
```
Since I can't write the latter, I should also be forbidden from writing the former.

belph:


jpolitz:
Thanks, @belph for the quick fix!

PR #970, opened 2017-01-14
maybe-surface-parse

awstlaur:
This PR adds a `maybe-surface-parse` function to the `parse-pyret` lib,  wrapping the possibility of parse failure in an `Either` data type. 

- `maybe-surface-parse` outputs either `left({exn: exn, message: message})` or `right(ast)`
- `surface-parse` either returns `ast`, or logs `message` to stderr and raises `exn` (i.e., the behavior of `surface-parse` is unchanged)
- `test-parse.arr` was generated from `tests/parse/parse.js`, with `it` closures becoming `check` blocks (hence the `"should do X"` labels)
- `test-parse-errors.arr` has new tests which expect specific variants of ParseError

### DIFF
--- a/src/js/trove/parse-pyret.js
+++ b/src/js/trove/parse-pyret.js
@@ -1315,6 +1315,7 @@
     }
 
     function parseDataRaw(data, fileName) {
+      var message = "";
       try {
         const toks = tokenizer.Tokenizer;
         const grammar = parser.PyretGrammar;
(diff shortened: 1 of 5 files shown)


awstlaur:
@jpolitz @blerner Have you had a chance to look at this yet?

src/js/trove/parse-pyret.js
```diff
@@ -1344,31 +1345,56 @@
...
             //            console.log(("" + asts[i]) === ("" + asts2[i]));
           }
-          return translate(ast, fileName);
+          return RUNTIME.ffi.makeRight(translate(ast, fileName));
         }
       } catch(e) {
-        // console.error("Fatal error in parsing: ", e);
-        throw e;
+        if (RUNTIME.isPyretException(e)) {
+          return RUNTIME.ffi.makeLeft(RUNTIME.makeObject({
```
blerner: This feels a bit odd to me -- having this not-quite-an-exception object that colludes with the handler code below just seems wrong, and more than a little bit non-type-correct.  @jpolitz is there a "cleaner" or "more officially sanctioned" design you can think of, that might tie in better with the existing compile- and runtime errors we have?  I feel like there should be, but I can't think of it yet.

jpolitz: The other pattern we have for this is `execThunk`, which is exposed as `run-task`.

In general, I don't know what cases this would throw a _non_ Pyret error.

Also, in general I think using `catch` like this isn't safe with the stack, because the handling code isn't added to the frame.  I think in this case the parser is assumed to be flat-ish (since all it does is call constructors), so the `catch` works.

I'm not bothered by the unwrapping below – that just seems like how the two versions of this has to work.  The exception-throwing version unwrapping an `Either` is a common pattern, right?  It could be done the opposite way, where `maybe-surface-parse` catches the exception and wraps, instead of `parseDataRaw` doing it, but I don't see a big difference.

I think I'm OK with this, since the construction part of parsing has to be flat anyway and has all sorts of issues with stack-unsafe constructors as it is.  So unless someone points out something I'm misunderstanding, I'm happy merging this.

PR #973, opened 2017-01-17
trying to fix constructor flatness

puppyofkosh:
This should fix it the problem of constructors not being counted as flat introduced in

https://github.com/brownplt/pyret-lang/commit/0521463c10e7c066051642ffafedd1c025932b2f

Passes the tests


jpolitz:
Awesome, thanks

PR #978, opened 2017-01-31
Replacing "+" by an explicit function call in anf-loop-compiler

puppyofkosh:
This should allow for a few dozen more caller optimizations (and maybe a handful more callee optimizations, but not many).

I only did the replacement where I was confident it wouldn't result in horribly unreadable code.


src/js/trove/global.js
```diff
@@ -26,10 +26,22 @@
...
                 "name":"print",
                 "typ":["arrow", ["Any"], "Any"]},
-      "test-print": ["arrow", ["Any"], "Any"],
-      "print-error": ["arrow", ["Any"], "Any"],
-      "display": ["arrow", ["Any"], "Any"],
-      "display-error": ["arrow", ["Any"], "Any"],
+      "test-print": {"bind":"fun",
+                     "flatness":0,
+                     "name":"test-print",
+                     "typ":["arrow", ["Any"], "Any"]},
```
jpolitz: These are verbose.  I wonder if we should add another abbreviation, `"flatarrow"`, that is specially handled by `type-util.js`.

PR #980, opened 2017-02-03
Exit code

awstlaur:
(See the comments on PR https://github.com/brownplt/pyret-lang/pull/961 for a discussion of design decisions about an exit code mechanism.)

This PR adds `exit(code :: Number)` and `exit-quiet(code :: Number)` error variants which are able to be `raise`d, leading directly to an invocation of `process.exit(<code>)`. Using the `exit` variant will cause the message `Exited with code <code>` to be written stdout prior to exiting.

This PR also changes `pyret.arr` to use the exit code mechanism (via `exit-quiet`*), rather than `raise`ing other errors (in response to invalid command-line arguments, etc). This is needed in the context of PR https://github.com/brownplt/pyret-lang/pull/961 so that an exit code based on checker results can be specified in Pyret (e.g. a successful run, but with failing checks, still needs a non-zero exit code).

*I opted for `exit-quiet` mainly in anticipation of an option to get detailed check results from `--run` in parse-able JSON, so that we will be able to do pipe the output into a script which reads valid JSON from stdin. (E.g. a grading script). See the `checker-api` branch for my progress on that.

Finally, this PR adds some examples and tests.

`make no-diff-standalone` and the travis build (`make all-pyret-test`) both pass.

### DIFF
--- a/src/arr/compiler/pyret.arr
+++ b/src/arr/compiler/pyret.arr
@@ -1,6 +1,7 @@
 #lang pyret
 
 import cmdline as C
+import error as ERR
 import file as F
 import render-error-display as RED
 import string-dict as D
@@ -10,7 +11,10 @@ import file("compile-structs.arr") as CS
 import file("locators/builtin.arr") as B
 import file("server.arr") as S
 
-fun main(args):
+success-code = 0
+failure-code = 1
+
+fun main(args :: List<String>) -> Number:
   options = [D.string-dict:
     "serve",
       C.flag(C.once, "Start the Pyret server"),
@@ -89,7 +93,8 @@ fun main(args):
         B.set-allow-builtin-overrides(r.get-value("allow-builtin-overrides"))
       end
       if not(is-empty(rest)):
-        raise("No longer supported")
+        _ = print("No longer supported")
+        failure-code
       else:
         if r.has-key("build-runnable") block:
           outfile = if r.has-key("outfile"):
@@ -97,7 +102,7 @@ fun main(args):
           else:
             r.get-value("build-runnable") + ".jarr"
           end
-          CLI.build-runnable-standalone(
+          _ = CLI.build-runnable-standalone(
               r.get-value("build-runnable"),
               r.get-value("require-config"),
               outfile,
@@ -113,11 +118,14 @@ fun main(args):
                 compiled-cache: compiled-dir,
                 display-progress: display-progress
               })
+          success-code
         else if r.has-key("serve"):
           port = r.get-value("port")
           S.serve(port)
+          success-code
         else if r.has-key("build-standalone"):
-          raise("Use build-runnable instead of build-standalone")
+          _ = print("Use build-runnable instead of build-standalone\n")
+          failure-code
           #|
           CLI.build-require-standalone(r.get-value("build-standalone"),
               CS.default-compile-options.{
@@ -151,25 +159,30 @@ fun main(args):
                 print-error(tostring(e))
                 print-error("\n")
               end
-              raise("There were compilation errors")
+              _ = print("There were compilation errors\n")
+              failure-code
             end
           end
         else if r.has-key("run"):
-          CLI.run(r.get-value("run"), CS.default-compile-options.{
+          _ = CLI.run(r.get-value("run"), CS.default-compile-options.{
               standalone-file: standalone-file,
               compile-module: true,
               display-progress: display-progress,
               check-all: check-all
             })
+          success-code
         else:
-          print(C.usage-info(options).join-str("\n"))
-          raise("Unknown command line options")
+          _ = print(C.usage-info(options).join-str("\n"))
+          _ = print("Unknown command line options\n")
+          failure-code
         end
       end
     | arg-error(message, partial) =>
-      print(message + "\n")
-      print(C.usage-info(options).join-str("\n"))
+      _ = print(message + "\n")
+      _ = print(C.usage-info(options).join-str("\n"))
+      failure-code
   end
 end
 
-_ = main(C.args)
+exit-code = main(C.args)
+raise(ERR.exit(exit-code))
--- a/src/arr/compiler/pyret.arr
+++ b/src/arr/compiler/pyret.arr
@@ -1,10 +1,10 @@
 #lang pyret
 
 import cmdline as C
-import error as ERR
 import file as F
 import render-error-display as RED
 import string-dict as D
+import system as SYS
 import file("cli-module-loader.arr") as CLI
 import file("compile-lib.arr") as CL
 import file("compile-structs.arr") as CS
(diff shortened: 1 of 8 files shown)


src/arr/compiler/pyret.arr
```diff
@@ -89,15 +93,16 @@ fun main(args):
         B.set-allow-builtin-overrides(r.get-value("allow-builtin-overrides"))
       end
       if not(is-empty(rest)):
-        raise("No longer supported")
+        _ = print("No longer supported\n")
```
jpolitz: I would use `block:` and `print-error` here.  Normally not this nitpicky, but I'd like to move towards `pyret.arr` being an exemplar of good style.

src/arr/compiler/pyret.arr
```diff
@@ -89,15 +93,16 @@ fun main(args):
...
+        failure-code
       else:
         if r.has-key("build-runnable") block:
           outfile = if r.has-key("outfile"):
             r.get-value("outfile")
           else:
             r.get-value("build-runnable") + ".jarr"
           end
-          CLI.build-runnable-standalone(
+          _ = CLI.build-runnable-standalone(
```
jpolitz: This is already in a `block:`, so the `_` should be unnecessary, right?

src/arr/compiler/pyret.arr
```diff
@@ -145,31 +153,38 @@ fun main(args):
...
               display-progress: display-progress,
               check-all: check-all
             })
+          success-code
         else:
-          print(C.usage-info(options).join-str("\n"))
-          raise("Unknown command line options")
+          _ = print(C.usage-info(options).join-str("\n"))
+          _ = print("Unknown command line options\n")
+          failure-code
```
jpolitz: In these cases, I wonder if we should just keep the "raise" around, and have the toplevel's behavior for uncaught raises be to set the exit code and show the exception.  In particular, just printing here gives _less_ information, because the `raise` would collect and print out a stack trace to aid debugging.

src/arr/compiler/pyret.arr
```diff
@@ -145,31 +153,38 @@ fun main(args):
...
-      print(C.usage-info(options).join-str("\n"))
+      _ = print(message + "\n")
+      _ = print(C.usage-info(options).join-str("\n"))
+      failure-code
   end
 end
 
-_ = main(C.args)
+exit-code = main(C.args)
+raise(ERR.exit-quiet(exit-code))
```
jpolitz: This makes sense, though I wonder if `exit` and `exit-quiet` should be runtime methods that have the effect of throwing this special value.  I suggest this mainly because it may give us some more flexibility to update the implementation behind the interface later.

jpolitz:
Jeez I don't like it that it puts a red X on my review when I request changes -- pretty negative UX that invokes ideas of success and failure based on merging/review decisions.  (This was my first attempt at a formal Github review).

Anyway, I'm delighted this work is getting done.

src/arr/compiler/pyret.arr
awstlaur: > just printing here gives _less_ information, because the `raise` would collect and print out a stack trace to aid debugging.

That is true, and would be more beneficial for developers. That then leads to this question: do we want using main `pyret.arr` blob to be _developer_-friendly, or _user_-friendly? I'd argue that stack traces are not user-friendly.

src/arr/compiler/pyret.arr
awstlaur: I'd be in favor of that. Should they live in the global namespace?

jpolitz: Good question.  I wonder if now is a good time to make a `system` module that lives in `trove`.

src/arr/compiler/pyret.arr
jpolitz: Yup, good point.

I think there's a good argument for finding a way to make this work in either way.  Like a CLI option that decides whether these get raised or just printed with a  return code set.

There are other `raise` uses for impossible cases in the compiler, which actually should be errors that get reported with a stack trace for debugging.  So these are distinct cases.  I get the reason for this now, and I think it's find to make this change and see if it actually hurts development.

Thanks for pointing out this distinction.

awstlaur:
Thanks for the feedback! I reacted with the thumbs-up icon to a couple of the minor things that I agree with and will be adding, and commented directly on some others.

> As an orthogonal comment, it would be nice if we had some way to run things in `examples/` to make sure they don't get stale.  Having code in the repo that isn't regularly run doesn't seem like a good idea -- I made the mistake of introducing this examples directory early on, and it seems like a magnet for bitrot.

I'm probably going to add tests of using `--run` on the command-line to https://github.com/brownplt/pyret-lang/pull/961 ; that test easily just go through `examples/` and run everything to make sure it all runs correctly. N birds, N/2 stones.

src/arr/compiler/pyret.arr
awstlaur: Cool! Here is what I am thinking.

- If using the main `pyret.arr` results in having a stack trace (Pyret or JS) in the output, it's either because the user supplied some kind of `--verbose-errors` option, or there is a legitimate bug that would need to get resolved, or maybe both.
- Otherwise, if there is some user error, we should print a friendly/helpful message.
- If `--verbose-errors` is _not_ supplied, then a non-zero exit code should only result either from such anticipated user errors, or due to failing `check` statements.

There may be cases I didn't consider, but I feel like the general spirit of the above list seems decent.

src/arr/compiler/pyret.arr
awstlaur: Perhaps; `system.exit` has a nice ring to it.

I imagine we may also want `system.argv`, `system.cwd`, etc.  wrappers for the counterpart node `process` methods. That'd maybe be beyond the scope of this PR, but something to think about.

src/arr/compiler/pyret.arr
awstlaur: Fixed in https://github.com/brownplt/pyret-lang/commit/09c4fa85c194b9751fb1a7a97a7f472fdeea11e2

src/arr/compiler/pyret.arr
awstlaur: Fixed in https://github.com/brownplt/pyret-lang/commit/09c4fa85c194b9751fb1a7a97a7f472fdeea11e2

src/arr/compiler/pyret.arr
awstlaur: `system` module interface added in https://github.com/brownplt/pyret-lang/commit/9562b2d362b81aa8486d8fabba782c739146f4a8

src/arr/compiler/pyret.arr
awstlaur: Of course, that^ is probably a bit beyond the scope of this PR. If you agree, then I believe I've addressed all the code review flags.

jpolitz:
OK, good to merge, @awstlaur?  I think my comments have been addressed.

awstlaur:
@jpolitz Yes, I believe this is all set to merge in.

PR #984, opened 2017-02-26
fix tosource method of s-when

nishanthprakash:
there needs to be a space between the "when" token and the following parenthesis, otherwise leads to syntax error


blerner:
I think the right fix here is actually to fix the tokenizer, such that when tokenizing `when`, it says `parenIsForExp: true`.  Then the whitespace isn't significant.

PR #987, opened 2017-03-06
Support no-variant data types

### LINKED ISSUE #986: Data declaration with no variants produces bad error message
Defining a data type with no variants produces an awkward error message.

    data NoVariants:
    end

> Evaluating the field lookup expression at `file:///home/travis/build/brownplt/code.pyret.org/node_modules/pyret-lang/src/arr/compiler/ast-util.arr:206:40-206:48` errored.
> 
> does not have a field named `first`
> 
> Please report this as a bug.

This is caused by the `collect-shared-fields` helper function in [ast-util.arr](https://github.com/brownplt/pyret-lang/blob/master/src/arr/compiler/ast-util.arr).  Possible fixes:
* Explicitly prohibit no-variant data types with a better error message.
* Add an `is-empty` check in `collect-shared-fields`.  This would allow empty data types to be created.  Non-instantiable data types are potentially useful for annotating non-returning functions.

nickgaya:
Fixes #986 by allowing empty data definitions.


blerner:
LGTM.  @jpolitz which branch should we land this on?

jpolitz:
memo-jsast seems like a good choice. 👍

PR #988, opened 2017-03-06
fix tests + use Number instead of option

puppyofkosh:
I had time to fix some failing tests and change the flatness analysis to use Number instead option. Travis build:

https://travis-ci.org/puppyofkosh/pyret-lang/builds/208263373


puppyofkosh:
Don't merge this! I think the change to js-of-pyret introduced a tricky bug into the flatness analysis.

PR #991, opened 2017-03-26
Change runtime to only bounce if a flag is set

puppyofkosh:
When running in "everything-flat mode" we'll need to make sure that no bounces ever happen, so I'm trying to change the runtime to not bounce unless this new "allowBounce" flag is set. 

The main things I had to change were:
1) make-standalone() takes a pyret-level callback since it relies on the async requirejs.optimize function

2) When the allowBounce flag is false and we are running a chunk of pyret code that might fail (the checker does this with execThunk(), for example), instead of clearing the entire stack, and then running the potentially failing code, we use a try-catch block and run it on the same stack

3) When the allowBounce flag is false we use the synchronous version of require() in some places

This runtime code is pretty confusing so I'd very much like some feedback from someone who's worked with it before. It passes the tests with the allowBounce flag true, which is the "normal" behavior of the runtime (just to make sure that I didn't break anything just by adding the flag).

I have a local build that compiles everything as flat (with the allowBounce flag false) that passes all the tests, but I think it makes sense to just try to get the changes to the runtime reviewed first.

Passing travis build:
https://travis-ci.org/puppyofkosh/pyret-lang/builds/215244436

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -307,6 +307,7 @@ clean:
 .PHONY : test-clean
 test-clean:
 	$(call RMDIR, tests/compiled)
+	$(call RM, tests/pyret/*.jarr)
 
 # Written this way because cmd.exe complains about && in command lines
 new-bootstrap: no-diff-standalone
--- a/src/arr/compiler/cli-module-loader.arr
+++ b/src/arr/compiler/cli-module-loader.arr
@@ -425,16 +425,18 @@ fun build-runnable-standalone(path, require-config-path, outfile, options) block
       end
 
       when options.collect-times: stats.set-now("standalone", time-now()) end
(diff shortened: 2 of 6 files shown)


blerner:
@puppyofkosh can you rebase on the latest memo-jsast changes I've merged?  The runtime changes should be easy enough; I'm not quite sure about the load-lib ones.

puppyofkosh:
Looks like the changes from the rebase cause some tests to fail. I'll respond again once I have them fixed.

puppyofkosh:
@blerner Okay, it's rebased and everything is passing.

puppyofkosh:
This PR is redundant. See
https://github.com/brownplt/pyret-lang/issues/1020

PR #992, opened 2017-03-26
Fix

nishanthprakash:


jpolitz:
Is this just to merge the one change with `%`?  I want to make sure there isn't other stuff you meant to include, since there are 5 commits listed, but the net effect only changes one line.

nishanthprakash:
yes, the other 4 commits are actually already merged (I merged them into my remote after the merge with this branch, thats probably why there still there)

PR #997, opened 2017-04-01
[WIP] Check Results API

awstlaur:
This PR adds a `--run-full-report` command-line flag whose usage is as follows:

```
node build/phaseA/pyret.jarr --run-full-report file.arr
```

This will parse, compile, and run `file.arr`, and write JSON data to standard out.

No matter what happens in `file.arr` (and any Pyret dependencies it has), I claim
this API is robust enough to not allow stray messages to be written to stdout.
This is important, because this API will be used by grading scripts to collect
data: a grading script may want to redirect stdout to a `file.arr.json` file,
and then do its remaining "business logic" outside of Pyret, in the language
of its author's choice.

For examples of input programs and their resulting JSON data, see
[TM030 Check Results API](https://github.com/brownplt/pyret-lang/wiki/TM030-Check-Results-API)

Some notes about the PR and its changeset:

- This is marked work-in-progress (WIP) because I haven't added tests of the
proposed functionality. Before I put effort into writing said tests, I would
like us to agree upon the new interface and expected behavior.

- The diff on `checker.arr` is a bit messy: the gist there is that the
functionallity of `results-summary` was expanded and moved into the new `results-report`
function; now, `results-summary` depends on `results-report`.

- The added `toJSON` in `ffi.js` has been robust enough for this use case, but
I do not claim it can cleanly convert _any_ Pyret data structure into JSON.
Its cases mirror those in `json-structs.arr:tojson`, but it does not handle
string-dicts.

- `load-lib.js` has some repeated logic: `renderCheckResults` and `renderCheckReport`
are similar, and `renderErrorMessage` and `renderErrorReport` are also similar.
In theory, the `renderCheckResults` and `renderErrorMessage` may be able to depend
on the `render*Report` functions, but (I believe) it would require messy stack
management and thus might not be worth the trouble.


jpolitz:
@jswrenn I'm curious if this design is useful to you, or if your recent work on doing lots of testing automation supersedes this in useful ways.

awstlaur:
I'm not coming back to this

PR #1002, opened 2017-04-04
fix error-rendering error for lookup-non-object

awstlaur:
Because code.pyret.org's configuration skips these code paths, it is
necessary to use the command-line / node.js environment to test this.

To reproduce:
- create a file `tmp.arr` with just `"hello world".length()`
- `make tmp.jarr`
- `node tmp.jarr`


blerner:
LGTM

PR #1008, opened 2017-04-18
Allow \b as escaped backspace in strings.

ds26gte:
Allows `\b` to stand for escaped backslash inside Pyret strings.

```
string-length("a\bc") is 3
string-to-code-point("\b") is 8
string-from-code-point(8) is "\b"
```

### DIFF
--- a/src/js/base/pyret-tokenizer.js
+++ b/src/js/base/pyret-tokenizer.js
@@ -6,7 +6,7 @@ define(["jglr/jglr"], function(E) {
   const GenTokenizer = E.Tokenizer;
   const STICKY_REGEXP = E.STICKY_REGEXP;
 
-  const escapes = new RegExp("^(.*?)\\\\([\\\\\"\'nrt]|u[0-9A-Fa-f]{1,4}|x[0-9A-Fa-f]{1,2}|[0-7]{1,3}|[\r\n]{1,2})");
+  const escapes = new RegExp("^(.*?)\\\\([\\\\\"\'bnrt]|u[0-9A-Fa-f]{1,4}|x[0-9A-Fa-f]{1,2}|[0-7]{1,3}|[\r\n]{1,2})");
   function fixEscapes(s) {
     var ret = "";
     var match = escapes.exec(s);
@@ -18,6 +18,7 @@ define(["jglr/jglr"], function(E) {
       else if (esc === "\r") {}
       else if (esc === "\n\r") {}
       else if (esc === "\r\n") {}
+      else if (esc === "b") { ret += "\b"; }
       else if (esc === "n") { ret += "\n"; }
       else if (esc === "r") { ret += "\r"; }
       else if (esc === "t") { ret += "\t"; }
@@ -195,7 +196,7 @@ define(["jglr/jglr"], function(E) {
                "\\\\[01234567]{1,3}" +
                "|\\\\x[0-9a-fA-F]{1,2}" +
                "|\\\\u[0-9a-fA-f]{1,4}" +
-               "|\\\\[\\\\nrt\"\'`]" +
+               "|\\\\[\\\\bnrt\"\'`]" +
                "|`{1,2}(?!`)" +
                "|[^`\\\\])*```", STICKY_REGEXP); // NOTE: Allow unescaped newlines
   const dquot_str =
@@ -203,14 +204,14 @@ define(["jglr/jglr"], function(E) {
                "\\\\[01234567]{1,3}" +
                "|\\\\x[0-9a-fA-F]{1,2}" +
                "|\\\\u[0-9a-fA-f]{1,4}" +
-               "|\\\\[\\\\nrt\"\']" +
+               "|\\\\[\\\\bnrt\"\']" +
                "|[^\\\\\"\n\r])*\"", STICKY_REGEXP);
   const squot_str =
     new RegExp("^\'(?:" +
                "\\\\[01234567]{1,3}" +
                "|\\\\x[0-9a-fA-F]{1,2}" +
                "|\\\\u[0-9a-fA-f]{1,4}" +
-               "|\\\\[\\\\nrt\"\']" +
+               "|\\\\[\\\\bnrt\"\']" +
                "|[^\\\\\'\n\r])*\'", STICKY_REGEXP);
 
   const unterminated_string = new RegExp("^(?:[\"\']|```).*", STICKY_REGEXP);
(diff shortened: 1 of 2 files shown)


blerner:
How should backspace chars `to-string` and `_torepr`?  Every other escape character we have has a non-escaped rendering: backspace doesn't.  Do we want it to appear as a zero-width invisible character?  (Misleading.)  Do we want it to make the subsequent text overlap the last character of the preceding?  (Weird, inefficient and possibly impossible to do.)  How should it show up in `text` images? (That *does* affect world programs, especially if we allow backspace characters to slip through `on-key` handlers.)

jswrenn:
[␈](http://www.fileformat.info/info/unicode/char/2408/index.htm)?

blerner:
What on earth glyph is that??  Looks like lint on my screen :(

ds26gte:
It shows up as tall rectangle glyph.  I.e., `a\bc` displays as `a[]c`.  (Jack's glyph is a small B superscript followed by a small S subscript, which I guess is a special Unicode character. There is also a Unicode character that looks like a flattened pentagon with an x inside it that we could try, but it is rather wide.)

Dealing with `world` programs is an orthogonal thing we have to address separately: Whether we address it or not, allowing backspace to be represented as `\b` in Pyret strings is still useful, even necessary.

This is what I do to have backspace treated correctly in `world` -- attach a `keypress` event to `on_key`, and have most keys treated by `keypress`, leaving only backspace (and compose too) to `keydown`. This lets text boxes made using `world` respond correctly to backspace and "European" characters. I have not made a PR for this yet.

```
    function on_key(press) {
      return function(thisWorldIndex) {
        var wrappedPress = function(e) {
          if (thisWorldIndex != worldIndex) { return; }
          if(e.keyCode === 27) { return; } // Escape events are not for world; the environment handles them
          if (e.type === 'keydown' && (e.key !== 'Compose') && (e.key !== 'Backspace')) {
            //regular alpha, not altgr or backspace
            return false; // try keypress instead
          }
          stopPropagation(e);
          preventDefault(e);
          change_world(function(w, k) { press(w, e, k); }, doNothing);
        };
        return {
          onRegister: function(top) {
            //http://www.w3.org/TR/html5/editing.html#sequential-focus-navigation-and-the-tabindex-attribue
            jQuery(top).attr('tabindex', 1);
            jQuery(top).focus();
            attachEvent(top, 'keypress', wrappedPress);
            //keydown event needed for backspace recognition, as it doesnt cause keypress
            attachEvent(top, 'keydown', wrappedPress);
          },
          onUnregister: function(top) {
            detachEvent(top, 'keypress', wrappedPress);
            detachEvent(top, 'keydown', wrappedPress);
          }
        };
      };
    }
    Jsworld.on_key = on_key;
```

jswrenn:
@blerner: Unicode contains a block called "[Control Pictures](https://en.wikipedia.org/wiki/Control_Pictures)" that contain symbols that represent (but are not themselves) control characters.

blerner:
I'm not asking as me. I'm asking as a confused student pestering a confused teacher. That glyph is meaningless to anyone except an expert. So is the empty rectangle Dorai mentions. If we're going to do this, we need to decide that strings are really computer sciencey strings, replete with non printable characters, rather than *text*, and we have to figure out a pedagogy for it.

Asked another way, why is backspace special and gets a control character, but the arrow keys get descriptive words? They have key codes too; should they be allowed into strings?

jswrenn:
> That glyph is meaningless to anyone except an expert.

Still, there's a unicode codepoint that represents this. I think that whatever the glyph for that codepoint happens to look like on a particular computer mostly orthogonal; we have control over the font and can substitute any other glyph.

ds26gte:
Pyret already lets you construct strings from all those characters based on their character codes, just not using escapes inside a literal. I can see your case that `\b` needn't be allowed (although then why `\n`, `\r`, `\t`?).  The reason I see `\b` has higher merit than similar escapes for the arrow and page keys is that it is routinely possible to embed this small ascii character (= 8) in text, but not (as far as I can tell) the arrow and page keys. (But this is admittedly a statistical argument based on mainstream text editors, and I would not rule out escapes for the arrow/page if folks want them.)

Be that as it may, try, in current Pyret:

```
string-from-code-point(8)  # backspace => '\b'
string-from-code-point(37) # left => '%'
string-from-code-point(38) # up => '&'
string-from-code-point(39) # right => "'"
string-from-code-point(40) # down => '('
string-from-code-point(33) # pgup => '!'
string-from-code-point(34) # pgdn => "\""
```

None of these results are particularly illuminating, although backspace is the best of the lot. Pyret already thinks of `\b` as a backspace representative, though it's half-minded about it. E.g., while I can get the string `"\b"` from `string-from-code-point(8)`, I can't actually _write out_ the string myself:

```
string-from-code-point(string-to-code-point(8)) => 8
string-from-code-point('\b') => error
```

(I have no particular commitment to see a tall rectangle as the displayed value of the backspace character, however it is created. But that is what is happening now: It is not my suggestion. A case can be made that these special characters (backspace, up arrow, etc) should all show up as a special glyph (or, more wastefully, distinct glyphs) when displayed, and a tall rectangle or a squat square is as good as any. I really don't think page-up and exclamation point should both display as the same glyph, which is what's happening now.)

blerner:
My main distaste for arrow keys and backspace (and delete) being single chars as represented in Pyret comes from the primary use I see of them: as inputs to on-key handlers.  A common world program to ask beginners to write is, "Build a text box: as the user types characters, the world state should be a string including them"  Then you refine it: add a cursor that can move, allow backspacing and deleting, etc.  The on-key handler is simply `on-key: lam(world, key): world + key end`.  But having arrow keys and backspaces etc all appear as single characters as key-arguments makes this simplistic handler seem like it keeps working, "except it's weird - when I type page-up I get an exclamation mark, so Pyret is broken..."  (I've gotten this reaction from students in Fundies, in Java and Racket; it's not farfetched.)  Having them show up as _strings_, saying "<left>", "<backspace>" etc, or whatever, makes it immediately apparent that these weren't likely to have been typed directly that way.

So much for their primary use case.  For creating these characters via escape codes: you're writing literals that aren't readily inputable, so it's kinda your own problem, sure.  But I really don't like either introducing arcane Unicode glyphs, or mandating that some characters can only be rendered as escape codes.  Newline, tab and carriage-return can all be rendered as whitespace _or_ as escape codes; backspace cannot, nor can arrow keys, page ups, or other control chars.

I don't know if I'm making too much out of a small detail, but I don't see a pressing need for this character, and don't want to make it easy to create in string literals.

jpolitz:
Closing as unnecessary – there are many control characters that we don't support specialized escapes for, and backspace will be one of them.  In world programs, the special value `"Backspace"` will be passed for `on-key`, which is consistent with shift, etc.

PR #1019, opened 2017-05-16
statistics module from horizon

jpolitz:


jpolitz:
Question for @SDooman : are these functions intended to only work on Numbers?  Legitimately curious about the intended design – modes is useful on string data or even more complex data, for example.  It affects the algorithmics of `modes` in the future if we know it takes numbers or not.

The return type is annotated as `List<Number>`, but the input is annotated as just `List`, so the code doesn't make it obvious to me.

shriram:
Why are there underscores? These should be called `index-of-median`, etc.

PR #1020, opened 2017-05-21
Fast mode

puppyofkosh:
This is the "fast-mode" branch

If we pass in -1 as the flatness-threshold, everything will be treated as flat. To get this to work I've had to change a few things:

1) The runtime now has a "bounceAllowed" flag which can be used to prevent bounces from happening (necessary when everything is flat)
2) I had to give more stack space to node (see changes in the makefile)
3) I had to change a test (see test-errors.arr) since runtime errors with fast-mode are much uglier

By default the makefile has fast mode off for the tests and for phaseB and so on, but I've run everything with fast mode on and it all seems to pass (test suite passes, phaseC builds fine).

Some performance numbers on my machine:
To build phaseC without any flatness took 6:47.66
To build phaseC with fast mode on (everything flat) took 5:15.37

There are a lot of changes here so I imagine you'll have a lot to say before it can get merged. I look forward to hearing your feedback!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,8 @@ PHASEB           = build/phaseB
 PHASEC           = build/phaseC
 RELEASE_DIR      = build/release
 
+FLATNESS_THRESHOLD = 5
+
 # CUSTOMIZE THESE IF NECESSARY
 PARSERS         := $(patsubst src/js/base/%-grammar.bnf,src/js/%-parser.js,$(wildcard src/$(JSBASE)/*-grammar.bnf))
 COPY_JS          = $(patsubst src/js/base/%.js,src/js/%.js,$(wildcard src/$(JSBASE)/*.js)) \
@@ -106,8 +108,8 @@ $(PHASEB)/pyret.jarr: $(PHASEA)/pyret.jarr $(PHASEB_ALL_DEPS) $(patsubst src/%,$
                       --builtin-arr-dir src/arr/trove/ \
                       --compiled-dir build/phaseB/compiled/ \
                       -no-check-mode $(EF) \
-                      --require-config src/scripts/standalone-configB.json
-
+                      --require-config src/scripts/standalone-configB.json \
+                      --flatness-threshold $(FLATNESS_THRESHOLD)
 
 .PHONY : phaseC
 phaseC: $(PHASEC)/pyret.jarr
@@ -119,7 +121,8 @@ $(PHASEC)/pyret.jarr: $(PHASEB)/pyret.jarr $(PHASEC_ALL_DEPS) $(patsubst src/%,$
                       --builtin-arr-dir src/arr/trove/ \
                       --compiled-dir build/phaseC/compiled/ \
                       -no-check-mode $(EF) \
-                      --require-config src/scripts/standalone-configC.json
+                      --require-config src/scripts/standalone-configC.json \
+                      --flatness-threshold $(FLATNESS_THRESHOLD)
 
 .PHONY : show-comp
 show-comp: build/show-compilation.jarr
@@ -242,8 +245,8 @@ tests/pyret/all.jarr: phaseA $(TEST_FILES) $(TYPE_TEST_FILES) $(REG_TEST_FILES) 
 	$(TEST_BUILD) \
 		--build-runnable tests/all.arr \
     --outfile tests/pyret/all.jarr \
-		-check-all
-		--flatness-threshold 5
+		-check-all \
+		--flatness-threshold $(FLATNESS_THRESHOLD)
 
 .PHONY : all-pyret-test
 all-pyret-test: tests/pyret/all.jarr parse-test
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 PYRET_COMP0      = build/phase0/pyret.jarr
 CLOSURE          = java -jar deps/closure-compiler/compiler.jar
-NODE             = node -max-old-space-size=8192
+NODE             = node -max-old-space-size=8192 --stack-size=8192
 SWEETJS          = node_modules/sweet.js/bin/sjs --readable-names --module ./src/js/macros.js
 JS               = js
 JSBASE           = $(JS)/base
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 PYRET_COMP0      = build/phase0/pyret.jarr
 CLOSURE          = java -jar deps/closure-compiler/compiler.jar
-NODE             = node -max-old-space-size=8192 --stack-size=8192
+NODE             = node -max-old-space-size=8192 --stack-size=16384
 SWEETJS          = node_modules/sweet.js/bin/sjs --readable-names --module ./src/js/macros.js
 JS               = js
 JSBASE           = $(JS)/base
@@ -16,6 +16,12 @@ PHASEC           = build/phaseC
 RELEASE_DIR      = build/release
 
 FLATNESS_THRESHOLD = -1
+ifeq ($(FLATNESS_THRESHOLD),-1)
+  # If we allow infinite flatness, then get rid of the system stack
+  # limit for node
+  TMPNODE:=$(NODE)
+  NODE=ulimit -s unlimited; $(TMPNODE)
+endif
 
 # CUSTOMIZE THESE IF NECESSARY
 PARSERS         := $(patsubst src/js/base/%-grammar.bnf,src/js/%-parser.js,$(wildcard src/$(JSBASE)/*-grammar.bnf))
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ PHASEB           = build/phaseB
 PHASEC           = build/phaseC
 RELEASE_DIR      = build/release
 
-FLATNESS_THRESHOLD = -1
+FLATNESS_THRESHOLD = 10
 ifeq ($(FLATNESS_THRESHOLD),-1)
   # If we allow infinite flatness, then get rid of the system stack
   # limit for node
--- a/Makefile
+++ b/Makefile
@@ -307,6 +307,7 @@ clean:
 .PHONY : test-clean
 test-clean:
 	$(call RMDIR, tests/compiled)
+	$(call RM, tests/pyret/*.jarr)
 
 # Written this way because cmd.exe complains about && in command lines
 new-bootstrap: no-diff-standalone
--- a/Makefile
+++ b/Makefile
@@ -243,6 +243,7 @@ tests/pyret/all.jarr: phaseA $(TEST_FILES) $(TYPE_TEST_FILES) $(REG_TEST_FILES) 
 		--build-runnable tests/all.arr \
     --outfile tests/pyret/all.jarr \
 		-check-all
+		--flatness-threshold 5
 
 .PHONY : all-pyret-test
 all-pyret-test: tests/pyret/all.jarr parse-test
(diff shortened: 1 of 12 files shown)


jpolitz:
If you have it running & it's convenient, how long does phaseC take to build with flatness set to, say, 5?  Curious how much might be coming from truly sync vs. the flatness optimization itself.  Starting a review; just wanted to get that idea out there.

Makefile
```diff
@@ -15,6 +15,14 @@ PHASEB           = build/phaseB
 PHASEC           = build/phaseC
 RELEASE_DIR      = build/release
 
+FLATNESS_THRESHOLD = 10
+ifeq ($(FLATNESS_THRESHOLD),-1)
+  # If we allow infinite flatness, then get rid of the system stack
+  # limit for node
+  TMPNODE:=$(NODE)
+  NODE=ulimit -s unlimited; $(TMPNODE)
```
jpolitz: @blerner do you know what the equivalent of this would be on Windows (not urgent)

src/arr/compiler/cli-module-loader.arr
```diff
@@ -425,16 +425,18 @@ fun build-runnable-standalone(path, require-config-path, outfile, options) block
...
+          standalone-end = time-now() - stats.get-value-now("standalone")
+          stats.set-now("standalone", [list: "Outputing JS: " + tostring(standalone-end) + "ms"])
+          for SD.each-key-now(key from stats):
+            print(key + ": \n" + stats.get-value-now(key).join-str(", \n") + "\n")
+          end
         end
       end
-      ans
+      MS.make-standalone(program.natives, program.js-ast,
+        JSON.j-obj(config.freeze()).serialize(), options.standalone-file, callback)
```
jpolitz: Ironically, on the `require-concat` branch (https://github.com/brownplt/pyret-lang/commit/7a3442cb115fcc6e4286f3ccf3e5c934088da9ea), `make-standalone` is now completely synchronous.  I don't think it's a bad idea to have this refactoring, though.

src/arr/compiler/compile-lib.arr
```diff
@@ -520,6 +520,13 @@ fun make-standalone(wl, compiled, options):
           end
       end
     end)
+
+  runtime-options = J.j-obj(
```
jpolitz: @blerner here's a place that has an opportunity to compile in stuff about whether to print check results at all (referring to an earlier conversation about the four combinations of “check mode on/off” for both compile and runtime.  Compile in an options object into the program that handalone sees.

src/js/base/runtime.js
```diff
@@ -3072,7 +3072,7 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
         stackFrame = $ar.args[2];
         $fun_ans = $ar.vars[0];
       }
-      if (--thisRuntime.GAS <= 0 || --thisRuntime.RUNGAS <= 0) {
+      if (thisRuntime.bounceAllowed && (--thisRuntime.GAS <= 0 || --thisRuntime.RUNGAS <= 0)) {
```
jpolitz: Oooh this is frustrating to see the copy/pasted pattern.  Hopefully this is simple enough that if we put it in a helper it would be inlined by the major engines.

src/js/base/runtime.js
```diff
@@ -5032,17 +5065,25 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
...
+                  var nativeInstantiated = Array.prototype.slice.call(arguments);
+                  restarter.resume(nativeInstantiated);
+                });
               });
-            });
+            } else {
+              /* require() should be synchronous */
+              var arr = [];
+              for(var i = 0; i < mod.nativeRequires.length; i++) {
+                var val = require(mod.nativeRequires[i]);
```
jpolitz: Making a note here that this requires that all native modules are not actually loaded asynchronously.  This rules out some uses cases (which we don't rely on at all, but theoretically could use with bounces allowed), where requirejs will fetch from a URL and asynchronously load.  There were past versions of the plot library with d3 that relied on this, for example.

src/js/base/runtime.js
```diff
@@ -5831,6 +5872,19 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
...
 
+    thisRuntime.bounceAllowed = true; // Default
+    if (theOutsideWorld.options) {
+      thisRuntime.bounceAllowed = theOutsideWorld.options.bounceAllowed;
+    }
+
+    // In principle it should be possible for a parent runtime to have bouncing off
+    // but a "child runtime" have bouncing on, though it might require some nasty
+    // modifications to the run function. For now we just require that the child runtime
+    // can only allow bounces if the parent runtime can as well
```
jpolitz: Oof yeah.  The child runtime would have to be synchronous still, right?  So it would have to be synchronous and bouncy.

Is there a way to do asynchronous bounces in a runtime created from a from a non-bouncing one?  I can't think of how that would work.

src/js/trove/load-lib.js
```diff
@@ -179,46 +179,60 @@
         };
         var getStackP = execRt.makeFunction(getStack, "get-stack");
         var checks = getModuleResultChecks(mr);
-        execRt.runThunk(function() { return toCall.app(checks, getStackP); },
-          function(renderedCheckResults) {
-            var resumeWith = {
-              message: "Unknown error!",
-              'exit-code': EXIT_ERROR_UNKNOWN
-            };
+        //      <<<<<<< b7f9c9016754ed2a86e852980708822e67ff1fed
```
jpolitz: Is this comment providing useful information?

src/js/trove/load-lib.js
```diff
@@ -310,31 +345,48 @@
...
+            retVal = makeModuleResult(otherRuntime, finalResult, makeRealm(realm), runtime.nothing);
           }
-        });
-      });
 
+          if (runtime.bounceAllowed) {
+            restarter.resume(retVal);
+          } else {
+            return retVal;
+          }
```
jpolitz: This pattern seems to come up a lot; I wonder if there could be a runtime method or a method on the restarter that abstracts this away so we can just write `return restarter.resumeOrReturn(retVal);`

In the call to `pauseFunction`, maybe we should pass in an argument to `pauseFunction` that has a dummy restarter?

It seems like there's a pattern here to abstract one way or another.

Another idea is to have `runtime.maybePauseStack` which simply calls the callback with the right dummy restarter if `bounceAllowed` is false, or does the normal pausing flow otherwise.  If a part of the program _will not work_ without pausing, it could use `runtime.pauseStack`, which will throw an error if `bounceAllowed` is false.

tests/pyret/tests/test-errors.arr
```diff
@@ -98,14 +98,18 @@ check:
 #    e5.typ is "Number"
 
   e10 = get-err(lam(): 5() end)
-  e10 satisfies E.is-non-function-app
+  e10 satisfies E.is-RuntimeError
+  # TODO: Test should only be relaxed in unsafe mode
```
jpolitz: I think with sourcemaps we can get this right.

puppyofkosh:
Thanks for the feedback!

I re-ran building phaseC with flatness as 5 and 10 and got:
flatness 5: 6:55.92
flatness 10: 6:29.01

This is strange considering what I got when I had flatness off earlier, (6:55) though I recorded that number earlier in the week, and could have been using a different version or something.

I'll get to work on your comments tomorrow!

jpolitz:
The difference b/w those three measurements seems like noise, to be honest, which is interesting in itself.

src/arr/compiler/cli-module-loader.arr
puppyofkosh: Oh, nice. Then when that gets merged in we can get rid of some ugly callback code I had to add

src/js/base/runtime.js
puppyofkosh: Changed this to use a helper (two actually)

src/js/base/runtime.js
puppyofkosh: Got it. I added a comment

src/js/base/runtime.js
puppyofkosh: Oh wow, you're definitely right. I don't see any way we could do bounces from a "child" runtime which has a non-bouncing parent. I added a comment explaining this.

src/js/trove/load-lib.js
puppyofkosh: My bad. Removed!

src/js/trove/load-lib.js
puppyofkosh: I added a dummy restarter and did what you described (I think) and the code in load-lib is quite a bit cleaner now.

puppyofkosh:
Regarding the timings: the 25 second difference seems like kind of a lot to just be noise, but we have experienced inconsistent timings for even really simple programs. Maybe another metric would be # of times each optimization is applied?

Something else I'm realizing is that testing this code is kind of awkward since we have to be sure everything works with fast mode both on and off. Right now I've been pushing builds to travis with fast mode off, and running the tests locally with fast mode on to be sure they pass in both cases, but we should figure out a better way to deal with this. Should the travis servers run the tests with fast mode both on and off? This might be annoying since it would double the amount of time a build takes.

src/js/base/runtime.js
```diff
@@ -3033,6 +3033,15 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
       return stackStr;
     }
 
+    /* Decrement gas or rungas and return whether or not it's time to return a continuation */
+    function spendGas() {
```
jpolitz: Good names.

jpolitz: Good names.

puppyofkosh: =)

src/js/base/runtime.js
```diff
@@ -3568,7 +3619,19 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
     }
 
     function runThunk(f, then) {
-      return thisRuntime.run(f, thisRuntime.namespace, {}, then);
+      if (thisRuntime.bounceAllowed) {
+        return thisRuntime.run(f, thisRuntime.namespace, {}, then);
+      } else {
+        // Just run it on this stack and use a try/catch handler
```
puppyofkosh: I moved some things around so this works now

jpolitz:
Does this rely on #991 or is it merge-able on its own?

puppyofkosh:
You should be fine just merging this. I closed #991

jpolitz:
@puppyofkosh can you try rebasing this onto `sourcemaps` to see if it's easy?  I'm curious if that would fix the error-based tests, as well.

puppyofkosh:
Looks like rebasing on sourcemaps fixes those tests! I made a separate pull request here: #1031 since there were a fair number of merge conflicts.

jpolitz:
Question – I think this has two different pieces:

1.  Using infinite flatness to trigger the optimization everywhere, with corresponding runtime changes for when flatness is known to be infinite
2.  A new optimization that always applies, which merges adjacent flat blocks together

puppyofkosh:
This PR contains (1) but (2) should come with the original flatness optimization. If (2) is referring to the `insert-flat-code` function I added, that was just a refactoring I did so that the code for generating flat function calls didn't need to be duplicated for flat method calls.

jpolitz:
Closing this because an updated version is at https://github.com/brownplt/pyret-lang/pull/1054

PR #1021, opened 2017-05-25
[SD] rewrite linear regression interface

SDooman:
**TLDR**:  While writing the scatter-plot/linear regression unit in Bootstrap Data Science, Emu and I have decided that the current interface for linear regression is overly complicated for beginners and their teachers.

[Link to earlier conversation about the interface for linear regression](https://github.com/brownplt/pyret-lang/pull/864)

There were originally 3 proposals for what a `linear-regression` function should return.  Right now, Pyret follows option 3.

1.  `linear-regression :: (x :: List<Number>, y :: List<Number>) -> (Number -> Number)`
       returns a lambda, which is the linear predictor function
2.   `linear-regression :: (x :: List<Number>, y :: List<Number>) -> {Number, Number}`
        returns a tuple `{alpha: ..., beta:  ...}`, giving the coefficients of the linear predictor function
3.  ` linear-regression :: (x :: List<Number>, y :: List<Number>) -> StatModel`
       returns a StatModel, which is a custom data type for statistical models.

The following concepts are required to create a scatter plot & overlay a line of best fit, in the current paradigm:
 - 3 different functions (`scatter-plot`, `function-plot`, `render-multi-plot`)
 - use of dot operator/concept of objects
 - 2 types (`Plot`, `StatModel`)
 - passing functions as arguments

With this PR (which reverts to option 1), they would need the following:
 - 5 different functions (`scatter-plot`, `function-plot`, `render-multi-plot`, `linear-regression`, `r-squared`)
 - 1 type (`Plot`)
 - passing function as argument

@schanzer and I believe we should prioritize reducing the cognitive burden on students/teachers, rather than the modularity of the statistical modeling capabilities of Pyret (especially considering we do not yet have plans for more models than the simple linear regression).

As for choosing between returning a lambda VS a tuple with coefficients, we prefer option 1 because it reduces the number of functions required to create a scatter plot.  5 functions is already a lot for something that math teachers will want to do fairly easily.

### DIFF
--- a/src/arr/trove/statistics.arr
+++ b/src/arr/trove/statistics.arr
@@ -10,33 +10,6 @@ import valueskeleton as VS
 import lists as L
 import math as math
 
-# A StatModel is a data type for representing
-# statistical models.  We create this type so that
-# users can extract, coefficients and meta-data from
-# models, as well as apply the model.
-#
-# This could be extended to include general linear 
-# regression, exponential and logistic regression, etc.
-
-data StatModel:
-  | simple-linear-model(alpha :: Number, beta :: Number, r-sqr :: Number) with:
-
-    method predictor(self :: StatModel) -> (Number -> Number):
-      doc: "the function predicting the value of a dependent variable"
-      lam(X): (self.beta * X) + self.alpha end
-    end,
-
-    method apply(self :: StatModel, l :: L.List<Number>) -> L.List<Number>:
-      doc: "applies the predictor function to a sample set of the independent variable"
-      L.map(self.predictor(), l)
-    end,
-
-    method r-squared(self :: StatModel) -> Number:
-      doc: "gives the coefficient of correlation"
-      self.r-sqr
-    end
-end
-
 fun mean(l :: L.List<Number>) -> Number:
   doc: "Find the average of a list of numbers"
   if L.length(l) == 0:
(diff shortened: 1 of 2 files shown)


src/arr/trove/statistics.arr
```diff
@@ -139,7 +112,7 @@ fun stdev(l :: L.List) -> Number:
   num-sqrt(sq-mean)
 end
 
-fun lin-reg-2V(x :: L.List<Number>, y :: L.List<Number>) -> StatModel:
+fun linear-regression(x :: L.List<Number>, y :: L.List<Number>) -> (Number -> Number):
   doc: "returns a linear regression model calculated with ordinary least squares"
   if x.length() <> y.length():
     raise("lin-reg-2V: input lists must have equal lengths")
```
jpolitz: Should change this error to refer to new name

src/arr/trove/statistics.arr
```diff
@@ -155,18 +128,24 @@ fun lin-reg-2V(x :: L.List<Number>, y :: L.List<Number>) -> StatModel:
     beta = covariance / variance
     alpha = mean(y) - (beta * mean(x))
 
-    y-mean = mean(y)
-    f = L.map(lam(xi): (beta * xi) + alpha end, x)
-    ss-tot = math.sum(L.map(lam(yi): num-sqr(yi - y-mean) end, y))
-    ss-res = math.sum(L.map2(lam(yi, fi): num-sqr(yi - fi) end, y, f))
+		fun predictor(in :: Number) -> Number:
```
jpolitz: Weird indentation.

src/arr/trove/statistics.arr
```diff
@@ -155,18 +128,24 @@ fun lin-reg-2V(x :: L.List<Number>, y :: L.List<Number>) -> StatModel:
...
-      1 - (ss-res / ss-tot)
-    end
- 
-    simple-linear-model(alpha, beta, r-sqr)
+		predictor
+	end
+end
+
+fun r-squared(x :: L.List<Number>, y :: L.List<Number>, f :: (Number -> Number)) -> Number:
+	y-mean = mean(y)
```
jpolitz: Weird indentation.

PR #1023, opened 2017-06-01
Sourcemaps stacktrace tests

puppyofkosh:
I added a few tests for sourcemaps. Most of them are pretty simple, testing calling a non function or non method, or accessing a field that doesn't exist. It also tests errors from within flat functions, to be sure that we get complete stack traces.

When testing the stack trace of a function that's tail recursive, we only check that the top and bottom frames are correct, since the middle "frames" are omitted by the optimization. This might be controversial so any thoughts are welcome.

I also made some changes in anf-loop-compiler. There are a few places where code that might throw an exception was not sourcemapped and so the stack trace would be wrong or incomplete. I added a few j-sourcenodes in places that I thought could use them.

### DIFF
--- a/src/arr/compiler/anf-loop-compiler.arr
+++ b/src/arr/compiler/anf-loop-compiler.arr
@@ -221,8 +221,15 @@ fun get-dict-field(obj, field):
   j-bracket(j-dot(obj, "dict"), field)
 end
 
-fun get-field(obj :: J.JExpr, field :: J.JExpr, loc :: J.JExpr):
-  j-app(get-field-loc, [clist: obj, field, loc])
+# Use when we're sure the field will exist
+fun get-field-unsafe(obj :: J.JExpr, field :: J.JExpr, loc-expr :: J.JExpr):
+  j-app(get-field-loc, [clist: obj, field, loc-expr])
+end
+
+# When the field may not exist, add source mapping so if we can't find it
+# we get a useful stacktrace
+fun get-field-safe(l, obj :: J.JExpr, field :: J.JExpr, loc-expr :: J.JExpr):
+  J.j-sourcenode(l, l.source, get-field-unsafe(obj, field, loc-expr))
 end
 
 fun get-field-ref(obj :: J.JExpr, field :: J.JExpr, loc :: J.JExpr):
@@ -1369,7 +1376,7 @@ compiler-visitor = {
   end,
   method a-dot(self, l :: Loc, obj :: N.AVal, field :: String):
     visit-obj = obj.visit(self)
-    c-exp(get-field(visit-obj.exp, j-str(field), self.get-loc(l)),
+    c-exp(get-field-safe(l, visit-obj.exp, j-str(field), self.get-loc(l)),
       cl-snoc(visit-obj.other-stmts, j-expr(j-assign(self.cur-apploc, self.get-loc(l)))))
   end,
   method a-colon(self, l :: Loc, obj :: N.AVal, field :: String):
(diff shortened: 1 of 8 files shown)


src/arr/compiler/anf-loop-compiler.arr
```diff
@@ -217,12 +217,27 @@ fun obj-of-loc(l):
...
+
 fun get-dict-field(obj, field):
   j-bracket(j-dot(obj, "dict"), field)
 end
 
-fun get-field(obj :: J.JExpr, field :: J.JExpr, loc :: J.JExpr):
-  j-app(get-field-loc, [clist: obj, field, loc])
+# Use when we're sure the field will exist
+fun get-field-unsafe(obj :: J.JExpr, field :: J.JExpr, loc-expr :: J.JExpr):
+  j-app(get-field-loc, [clist: obj, field, loc-expr])
```
jpolitz: In this case, I bet we could do `j-dot(j-dot(obj, "dict"), field)`

jpolitz: Oh, I take it back.  Methods and `ref` fields.  My bad.

src/js/trove/load-lib.js
```diff
@@ -97,6 +97,16 @@
...
     }
+    function getResultStackTrace(mr) {
+      var runtime = mr.val.runtime;
+      if (!runtime.isFailureResult(mr.val.result)) {
+        runtime.ffi.throwMessageException("Tried to get stacktrace of non-failing module result.");
+      }
+
+      var res = getModuleResultResult(mr);
+      var stackString = runtime.printPyretStack(res.exn.pyretStack);
+      return runtime.makeString(stackString);
```
jpolitz: Would getting back a `raw-array` be more useful, split on new lines?  Just thinking about possible interfaces.

jpolitz:
Awesome work.

Two other things I'd like to see here:

- Traces through higher-order functions like `map`
- Traces after bouncing – I realized we're probably missing a case here.  What about calling something that recurses (non-tail) for a while to cause the stack to be collected, _then_ throws an exception.

jpolitz:
Also traces through higher-order builtins like `raw-list-map`, which is different from `map` in lists because there is no Pyret code involved at all.

src/js/base/exn-stack-parser.js
```diff
@@ -30,9 +30,40 @@ define("pyret-base/js/exn-stack-parser", [], function() {
     return lines.filter(isSourcePyretFrame).map(parseFrame);
   }
 
+  function convertExceptionToPyretStackTrace(e, program) {
+    var parsedStack = parseStack(e.stack);
```
puppyofkosh: This was duplicated in handalone and load-lib

src/js/base/exn-stack-parser.js
```diff
@@ -30,9 +30,40 @@ define("pyret-base/js/exn-stack-parser", [], function() {
...
+        sourceMap.SourceMapConsumer.LEAST_UPPER_BOUND);
+      console.log(original);
+      var posForPyret = original.name.split(",");
+      for(var i = 1; i < posForPyret.length; i += 1) {
+        posForPyret[i] = Number(posForPyret[i]);
+      }
+      return posForPyret;
+    });
+
+    if (e.pyretStack) {
```
puppyofkosh: This is the only actual "new" code here. It appends the pyret stack which was converted from a javascript stack (using sourcemaps) to the stack frames which were collected during bounces.

src/js/base/runtime.js
```diff
@@ -7,6 +7,7 @@ define("pyret-base/js/runtime",
    "pyret-base/js/secure-loader",
    "seedrandom"],
 function (Namespace, jsnums, codePoint, util, exnStackParser, loader, seedrandom) {
+  Error.stackTraceLimit = Infinity;
```
puppyofkosh: Had to add this for some of the tests with larger stack traces. If you think there's a better place for it let me know

tests/pyret/tests/test-repl.arr
```diff
@@ -138,6 +145,118 @@ check:
...
+  # stacktrace through list.map()
+  result45 = restart("fun f():\n" +
+    "h = lam(x): 9() end\n" +
+    "[list: 1, 2, 3].map(h)\n" +
+    "end", false)
+  result46 = next-interaction("f()")
+  result46.v satisfies L.is-failure-result
+  stacktrace46 = L.get-result-stacktrace(result46.v)
+
+  raw-array-length(stacktrace46) is 5
```
puppyofkosh: Here are the new tests which should do what you suggested

puppyofkosh:
Okay I added some new tests (the second one you suggested is particularly fun I think. I added some code to "append" the two stacks we have (see comment)).

There was a bit of trouble with the stack trace size limit being small by default, but following your comment in #random I changed it to infinity.

PR #1024, opened 2017-06-01
Horizon stacktrace tests

puppyofkosh:
These are the same tests that I made for the sourcemaps branch.

See https://github.com/brownplt/pyret-lang/pull/1023


jpolitz:
This seems like an obvious good thing to have no matter what, merging these in.

PR #1027, opened 2017-06-11
Sourcemaps

jpolitz:
Source map changes, creating a PR in order to accurately compare.

Relevant descriptions are in:

ae1250e6441277ce1066fd09496e56b512949122
558481de7fd481bae0abeca9803c61ac32e315f5
93568bdc38692363a92f759198092f2b54c8f7bf


src/js/base/runtime.js
```diff
@@ -1795,10 +1803,12 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
       if(thisRuntime.isObject(val) &&
          (thisRuntime.hasField(val, "render-reason")
           || thisRuntime.hasField(val, "render-fancy-reason"))){
-        throw new PyretFailException(val);
+        var err = new PyretFailException(val);
       } else {
-        throw new PyretFailException(thisRuntime.ffi.makeUserException(val));
+        var err = new PyretFailException(thisRuntime.ffi.makeUserException(val));
       }
+      err.stackAtThrow = new Error().stack;
```
blerner: This `stackAtThrow` field is only ever used once -- is this a typo?

src/js/base/runtime.js
```diff
@@ -3352,6 +3362,7 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
               // CONSOLE.log("Frame returned, val = " + JSON.stringify(val, null, "  "));
             }
           } catch(e) {
+            debugger;
```
blerner: Stray `debugger`?

src/js/base/runtime.js
jpolitz: It turns out this isn't necessary.  Thanks.  I had included this for debugging but the `stack` field on the exception itself turned out to be sufficient.

jpolitz:
An overview of how exceptions now work:

## Stack Management

Exceptions representing continuations are handled by checking the `$ans`
variable after the main `while` loop.  This was the strategy originally
introduced on the `return-stack` branch, and removes the stack-management need
for exception handlers.  In theory, this is orthogonal to how sourcemaps work.
In practice, these two changes happened together, and sourcemaps are largely
useful because they are part of a strategy to avoid needing catches at every
frame.

## Exceptions 

### In `runtime.js`

All exceptions, both those representing errors and those representing
continuations, are caught by `runtime.run`.  Continuation exceptions get the
usual treatment in the `isCont` branch.  Error exceptions, from the point of
view of `run`, also have unchanged treatment, but it's useful to refresh our
memory that when an error is caught, we copy all of `theOneTrueStack` into
`e.pyretStack`.  This ensures that pending frames from paused computation
(either due to stack depth or `pauseStack`) are included in the stack later.
The key difference now is that the `pyretStack` of these exceptions is _always
empty_ before this happens.

### When reporting errors

The rest of the stack—the frames currently in progress on the real JS stack
when the error occurred—are recovered from the `.stack` field of the
exception.  The work for this happens in `src/js/base/exn-stack-parser.js`,
which combines information from (a) the function names in the stack trace and
(b) the source maps stored along with programs, in order to generate a stack
trace with corresponding Pyret locations.  This happens in `load-lib` and in
`handalone.js`.

## Generating Source Maps and Function Names

In order to get Pyret-level stack trace information from error stacks, we need
to store information for source maps, and also some special function names to
find the Pyret frames within stack strings.

For each function we create from a Pyret function, we add a special name that
contains the SHA-256 hash of the module URI the function came from.  These
stack frames stand out in format:

```
      _781694b32b8821611a0852dc4238c1f1a3cc3fd414551f52156ec10cbbc609ab__0@file:///Users/joe/src/pyret-lang/test.html:93:10
``` 

The code that reconstitutes the stack frames first filters for only frames that
contain `_<SHA>__`-shaped strings (`isSourcePyretFrame`).  It's easy to extract
the hash from those strings, and the standalone program bundle contains a map
from hashes back to URIs.  The generation of these names happens in
`anf-loop-compiler.arr` at all the calls to `make-fun-name`.  Note that the
number after the `__` in the function name corresponds to the index in the
source location list that the function _definition_ appeared at; this is
currently unused.

To produce a meaningful stack trace, we actually want _caller_ locations, and
all we can encode in the function name is _callee_ information.  The line and
column at the end of the stack frame report above are relative to the
JavaScript code that ran for the given module.  Along with each generated
JavaScript module, we now create a source map that maps all application
expressions generated from Pyret function applications back to the original
Pyret location.  The infrastructure for doing this is in `js-ast.arr` in the
`j-sourcenode` variant, and in `source-map-lib.js`.

Note that sourcemaps typically only track start line and start column
information.  This wouldn't typically be sufficient for tracking all the
information about spans that Pyret needs.  However, source maps also allow
tracking of a string intended to describe the original function name.  We use
this facility to encode all 6 numbers from a Pyret location (start line, col,
char, end line, col, char) into one string.  This information is then encoded
into the source map.  This generated map lets us map from _generated_
line/column numbers (as in the stack trace above) to _Pyret_ source locations
with all the necessary span information.

Issue #664, opened 2016-03-03
printing of roughnums considered very mildly harmful


shriram:
I wonder if we're printing roughnums quite right. When we print
```
~75.92759709091287
```
that sounds like an extreme degree of precision, so an error message like
```
  Attempted to compare two Roughnums for equality, which is not allowed:
  ~75.92759709091287
  ~75.92759709091287
```
looks really odd. I wonder if we can just add a little `...` at the end:
```
  Attempted to compare two Roughnums for equality, which is not allowed:
  ~75.92759709091287...
  ~75.92759709091287...
```
meaning "there's probably stuff here, we just don't know", so it's a bit less likely that a reader will think "wait, these are obviously the same!".

This potentially has implications for input. I feel they don't have to write `...` on the input. That's just a PITA.

ds26gte:
> This potentially has implications for input. I feel they don't have to write ... on the input.

So the `...` is not just displayed in the error message, but is part of the roughnum literal syntax? Hope we don't go there.

How about the first `...` replaced by ` ± δ₁` and the second by ` ± δ₂`. The different suffixes should be enough of a hint that the two numbers are probably unequal and only possibly equal, and that in any case they are only a tiny δ apart.

shriram:
No, I'm explicitly saying `...` is _not_ part of the literal syntax. (Though I don't mind if a person wants to put it at the end…)

We can't use all your fancy glyphs. We're ASCII.

schanzer:
Are we _really_ ASCII? Given that we already print 1/3 with the bar on top, it seems like that horse is already out of the barn...

shriram:
For non-string input, yes, we are.

Also, @ds26gte, I would not assume the average reader knows what your notation means — remember, our target audience includes middle-school kids.

schanzer:
This is utterly compelling. I'm sold on `...` !

blerner:
There's a small part of me that's not convinced: if I have roughly one half, it'd be weird for me to see `~0.5000000000...` as it violates my implicit assumption of "show me only the digits that are meaningful".  But on the other hand, `~0.33333333333....` seems useful.  I'm not rejecting this; just mentioning an odd-ish corner case.

schanzer:
~0.5 currently prints as `~0.5`

Why not just append `...` to the existing roughnum printout?

blerner:
I'm totally fine with that, then.  Thanks :)

ds26gte:
PR #1028 submitted.

[...]

PR #1028, opened 2017-06-12
equality-failure: if comparing two roughnums, mark them with ellipses

ds26gte:
Fix for #664. When two roughnums are compared for equality, the error message disallowing this shows the offending roughnums with `...` appended. This minimizes user confusion when two equal-seeming roughnums are compared.

### DIFF
--- a/src/arr/trove/error.arr
+++ b/src/arr/trove/error.arr
@@ -1744,19 +1744,23 @@ data RuntimeError:
     method render-reason(self):
       value1 = self.value1
       value2 = self.value2
+      var svalue1 = num-to-string(self.value1)
+      var svalue2 = num-to-string(self.value2)
       ask:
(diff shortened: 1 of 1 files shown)


blerner:
I think this is the wrong place to handle fixing #664 -- seems like the consensus was to change how roughnums render, always, to include the ellipsis.  That belongs in js-numbers's `Roughnum.prototype.toString`, no?

shriram:
@blerner is correct. This was meant to address _all_ rendering of roughnums, not only at equality checks. Equality is just where it manifests most aggressively as a problem.

ds26gte:
New fix is at brownplt/code.pyret.org#178.

Chose to modify the renderer in `code.pyret.org` rather than `Roughnum.prototype.toString` in `pyret-lang` because the latter is called by things other than the renderer, and changing it would break many tests.

PR #1029, opened 2017-06-13
roughnums rendered with trailing ellipsis; input syntax can have optional trailing ellipsis.

ds26gte:
Fix for #664.

This change, unlike my previous attempts, does not touch `code.pyret.org`. The default renderer in `runtime.js` adds a trailing ellipsis. In addition, `pyret-tokenizer.js` and `js-numbers.js` allow an optional trailing ellipsis to tacked on to the roughnum literal input syntax.

(`Roughnum.prototype.toString` continues to present a roughnum without trailing ellipsis, as otherwise, many tests that convert numbers to strings fail. Changing the tests is a way to go, but I think there is value in retaining `toString()`'s non-redundant representation. We seem to want users to continue to be able to use the non-redundant syntax for input. If there is interest in making the ellipsis non-optional for input too, that can be done -- again that will require widespread rewriting of tests.)

### DIFF
--- a/src/js/base/runtime.js
+++ b/src/js/base/runtime.js
@@ -161,7 +161,6 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
       delete parameters[param];
     }
 
-
     /**
        Get the brands on an object
 
@@ -265,7 +264,14 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
 
     var DefaultReprMethods = {
       "string": String,
-      "number": String,
+      //"number": String,
+      "number": function(val) {
+        if (jsnums.isRoughnum(val)) {
+          return val.toString() + '...';
+        } else {
+          return String(val);
+        }
+      },
       "boolean": String,
       "nothing": function(val) { return "nothing"; },
       "function": function(val) { return "<function>"; },
@@ -289,7 +295,7 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
         }
         s += " }";
         return s;
-        
+
       },
       "object": function(val, pushTodo) {
         var keys = [];
@@ -419,7 +425,7 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
         if (val.$constrFor !== undefined) {
           thisRuntime.ffi.throwLookupConstructorNotObject(makeSrcloc(loc), val.$constrFor, field);
         }
-        thisRuntime.ffi.throwLookupNonObject(makeSrcloc(loc), val, field); 
+        thisRuntime.ffi.throwLookupNonObject(makeSrcloc(loc), val, field);
       }
       var fieldVal = val.dict[field];
       if(fieldVal === undefined) {
@@ -1043,7 +1049,6 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
       return true;
     }
 
-
     /*********************
             Object
     **********************/
@@ -1171,7 +1176,7 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
     function makeArray(arr) {
       return arr;
     }
- 
+
     /************************
           Type Checking
     ************************/
@@ -1842,7 +1847,7 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
             return;
           }
         }
-// throw new Error("Internal error: tried to 
+// throw new Error("Internal error: tried to
       }
       function cachePair(obj1, obj2) {
         cache.left.push(obj1);
@@ -2489,7 +2494,6 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
       });
     }
 
-
     function _checkAnn(compilerLoc, ann, val) {
       if (isCheapAnnotation(ann)) {
         var result = ann.check(compilerLoc, val);
@@ -2742,8 +2746,6 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
       });
     }
 
-
-
     function PTupleAnn(locs, anns) {
       this.locs = locs;
       this.anns = anns;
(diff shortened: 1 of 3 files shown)


src/js/base/runtime.js
```diff
@@ -265,7 +264,14 @@ function (Namespace, jsnums, codePoint, seedrandom, util) {
 
     var DefaultReprMethods = {
       "string": String,
-      "number": String,
+      //"number": String,
+      "number": function(val) {
+        if (jsnums.isRoughnum(val)) {
+          return val.toString() + '...';
```
blerner: Does this work on primitive doubles, or does this need to be `String(val) + '...'` instead?

ds26gte: Since it's conditionalized on `val` being a roughnum (a properly boxed number and never a primitive double), which is guaranteed to have a `toString()` method (that tacks on the prefix `~`), this is fine.

jpolitz:
I'm concerned about `...` as input, because `...` also means “unfinished template expression.”

ds26gte:
Pyret's been rigorous about requiring inter-token spaces around operators where most other infix languages don't. Why not require a space before the three-dots meaning unfinished template?

blerner:
I do like that idea. But Joe's suggested concern is about exactly that case. The idea that the same lexeme, differing only on its surrounding whitespace, could drastically change the meaning of the program, seems a bit too severe in this case. Yes, `a-b` is a name and `a - b` is subtraction, and those only differ by whitespace surrounding some punctuation marks, but somehow those feel less punitive than saying "Error: not yet implemented".

ds26gte:
OK, the only "punitive" scenario I see here is when the user gets burned by mistyping a roughnum-with-ellipsis with a space before the ellipsis. That's the only time  they'd be punished by an "Error: not yet implemented" when they were thinking they wrote a roughnum.

But, remember: We don't really expect users to use the optional ellipsis actively when entering roughnums. It is primarily there to ensure that the printed output form of a roughnum remains self-evaluating. Users will not typically be writing `~5...` in their programs (they would write just `~5`), and if they did write `~5 ...` with the intervening space, they likely meant it to be two separate tokens. (Even that is a very rare template spec: so whoever's writing it can be expected to be more than usually circumspect about syntax. So it's a hard argument that they may have made the reverse typo, i.e., meant to type `~5 ...` but ended up with `~5...`.) 

There's really precious little wiggle room if 

1. we want @shriram's syntax for roughnums, _and_ 
2. we want it to be self-evaluating, _and_ 
3. we want the ellipsis to be three ascii dots (and not, say, U+2026, given we want to steer clear of Unicode). 

One way out was my original approach (nixing № 2) of having the ellipsis only in the error _diagnostic_, but that didn't meet consensus. Perhaps that's looking better now?

shriram:
I think the real problem here is the content and quality of the `...` error message, which sounds like it should be fixed anyway (CC @jswrenn). As a user I'm not sure what "not implemented" even means: was it not implemented by me, the programmer, or by Pyret, the system (i.e., it's the equivalent of a known internal error)? A slightly longer error message can say something about roughnums.

blerner:
The semantics of ... is to desugar to a `raise "Not yet implemented"` expression. You can't get a different message about roughnums from that. If you write `~5 ...` with a space, on one line, I believe you get a well formedness error about multiple expressions on one line. I'm not a huge fan of complicating that check to say "well, maybe this isn't well formed; maybe you meant not to have a space here", but that's the only good place to have some very ad hoc heuristic here.

shriram:
Okay. Then we have to just improve the error message in the raise as best as we can.

I am not too concerned about read-write equivalence. We make a pretty big deal out of it, but

1. I don't know how much students actually do it, and

2. the un-copyability of our output printing has been a nightmare for _years_ now, so this seems like the wrong issue to be stressing about so much. Just now, I typed
```
> [list: 1, 2]
```
and got back out something that, when copied, renders as
```
[list: 12]
```
So I don't think our concern on this thread about pasting output as input has at all been matched by the implementation (despite this issue having been reported many times).

blerner:
Please don't combine this issue with your justified gripe about lists not having commas in the copied output. That's a separate problem, requiring a fairly minor css fix and some crappy html fixing, and had never been a priority compared to the other more important stuff. Put an undergrad on it, and it can be fixed in an hour.

This issue is about the ergonomics of ... as part of roughnum lexing, and I suspect that until you have a concrete proposal of what the error wording is that you'd like, this thread will spiral around with no conclusion, with Dorai implementing one this, you expecting another, and me expecting the engineering of it to be undesirable overall.

shriram:
I brought that in because we're using that as a design principle, but don't adhere to that principle ourselves in very obvious ways.

As for the error wording, I brought @jswrenn into this because he's been the keeper of error messages. It's fair to ask for a concrete proposal, but I'd like Jack to come up with one that is consistent with all our other messages.

jpolitz:
We reached a consensus:

- `...` will be optional in the input syntax on `~` numbers
- `...` will be disallowed in the input syntax on non-`~` numbers
- `...` will be _always_ be rendered after a printed roughnum, which will also be rendered with a leading `~`

We should also add ellipsis characters to the find-and-replace on paste list along with smart quotes.

blerner:
@ds26gte I think the bad-numbers syntax should be
```js
  const roughnum_numbers_part = "(?:" + unsigned_rational_part + "|" + unsigned_decimal_part + ")";
  const badNumber = new RegExp(
    "^(?:~?[+-]?\\.[0-9]+(?:[eE][-+]?[0-9]+)?)|(?:" + roughnum_numbers_part + "(\.(\.(\.\.+)?)?)?)", STICKY_REGEXP);

  const roughnum = new RegExp("^~[-+]?" + roughnum_numbers_part, STICKY_REGEXP);
```

This is *close* -- there's going to be some interaction between roughnums and trailing dots and badnumbers.  I.e. the tricky cases are `~2` should be a roughnum, `~2.` should be an error, `~2..` is an error, `~2...` is a roughnum, `~2....` should be an error again.

PR #1030, opened 2017-06-14
Sourcemaps html standalone

puppyofkosh:
Don't actually merge this! It's just here for Joe to review.

I put comments in the places that I think are worth looking at. Don't bother looking at the other stuff.


browserify/browserify-require-combined.js
```diff
@@ -0,0 +1,21 @@
+sexpr = require("s-expression");
```
puppyofkosh: We run browserify on this file

browserify/build.sh
```diff
@@ -0,0 +1,2 @@
+browserify browserify-require-combined.js > nodemods.js
```
puppyofkosh: The build process is:
Run browserify on the browserify-require-combined.js (bad name, I know). This gives us all of the node_modules dependencies in one file. Then we paste that in front of the actual jarr file and we have a pyret standalone with no dependencies.

browserify/my-loader.js
```diff
@@ -0,0 +1,167 @@
+var define, myrequire;
```
puppyofkosh: This is meant to do roughly what requirejs does.

puppyofkosh:
I've cleaned up this branch quite a bit, so I think it's in a more reviewable state now. I'm leaving Providence Wednesday, so some of this work was rushed together (discussed below), though it does "work."

**These are the main changes in this branch:**
1) It removes requirejs as the module loader and replaces it with a handwritten one
2) It adds two compiler flags:
   `-bundle-dependencies`: Produce a standalone that has the dependencies in `node_modules` (so things like the source-map library and websocket) included in the generated standalone.
  `--html-file`: Only valid if you've already specified `bundle-dependencies`. This tells the compiler to produce a "template" html file that runs the pyret program that was compiled. Here's an example use:

```
node -max-old-space-size=8192 build/phaseA/pyret.jarr --outfile examples/ahoy-world.jarr \
                      --build-runnable examples/ahoy-world.arr \
                      --builtin-js-dir src/js/trove/ \
                      --builtin-arr-dir src/arr/trove/ \
                      --compiled-dir compiled/ \
                      --require-config src/scripts/standalone-configA.json \
                      -bundle-dependencies \
                      --html-file examples/ahoy-world.html
```

This will produce `ahoy-world.jarr` and an `ahoy-world.html` that includes `ahoy-world.jarr` in a `<script>` tag.

**How `bundle-dependencies` works:**
As mentioned, the goal of the `bundle-dependencies` flag is to include all of the dependencies in `node_modules` as part of the `.jarr` file. The `Makefile` has a new target called `BUNDLED_DEPS` for producing a file that has all (I think) of the `node_modules` in bundled together (and properly defined so that our module loader can use them just as if they were in separate files). When we want to produce a `.jarr` file with the dependencies bundled we just dump this file in front of the Pyret program (see `make-standalone`).

The un-ideal thing about this is that the generated `.jarr` file might contain modules that aren't even dependencies of the Pyret program. Fixing this will be annoying for the two following reasons:
1) We'd need a way to statically determine all the JS modules a Pyret program depends on (maybe by parsing `define()` calls?). This seems doable.
2) Browserify, the tool I use to generated the `BUNDLED_DEPS` file, does not seem like it's able to `browserify` a program that uses Browserify. So if the compiler directly includes Browserify, and it tries to compile the compiler, we'll have some issues. I haven't had time to look into fixing this.

src/js/trove/make-standalone.js
```diff
@@ -1,8 +1,30 @@
 ({
   requires: [],
-  nativeRequires: ["path", "fs", "requirejs"],
+  nativeRequires: ["path", "fs"],
   provides: {},
   theModule: function(runtime, namespace, uri, path, fs, requirejs) {
+    var NODE_BUNDLED_DEPS_FILE = "build/bundled-node-deps.js";
```
puppyofkosh: We could make some of these part of the config rather than hardcoding them here in the future

src/js/trove/cmdline-lib.js
```diff
@@ -5,8 +5,13 @@
   theModule: function(runtime, _, uri) {
     return runtime.makeModuleReturn(
       {
-        "command-line-arguments": runtime.makeFunction(function() {
-          return runtime.ffi.makeList(runtime.getParam("command-line-arguments").map(runtime.makeString));
+          "command-line-arguments": runtime.makeFunction(function() {
+              var args = runtime.getParam("command-line-arguments");
+              if (args.length == 0) {
```
puppyofkosh: `cmdline.arr` depends on there being at least one CLI argument to get the file name

src/js/trove/require-node-dependencies.js
```diff
@@ -0,0 +1,34 @@
+//
+// When building a web-standalone, browserify will parse this file
+// and produce a version which include each dependency that is required()
+//
```
puppyofkosh: Having to maintain this file will be annoying, but I can't think of another way without using browserify in the compiler (see my earlier comment on this PR for why that's problematic)

src/scripts/web-standalone-template.html
```diff
@@ -0,0 +1,25 @@
...
+          }
+          }
+
+        </script>
+      </head>
+
+    <body>
+        <h1>My App</h1>
+    </body>
+    <script type="text/javascript" src="{{{compiled-jarr-file}}}"></script>
```
puppyofkosh: In `make-standalone`, we replace `{{{compiled-jarr-file}}}` with the actual name of the file.

src/js/base/secure-loader.js
```diff
@@ -1,4 +1,4 @@
-define("pyret-base/js/secure-loader", ["require", "q", "pyret-base/js/runtime-util"], function(rjs, Q, util) {
+define("pyret-base/js/secure-loader", ["q", "pyret-base/js/runtime-util"], function(Q, util) {
```
puppyofkosh: `rjs` is never used.

src/js/base/runtime.js
```diff
@@ -1230,7 +1224,6 @@ function (Namespace, jsnums, codePoint, util, exnStackParser, loader, seedrandom
     ************************/
     function checkType(val, test, typeName) {
       if(!test(val)) {
-        debugger;
```
puppyofkosh: I removed these `debugger;` statements since they make running the test-suite in the browser impossible without clicking the "resume" button a whole lot.

blerner: Hmmm, we like those debugger statements; they're sometimes the only way we can convince the browser to stop when we want it to...

jpolitz:
Note – the build failure was really random, and re-running it succeeded (https://travis-ci.org/brownplt/code.pyret.org/builds/245043567?utm_source=github_status)  Bypassing the Travis failure.

puppyofkosh:
It looks like that travis build didn't run the tests

src/js/base/runtime.js
puppyofkosh: I'll put them back in

PR #1044, opened 2017-06-20
Calling a method on by itself error message fix

puppyofkosh:
The program:

```
m = method(x): x end
m()
```

Should now produce a reasonable Pyret error message (rather than giving a JS stacktrace).

### DIFF
--- a/tests/pyret/tests/test-errors.arr
+++ b/tests/pyret/tests/test-errors.arr
@@ -125,6 +125,14 @@ check:
     e-non-ctor-app2.loc satisfies S.is-srcloc
   end
 
+  some-method = method(x): x end
+  e-method-app = get-err(lam(): some-method() end)
+  e-method-app satisfies E.is-non-function-app
+  when E.is-non-function-app(e-method-app) block:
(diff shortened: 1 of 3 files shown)


blerner:
Shucks, you missed the sourcemaps merge by like 1 hour...  :-)  Can you rebase onto horizon directly?  Some of the changes to runtime.js have already been made.

tests/pyret/tests/test-errors.arr
```diff
@@ -125,6 +125,14 @@ check:
     e-non-ctor-app2.loc satisfies S.is-srcloc
   end
 
+  some-method = method(x): x end
+  e-method-app = get-err(lam(): some-method() end)
+  e-method-app satisfies E.is-non-function-app
+  when E.is-non-function-app(e-method-app) block:
+    tostring(e-method-app.non-fun-val) is "<function>"
```
blerner: This is just annoying :(  "It's a non-function application.  The non-function thing was <function>."  Wha?  I know you can't do better, because dot-access automatically curries the method over its receiver, but this is still annoying...

Issue #1046, opened 2017-06-20
Expressions like ~10/2 cause internal error in  repl.


fedel:
Expressions like ~10/2 cause internal error in  repl.

I attach a screenshot:

![pyret-bug](https://user-images.githubusercontent.com/509238/27356687-d6aeffa6-55e5-11e7-94de-d86eb8b95d27.png)

blerner:
This isn't your errors' fault, @jswrenn  :)  It's a bug in js-numbers.  @ds26gte -- you have an optimization in Rational.makeInstance, such that when the result is an exact integer, you return the exact integer.  Then, that integer doesn't have a toRoughnum method, so things blow up...

jswrenn:
Thanks for reporting! This is an interesting bug.

This is the result of an internal javascript error (`"Rational.makeInstance(...).toRoughnum is not a function"`) leaking into pyret-land, being misinterpreted as a parse error, and going down the wrong code path. Consequently, [`result.exn.exn`](https://github.com/brownplt/code.pyret.org/blob/master/src/web/js/repl-ui.js#L103) is `undefined`, and attempted to render `undefined` as an error message completely chokes.

fedel:
blerner, jswrenn, thanks to you.

Also ~10/0 throw me an error.

PR #1047, opened 2017-06-21
fromString(): ~p/q where p/q is an integer should not error.

ds26gte:
Fix for #1046.


blerner:
Joe and I noticed yesterday that the original commit that added ~p/q syntax didn't add any tests for it at all. We didn't have time yesterday to audit whether other number syntaxes were also missing tests, but... Could you add some tests for this syntax, to make sure it (and any other untested number syntaxes) actually do behave correctly and don't have lingering js errors?

blerner:
Also especially note `~1/0` as a problem: `1 / 0` gives a runtime error, `1/0` gives a well formedness error (I think; not sure exactly which error it is), but `~1/0` I think slips past all your checks so far...

blerner:
@ds26gte ping -- this PR still has no tests, and I'm not sure that `~1/0` is handled properly by this yet.

jpolitz:
Another ping about tests.

ds26gte:
Have updated PR to catch `p/0+` and `~p/0+` (with optional sign) as invalid tokens. Tests in `parse.js` as these errors are caught at parse time.

ds26gte:
Don't understand the "conflicts" message here. It seems to be exactly and entirely the change I propose to `parse.js`. Is it because the file wasn't modified in the initial PR? I don't want to click `Resolve conflicts` in case it aggressively merges the file into `horizon` before you review it, which is what the infobubble suggests it will do...

blerner:
These should not be *parse* errors -- I want as few wonkery things in the `bad-expr` production as possible, and since tokenizing number is already getting quite fraught, I do not want to touch this.  Zero denominators are handled in https://github.com/brownplt/pyret-lang/blob/master/src/arr/compiler/well-formed.arr#L764-L769.  All you need to do is make ~p/q create a new `ROUGHRATIONAL` token in pyret-tokenizer, add that token to `frac-expr` in pyret-grammar, and then everything should just flow naturally.

ds26gte:
@blerner, I'll defer to you but...

I'd like to record here (in case we revisit this later) that catching as many errors as possible at the earliest lexing stage is a Good Thing in general and for Pyret in particular, given wishes expressed elsewhere about making Pyret snappier. You're in effect asking me to defer the error recognition till after the AST is created (I wasn't too happy about the late recognition of the non-roughnum 1/0 either), which means we're increasing time-to-compile (for all programs) by a non-finite amount that won't be compensated by having the lexing regexp for bad numbers be slightly less complicated.

Sorry for continuing the discussion, but it feels important to put this concern out there, and I'll of course do it your way.

blerner:
True -- but the counterargument is our parse errors suck compared to our well-formedness errors, and there's nothing we can do about that short of rewriting the parser completely.  "Lexing" and "parsing" are not concepts that make any sense to our students -- just, "it works" and "it yelled at me".  Moreover, well-formedness errors allow us to report *multiple* mistakes at once, rather than one at a time (because the first parse error is fatal to the parse).

blerner:
Closing as superseded by #1104

PR #1053, opened 2017-06-26
Updated PR for building standalone files with no npm/requirejs dependencies

jpolitz:
This is a cleaned up version of the original PR

https://github.com/brownplt/pyret-lang/pull/1030

I botched that by accidentally merging it; this version is cleanly merged on top of horizon with some additional fixes.


jpolitz:
Editorial remark – the big TODO for me here is to integrate this with CPO and make sure all is shipshape.  The potential useful thing for CPO is removing the dependency on requirejs, while still supporting AMD-style module definitions, _and_ supporting any `npm` modules that work with `browserify`.  This makes the build faster and less janky w.r.t. loading things like the SHA and seeded-random-number libraries via handwritten `<script>` tags.

The longer-term nice effect of this is the ability to build standalone web pages that start running arbitrary Pyret programs.  This works well for simple cases now, which aside from the obvious application of writing web pages in Pyret, also gives us a path for browser testing of Pyret that doesn't require starting whole CPO instances.

src/arr/compiler/cli-module-loader.arr
```diff
@@ -77,12 +77,12 @@ type Loadable = CL.Loadable
 
 type Either = E.Either
 
-fun uri-to-path(uri):
-  crypto.sha256(uri)
+fun uri-to-path(uri, name):
+  crypto.sha256(uri) + "-" + name
```
jpolitz: This should be changed to have name first!

src/js/base/amd_loader.js
```diff
@@ -0,0 +1,149 @@
...
+        console.log("Evaluating " + moduleObj.name);
+
+        if (moduleObj.val != null) {
+            throw new Error("Already evaluated " + moduleObj.name);
+        }
+
+        var callbackArgs = buildCallbackArgs(moduleObj.deps);
+
+        // Now call it
+        moduleObj.val = moduleObj.callback.apply(null, callbackArgs);
```
jpolitz: NOTE (add comment): This should call out explicitly that no stack management is needed, because this happens long before any Pyret code runs – this is for pure JS dependencies.

src/js/base/secure-loader.js
```diff
@@ -1,10 +1,8 @@
-define("pyret-base/js/secure-loader", ["require", "q", "pyret-base/js/runtime-util"], function(rjs, Q, util) {
+define("pyret-base/js/secure-loader", ["q", "pyret-base/js/runtime-util"], function(Q, util) {
   function compileInNewScriptContext(src) {
     var promise = Q.defer();
-    var loader_callback_count = 0;
     if(util.isBrowser()) {
-      loader_callback_count += 1;
-      var callback = "loader_callback_" + loader_callback_count;
+      var callback = "loader_callback_" + Math.floor((Math.random() * 1000));
```
jpolitz: This should be the counter version – there's no reason to do this randomly when they are being appended/removed, and sequential numbers are better gensyms.

src/js/trove/cmdline-lib.js
```diff
@@ -5,8 +5,13 @@
   theModule: function(runtime, _, uri) {
     return runtime.makeModuleReturn(
       {
-        "command-line-arguments": runtime.makeFunction(function() {
-          return runtime.ffi.makeList(runtime.getParam("command-line-arguments").map(runtime.makeString));
+          "command-line-arguments": runtime.makeFunction(function() {
+              var args = runtime.getParam("command-line-arguments");
+              if (args.length == 0) {
+                  args = ["<browser>"];
```
jpolitz: @puppyofkosh clarification q – is this saying “the executable for this run is the browser, not node”?

src/js/trove/make-standalone.js
```diff
@@ -1,8 +1,30 @@
 ({
   requires: [],
-  nativeRequires: ["path", "fs", "requirejs"],
+  nativeRequires: ["path", "fs"],
   provides: {},
   theModule: function(runtime, namespace, uri, path, fs, requirejs) {
```
jpolitz: Should remove requirejs from this list of arguments since it's no longer required above.

src/js/trove/make-standalone.js
```diff
@@ -41,68 +67,53 @@
...
-            });
-          }, function(err) {
-            console.error("Error while using requirejs optimizer: ", err); 
-            restarter.error(runtime.ffi.makeMessageException("Error while using requirejs optimizer: ", String(err)));
-          });
-        });
+      var outFile = fs.openSync(realOut, "w");
+
+      // Write the amd loader first
+      var loaderContents = fs.readFileSync(AMD_LOADER, READ_OPTIONS);
```
jpolitz: This should probably be (a) configurable and (b) tested for relative-path-ness depending on where the compiler was run from.

src/js/trove/make-standalone.js
```diff
@@ -41,68 +67,53 @@
...
+      fs.writeSync(outFile, loaderContents);
+
+      // Now either write the file containing all dependencies or the file which
+      // just defines() the dependencies.
+
+      var depsFile;
+      if (runtime.isPyretTrue(bundleDependencies)) {
```
jpolitz: This is essentially choosing statically-linked (if true) and dynamically linked (if false) npm packages.

src/scripts/web-standalone-template.html
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html>
+<html>
+    <head>
+      <head>
+        <script>
```
jpolitz: Let's indent this script tag correctly.

src/js/trove/cmdline-lib.js
puppyofkosh: Yes!

I made this change because in `trove/cmdline.arr` there is the following code:
```
all-cmdline-params = CL.command-line-arguments()
file-name = all-cmdline-params.first
```

This assumes that the list returned by `command-line-arguments()` isn't empty (which seemed reasonable to me) so I changed the function to always return a non-empty list.

PR #1054, opened 2017-06-26
Cleaned up fast mode PR

jpolitz:
I resolved the conflicts in 

https://github.com/brownplt/pyret-lang/pull/1020

And made a fix for missed `isFunction` checks when flatness is set to infinity.  There's an outstanding issue in how source-mapped stack traces work on stack traces that come from runs with multiple calls to `run`.


jpolitz:
Some overall comments from a review with @blerner today:

## When is the program done?

This PR sneaks by some pretty interesting issues with runtime management – even just the simple issue of knowing when a program is “done.” This is clearly an issue because the PR removes calls to `process.exit` in favor of setting the exit code and letting it finish when `node` decides it is done. This is because in asynchronous cases, Pyret is now registering callbacks without clearing its stack and controlling the callbacks itself. This means that the program is complete only when node's event queue is empty and the last Pyret callback has been run. This is in contrast to any past uses of asynchrony, which would _pause_ the Pyret stack and _resume_ it after the asynchronous work happened.

In order to do this correctly, the Pyret runtime needs some kind of state that tracks if there are outstanding Pyret callbacks – it shouldn't finish and, say, render the check results until there's no Pyret code left. One way to do this would be to have a runtime function that wraps all Pyret callbacks that are to be used in conjunction with underlying async `node` libraries, and updates state appropriately when those functions finish. Then, when each Pyret callback finishes, the wrapper can decide whether to trigger the finish-and-cleanup of the entire Pyret program, including reporting tests and setting error codes.

## Changes for library authors

In fast mode, some libraries are fundamentally unwritable. For example, it's not possible to write a fast mode library that pretends to by synchronous, as `image-url` does, or as the `gdrive-locators` library does. These libraries would have to take user-supplied Pyret callbacks, wrap them as above, and force the user to think about the asynchrony.

This also means that a library author needs to understand, consider, and document the interaction of the library with fast mode. A library that works in fast mode _can_ work in regular mode, but a library that requires e.g. pauseStack's true asynchrony _cannot_ work in fast mode. This might require metadata on packages that indicate in which mode the compiler should run, for instance.

A general atmosphere of “use at your own risk” should rightfully surround the `--fast-mode` flag, as a result.

## No changes to the compiler?

It turns out that with the standalone compilation PR, the compiler actually can be fully synchronous, and never actually need to register a callback with node (the only one needed in this PR is for an async call to `requirejs`, which the standalone PR removes as a dependency). This is only true for offline compilation, where sensible synchronous file access exists (it would be a non-starter to, say, load all of the GDrive files synchronously, which does require that the compiler be async in CPO).

We could accept and enforce the synchronous constraint on the compiler for offline code. This would get us the performance benefits described in #1020 in exchange for restricting the future of offline compiler development.

## Changes for students

The biggest question is if this is worth it to expose to end-users and students. Students working on large datasets could certainly benefit from a performance boost. Their programs (and their instructors), would be constrained by providing fast-mode-compatible libraries, which might mean teaching callbacks and/or carefully choosing APIs to make it possible to do things synchronously. This might limit the ability to interact with web-facing APIs, for instance, instead preferring the easier-to-use synchronous file system APIs.

If these libraries make use of offline resources, there's perhaps no effort wasted since they would never work in the browser. But sharing code that works over, say, http might be a lot harder offline and online depending on the synchronous support available offline to make the APIs behave the same, and there will be duplicated effort in working with differing backends for libraries.

jpolitz:
Closing for the reasons in the long review above. I think the v-hull branch provides a better way forward than this effort at this point.

PR #1055, opened 2017-06-26
Lift case dispatch object to the top level

sorawee:
For example:

```
data Test:
  | test-a(x)
  | test-b(x)
end

cases (Test) test-a(123):
  | test-a(x) => x + 5
  | test-b(x) => x - 5
end ^ print
```

Prior this PR, this compiles to:

```
var $toplevel28 =
  function($$resumer34) {
    var $step27 = 0;
    ...
    while(!R.isContinuation($ans30)) {
      switch($step27) {
      case 0:
        ...
      case 1:
        var $cases_dispatch76 = { "test-a": 2, "test-b": 3 };
        $al31 = L[9];
        $step27 = $cases_dispatch76[cases62.$name] || 4;
        break;
      case 2:
        ...
      case 3:
        ...
      case 4:
        ...
      case 5:
        ...
      case 6:
        ...
      case 7:
        ...
      case 8:
        ...
      case 9:
      ...
    case 10:
      ...
    default: R.ffi.throwSpinnakerError(L[13], $step27);
    }
  }
  ...
  return $ans30;
};
```

With this PR, it compiles to:

```
var $cases_dispatch76 = { "test-a": 2, "test-b": 3 };
var $toplevel28 =
  function($$resumer34) {
    var $step27 = 0;
    ...
    while(!R.isContinuation($ans30)) {
      switch($step27) {
      case 0:
        ...
      case 1:
        $al31 = L[9];
        $step27 = $cases_dispatch76[cases62.$name] || 4;
        break;
      case 2:
        ...
      case 3:
        ...
      case 4:
        ...
      case 5:
        ...
      case 6:
        ...
      case 7:
        ...
      case 8:
        ...
      case 9:
      ...
    case 10:
      ...
    default: R.ffi.throwSpinnakerError(L[13], $step27);
    }
  }
  ...
  return $ans30;
};
```

The sole purpose of doing this is to avoid case dispatch object allocation,
especially in recursive functions.

This transformation preserves the semantics of the program because the case
dispatch objects do not refer to any other variable, and all identifiers
associated with case dispatch objects are fresh (so it's impossible to
get accidentally captured).

### DIFF
--- a/src/arr/compiler/anf-loop-compiler.arr
+++ b/src/arr/compiler/anf-loop-compiler.arr
@@ -101,16 +101,6 @@ j-for = J.j-for
 j-raw-code = J.j-raw-code
 make-label-sequence = J.make-label-sequence
 
-# cases-dispatches stores bindings of case dispatch objects
-# this is so that the objects can be allocated once in the top level, avoiding
-# multiple allocations which could affect performance, particularly in recursive
-# functions.
-# this object is initialized here to be empty (concat) list, and get mutated
-# to accumulate bindings. When inserting the bindings to the top level, this
-# variable gets reset to empty list again so that it can be subsequently used
-# in the future.
-var cases-dispatches = cl-empty
-
 fun console-log(lst :: CL.ConcatList) -> J.JStmt:
   j-expr(j-app(j-id(A.s-name(A.dummy-loc, "console.log")), lst))
 end
@@ -120,6 +110,14 @@ is-t-data = T.is-t-data
 data BindType:
   | b-let(value :: N.ABind)
   | b-array(value :: N.ABind, idx :: Number)
+end
+
+# this structure stores bindings of case dispatch objects
+# so that the objects can be allocated only once in the top level, avoiding
+# multiple allocations which could affect performance, particularly in recursive
+# functions.
+data Dispatches:
+  | dispatches-box(ref dispatches :: CList<J.JStmt>)
 end
 
 js-names = A.MakeName(0)
@@ -549,7 +547,7 @@ fun compile-fun-body(l :: Loc, step :: A.Name, fun-name :: A.Name, compiler, arg
   main-body-cases.each(lam(c): when J.is-j-case(c): c.exp.label.get() end end)
 #  compiler.add-phase("Compile anns: " + l.format(true), nothing)
   start = time-now()
-  main-body-cases-and-dead-vars = DAG.simplify(compiler.add-phase, main-body-cases, step, cases-dispatches)
+  main-body-cases-and-dead-vars = DAG.simplify(compiler.add-phase, main-body-cases, step, compiler.dispatches!dispatches)
   finish = time-now() - start
 #  compiler.add-phase("Simplify body: " + l.format(true), nothing)
   total-time := total-time + finish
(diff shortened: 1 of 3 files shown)


src/arr/compiler/js-dag-utils.arr
```diff
@@ -339,25 +339,16 @@ fun stmts-of(blk :: J.JBlock):
   end
 end
 
-fun find-steps-to(stmts :: ConcatList<J.JStmt>, step :: A.Name, acc :: D.MutableStringDict<J.Label>) -> D.MutableStringDict<J.Label>:
+fun find-steps-to(stmts :: ConcatList<J.JStmt>, step :: A.Name, acc :: D.MutableStringDict<J.Label>, cases-dispatches :: ConcatList<J.JStmt>) -> D.MutableStringDict<J.Label>:
   var looking-for = none
```
blerner: I think you can nuke this variable now, too, right?  If so, hooray!  It was always an ugly hack...

src/arr/compiler/anf-loop-compiler.arr
```diff
@@ -101,6 +101,8 @@ j-for = J.j-for
 j-raw-code = J.j-raw-code
 make-label-sequence = J.make-label-sequence
 
+var cases-dispatches = cl-empty
```
blerner: Would be good to have a comment here explaining the purpose of this variable, and in particular its lifetime: when does it start off as cl-empty, when does it grow, when does it get read, and when does it get reset back to cl-empty?  (I don't see it go back to empty, and that worries me a bit, if we try to compile multiple files in a single run of pyret.jarr: will the cases-dispatches from file1.arr get declared in file2.arr, as a result?

src/arr/compiler/js-dag-utils.arr
```diff
@@ -370,8 +361,12 @@ fun find-steps-to(stmts :: ConcatList<J.JStmt>, step :: A.Name, acc :: D.Mutable
           else if J.is-j-binop(expr.rhs) and (expr.rhs.op == J.j-or):
             # $step gets a cases dispatch
             # ASSUMES that the dispatch table is assigned two statements before this one
```
blerner: Nuke this comment too, since it's outdated.

src/arr/compiler/js-dag-utils.arr
sorawee: Ah, yes, I forgot to remove it (in the original PR I did).

src/arr/compiler/anf-loop-compiler.arr
sorawee: A totally handwaving answer (because I am not sure of the actual answer):

1. Below this line, there are `js-ids = D.make-mutable-string-dict()` and `effective-ids = D.make-mutable-string-dict()`. If this doesn't get reset, those lines wouldn't get reset as well, and we would have had problem because of that long time already.
2. Experimentally, when I compiled `test.arr` to `test.js`, other libs get bundled too and somehow it gets reset.

blerner: Not really, because we never iterate over those dictionaries and output everything they contain (the way you have to, with including cases-dispatch in the compiled output).

When you compile test.arr, a single file, that doesn't necessarily mean you needed to recompile anything else.  Try compiling phaseB, and see if things get corrupted.

sorawee: You are absolutely right :(

jpolitz: Could you make it be a box on the compiler object?  We don't really support raw `ref` fields on the compiler object itself, but you could make a small data instance:

```
data Dispatches:
  | dispatches-box(ref dispatches)
end
```

and then construct a new one and put it aside the other accumulators on the compiler like get-loc and current-ans.  It would still be stateful but it would be instantiated per-call to compile-module

sorawee: So this value must be accessible from `compile-fun-body`, `compile-split-cases`, and `compile-module`. If we were to initialize it in `compile-module`, we would need to juggling this value everywhere. Having it as the global variable gets rid of this problem.

And it's easy to reset even in the global variable version. Right after we output the bindings, reset it to empty list.

Of course, I'm not saying that it's the best practice. If you think that it's objectively better to instantiate in `compile-module`, I will do that.

jpolitz:
Note that we pass the value `compiler` around variously as `self` in
visitors and as `compiler` in helpers.  The state is already threaded
through everywhere, so it should be accessible in all these places if you
make it a field of the compiler visitor object.

sorawee:
Done!

blerner:
I...think this is fine.  Worth merging.

jpolitz:
@blerner Can you click "approve" in your review?

PR #1071, opened 2017-07-15
Allow roughnums to have optional suffix of ellipsis in input.

ds26gte:
In output always print ellipses.
Generate bad-number error if ellipsis-like dots follow non-roughnums, or
if too few or too many dots follow roughnums.

This is fresh PR for issue #664 and superseded PR #1029, including all the suggestions therein, and built against current `horizon`.


blerner:
Really, really want to see tests for these...  The best place for tests would be in tests/parse/parse.js, around lines 209--218.  You can use the `lex` and `parse` functions for this; I suspect you'll want to say `lex("some bad number").not.toBe(false) && parse("some bad number").toBe(false)`, for various bad numbers.

src/js/base/pyret-tokenizer.js
```diff
@@ -119,9 +119,19 @@ define("pyret-base/js/pyret-tokenizer", ["jglr/jglr"], function(E) {
 
   const number = new RegExp("^[-+]?" + unsigned_decimal_part, STICKY_REGEXP);
 
-  const badNumber = new RegExp("^~?[+-]?\\.[0-9]+(?:[eE][-+]?[0-9]+)?", STICKY_REGEXP);
+  const unsignedRationalorDecimal = "(?:" + unsigned_rational_part + "|" + unsigned_decimal_part + ")";
 
-  const roughnum = new RegExp("^~[-+]?"  + "(?:" + unsigned_rational_part + "|" + unsigned_decimal_part + ")", STICKY_REGEXP);
+  const numberwithoutIntegerPart = "^~?[+-]?\\.[0-9]+(?:[eE][-+]?[0-9]+)?";
+
+  const nonRoughnumwithDots = "^[+-]?" + unsignedRationalorDecimal + "\\.\\.+";
```
blerner: Is this really supposed to be "two or more trailing dots"?

ds26gte: Yes. I tried making it one or more, with a negative lookahead to make sure it isn't followed by a digit, but there is a test `5.{1}` that is supposed to elicit a different error.

ds26gte:
Have added tests that for bad non-roughnums with trailing dots and roughnums with wrong number of trailing dots.

tests/pyret/tests/test-parse-errors.arr
```diff
@@ -74,7 +74,36 @@ end
 bad-number = [list:
   ".0",
   "~.0",
-  "foo.0"
+  "foo.0",
+  #"2.",
```
blerner: Why is this (and `~2.` below) commented out?

ds26gte: Those are already caught as a "program ended too soon" rather than as a bad-number parse error.  Devising a sufficient lookahead to trap them as bad numbers would involve a very complicated regexp, and it may not be worth the inefficiency given the errors are already caught via another route.

blerner: I see.  (This is why I'd recommended these be done as JS lexer tests, rather than Pyret test cases... It'd be more obvious to see which ones don't lex, and which ones lexed as two tokens.)

ds26gte: Added related lexing tests to `tests/parse/parse.js`.

jpolitz:
I have two comments:

- I am mostly convinced that we should say that a number followed by any number of dots other than 3 is a lexing error, which makes the `5.` in `5.foo` be a “bad number” error rather than parsing and producing a “field-not-found” error. @blerner's remarks on it simplifying the lexing are convincing.
- Upon reflection, I think that we should use the `5.3...e+10` printed representation, and never put or allow the dots after the exponent.

shriram:
Have we agreed that we cannot have roughnums be only … at the end with no ~
at the beginning? That would get rid of the "is it - or is it ~" confusion
and free up =~ for equals-now, etc.

blerner:
The tilde prefix is not optional: having a trailing ellipsis is optional. I don't see how this helps at all in reframing the meaning of =~ though -- that seems like a non sequitur.

shriram:
Because we've just talked about how ~ is used for both roughness and time.
If ~ no longer tied to roughness, then =~ could continue to be about time,
and can't be mistaken for "roughly-equals".

blerner:
That doesn't solve the problem: we're lacking an equality operator meaning is-roughly-equal. Making =~ even less relevant doesn't help that...

jpolitz:
This seems to have stalled on scientific notation, which seems like a bummer.

shriram:
I'm sure everyone else has already considered and rejected this, but let me ask the dumb question: do we need scientific notation? Has anyone here actually used it? @schanzer Would it kill us in the field?

shriram:
@blerner in particular can comment on physics.

blerner:
Commenting in the underlying #664 issue

ds26gte:
An updated PR, #1197, is now submitted. It puts the optional ellipsis before the exponent for roughnums using scientific notation. Also: tons of tests to check that parsing indeed fails on illegal numbers that put dots where they shouldn't be, or too many or too few of them when they are required.

schanzer:
If #1197 subsumes it, can this PR be closed?

PR #1095, opened 2017-07-21
linear-regression: minimize overflow errors

ds26gte:
by coercing y's members into roughnums before adding them.


shriram:
Merged but not fixed!

PR #1098, opened 2017-07-21
linear-regression, r-squared: coerce y arg to list of roughnums from get-go.

ds26gte:
Fix to previous PR.

### DIFF
--- a/src/arr/trove/statistics.arr
+++ b/src/arr/trove/statistics.arr
@@ -157,14 +157,15 @@ fun linear-regression(x :: List<Number>, y :: List<Number>) -> (Number -> Number
   else if x.length() < 2:
     raise(E.message-exception("linear-regression: input lists must have at least 2 elements each"))
   else:
-    xpt-xy = math.sum(map2(lam(xi, yi): xi * yi end, x, y))
-    xpt-x-xpt-y = (math.sum(x) * math.sum(map(num-to-roughnum, y))) / x.length()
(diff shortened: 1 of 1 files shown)


blerner:
For those of us without immediate context: what overflow is this trying to solve?

ds26gte:
Shriram's client had a program that was calling `linear-regression` and `r-squared` with an argument that was an 186-element list of rational numbers. Doing arithmetic (various summing operations) on this list created an intermediate rational number p/q with very large p and q, which created an overflow error when p/q was converted to roughnum by individually converting p and q into fixnums. While it would be great to have p/q convert to roughnum correctly (which we should), for now, the problem is solved by coercing the large list of rationals into a list of roughnums before doing any arithmetic on it.

jpolitz:
I'm closing this because a355c7c7db23c2be667be5c188065d431b6124db solved the problem; either would work but I think the `shadow` is clearer.  Also, this PR doesn't make the `x` list be rough, which could cause other issues.

This is an interesting case overall – I think a lot of statistical functions are going to need to coerce to rough directly to avoid similar issues.

ds26gte:
Sounds good, although I'm still holding out for a solution where this kind of strategic pre-coercion of list members is unnecessary by simply making `Rational.toRoughnum` work correctly even for huge p and q. `jsnums` already has a `toRepeatingDecimal` that is used by the repl's output toggle, and it should be rejiggerable for this purpose. Soon...

Issue #1102, opened 2017-07-25
big integers and rationals don't convert to converging fixnum or roughnum


ds26gte:
By big rationals, this is especially egregious for p/q where p and q are large, but p/q itself is quite handy, as in the following case:

```
var p = 
81268721875292322346006428361277486504132538854011222666475817166249350496626931570388521644403314190415953881849740041567819628199876803591646157379114065524563985446908892262196425992509944671317287759517001454359426892629201797719306000630092659618487347668341846237234687585768929352898628996784695074874292706491575251948097349309856501057270192988393033514058115

var q = 31074826015292464136352850911368809028150373077645694092712513113641716099394624728813926091948046541979298815056591584707047923844529494021740153955742689843985239712176553523899854808158284912207177902877491879709295462823402583883946947080040968647549604073979681744304215760537682790820012842244687066165442575364044873225935129869506345618953111615635565744

var z = num-to-roughnum(p / q)

var w = num-to-roughnum(p)
```

ds26gte:
Now that #1104 uses a different token (`RATIONAL` rather than `NUMBER`) to capture rough rationals, literal `~p/q` don't take advantage of the #1103 ability to converge in cases where

```
num-to-roughnum(p)     --> diverges, AND/OR
num-to-roughnum(q)     --> diverges, BUT
num-to-roughnum(p / q) --> converges
```

Fixable by having rough rationals generate an `s-rfrac` AST phrase different from `s-frac`, and treating them differently when desugaring.

PR #1103, opened 2017-07-25
ensure integer/rational coercion to roughnum if at all feasible

ds26gte:
This PR more durably solves the problem addressed by a355c7c and discussed in #1098, #1102.

Problem statement: Doing a chain of rational arithmetic on a bunch of rational numbers can result in a substantial but not necessarily large rational number, which causes overflow on coercion to roughnum. By "substantial but not large rational", I mean a number `p/q` where `p` and/or `q` are large ints that would individually overflow on coercion to roughnum, but their quotient `p/q` is small enough to not overflow.

Solution: (a) There was a bug in the `toFixnum` method for `BigInteger` which wasn't allowing coercion for BigIntegers that were quite within `[MIN_VALUE, MAX_VALUE]`. Fixed by introducing a subroutine `splitIntintoMantissaExpt` that, given an int (however large), finds a guaranteed fixnumable (i.e., small) mantissa and the corresponding exponent. The fixnum of the whole may yet fail if the exponent is too large, but that's a genuine overflow.

(b) The `toFixnum` method for `Rational` was separately coercing the numerator `m` and denominator `n` to fixnum, and then dividing, which causes needless overflows. Solved by rewriting `_integerDivideToFixnum` to use the same `splitIntintoMantissaExpt` to find the mantissae and exponents for `m` and `n`:

```
  m = a * 10^b
  n = c * 10^d
```

where `a`, `c` are guaranteed fixnums, and `b`, `d` are ints.

We now calculate

```
  (a/c) * 10^(b - d)
```

which will be a fixnum _if_ `b - d` is small enough -- even if `b` and/or `d` are too large.

These `toFixnum` methods are used by `toRoughnum`, which is either explicitly employed by the user (as `num-to-roughnum`) or is implicitly invoked by roughnum contagion.

A benefit from this PR is that we don't have to make precautionary explicit coercions of rationals to roughnum to forestall overflow. In particular, the case in `statistics.arr` -- where a sum of many rationals was rewritten to be a sum of the roughnums corresponding to the rationals -- can now be left in the original natural and uncluttered style.

Note that this solution will still cause a BigInteger or Rational to overflow on coercion if its value is outside `[MIN_VALUE, MAX_VALUE]`. This is as it should be: We of course cannot have a roughnum whose value is outside the range allowed to JS doubles.


tests/pyret/tests/test-numbers.arr
```diff
@@ -276,3 +276,19 @@ check "floor/ceiling/round of bigrats":
   num-ceiling(+123123123123123123123123124.5) is 123123123123123123123123125
   num-ceiling(-123123123123123123123123124.5) is -123123123123123123123123124
 end
+
+check "non-erroring roughnum coercion for rationals and bigints whenever feasible":
+  var p = 81268721875292322346006428361277486504132538854011222666475817166249350496626931570388521644403314190415953881849740041567819628199876803591646157379114065524563985446908892262196425992509944671317287759517001454359426892629201797719306000630092659618487347668341846237234687585768929352898628996784695074874292706491575251948097349309856501057270192988393033514058115
```
blerner: No need for `var` -- we are Pyret, not JS :)

src/js/base/js-numbers.js
```diff
@@ -1201,16 +1201,44 @@ define("pyret-base/js/js-numbers", function() {
...
+      str = str.substring(1);
+      if (c0 === '-') {
+        sign = '-';
+      }
+    }
+
+    var d0 = str[0];
+    str = str.substring(1);
+
+    var mantissa = Number(sign + d0 + '.' + str);
```
blerner: Could definitely do with a comment or two here; I've read this code three times now and I'm not quite sure whether it works or not...  Is the invariant here that `s` is a (big) integer, so there should be no decimal point in the `toString` output, that you are then inserting?  Is the expt off by one below?

blerner: Oh, it is an int, based on the name (which I misread as "split into mantissa expt") -- need another capital letter in there...

src/js/base/js-numbers.js
```diff
@@ -1201,16 +1201,44 @@ define("pyret-base/js/js-numbers", function() {
...
   // _integerDivideToFixnum: integer-pyretnum integer-pyretnum -> fixnum
   var _integerDivideToFixnum = makeIntegerBinop(
     function(m, n) {
       return m / n;
     },
     function(m, n) {
-      return toFixnum(m) / toFixnum(n);
+      var xm = splitIntintoMantissaExpt(m);
+      var xn = splitIntintoMantissaExpt(n);
+      var r = Number(String(xm[0] / xn[0]) + 'e' +
```
blerner: This math confuses me a bit, because it's not numerically the same thing as before.  `splitIntintoMantissaExpt` discards precision, doesn't it?  Does that potentially affect anything?

src/js/base/js-numbers.js
```diff
@@ -1201,12 +1201,26 @@ define("pyret-base/js/js-numbers", function() {
       return bnRemainder.call(m, n);
     });
 
-  var splitIntintoMantissaExpt = function(s) {
+  // splitIntIntoMantissaExpt: integer-pyretnum -> [JS-double, JS-int]
+  // 
+  // splitIntIntoMantissaExpt takes an integer s (either unboxed or BigInteger)
+  //   and returns [mantissa, exponent]
+  //   such that s ~= mantissa * 10^exponent.
+  // mantissa is a JS-double, and is chosen to be have one non-zero digit
```
blerner: typo "to be have"

src/js/base/js-numbers.js
```diff
@@ -1224,13 +1238,24 @@ define("pyret-base/js/js-numbers", function() {
   };
 
   // _integerDivideToFixnum: integer-pyretnum integer-pyretnum -> fixnum
+  //
+  // _integerDivideToFixnum takes two integers (possibly BigIntegers) and 
+  //   returns the best fixnum representing their quotient.
+  // If the args are both JS-doubles, the JS quotient is returned if it
+  //   doesn't overflow.
+  // If it does overflow, or if at least of the args is a BigInt, then
```
blerner: typo "at least of the args"

src/js/base/js-numbers.js
ds26gte: Yes, `splitIntIntoMantissaExpt` discards precision (`_integerDivideToFixnum` is all about losing precision in the least lossy way), but so too does `toFixnum`. Using two calls to `splitIntIntoMantissaExpt` instead of two calls to `toFixnum` at least gives a chance for the exponents to cancel out in cases where otherwise overflow would result.

src/js/base/js-numbers.js
ds26gte: I've added comments for both `splitIntIntoMantissaExpt` (was using English titlecase that keeps prepositions lowercase) and `_integerDivideToFixnum`. Yes, `split...` takes an arbitrarily precise int and returns a [mantissa, exponent] with generally some loss of precision, as both mantissa and exponent are constrained to be JS doubles. In addition, mantissa is in the interval [1, 10).

tests/pyret/tests/test-numbers.arr
ds26gte: Changed.

src/js/base/js-numbers.js
ds26gte: Thanks -- both typos fixed.

shriram:
Is this commit combined with a concomitant edit to the docs, as needed? It shouldn't be merged without a docs update.

ds26gte:
@shriram, do you mean user-facing docs? This change by intent does not change the user's API. All it does is make roughnum coercion (implicit or explicit) not diverge as often.

shriram:
Good, thanks. Then we're fine.

PR #1104, opened 2017-07-26
roughnum fractions recognized, except for 0 denominator

ds26gte:
This supersedes PR #1047 (replacing parsing error with well-formedness error). Fixes issue #1046.

- Well-formed "rough rationals", `~p/q` are recognized correctly, whether they reduce to integers, or proper or improper fractions. Test in `test-numbers.js`.

- Denominator zero in `~p/q` triggers a well-formedness-error. Test in `well-formed.js`.

### DIFF
--- a/src/js/base/js-numbers.js
+++ b/src/js/base/js-numbers.js
@@ -2034,7 +2034,8 @@ define("pyret-base/js/js-numbers", function() {
 
     aMatch = x.match(roughnumRatRegexp);
     if (aMatch) {
-      return Rational.makeInstance(fromString(aMatch[1]), fromString(aMatch[2])).toRoughnum();
+      return toRoughnum(Rational.makeInstance(fromString(aMatch[1]), fromString(aMatch[2])),
+        errbacks);
     }
 
     aMatch = x.match(roughnumDecRegexp);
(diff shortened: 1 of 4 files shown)


blerner:
This looks pretty good to me :)  I'm slightly surprised by the handling of the rough-rational as being "get the two parts, make a Rational, and then make it a roughnum": could you use your new fancy rough-integer-division code instead to avoid the intermediate rational?

Also -- should probably have a few more tests of negative rough-rationals, and explicitly-positive ones too, just to make sure the sign-handling works properly.

ds26gte:
Making an intermediate rational isn't a problem, because the fancy splitter is automaticaly invoked by `toFixnum` on rationals. 

Alas, that whole part is now deadc0de, because rough rationals go through a different token (`RATIONAL` rather than `NUMBER`) because of this very PR, and the numerator and denominator get individually numberized into a roughnum and bigint respectively, so we don't get the benefit.

But, clearly, the splitter is way too effective to not want to use for capturing a lot more `~p/q` literals, so I'll try to bring it into play. So don't merge this PR yet. :-)

ds26gte:
(a) Added more tests.

(b) On second thought, I'd merge this PR for now, and leave making literal `~p/q` use the mantissa-splitter to a subsequent PR. This PR solves an important issue already. To put things in perspective: the edge cases that a smart `~p/q` parser would save from divergence would have to have at least `p` or `q` be a 308-digit number. Such a number would have to be literally [sic] present in a program text, hence hardly a common scenario. (Rationals that arise in the course of computation coerce just fine, with #1103.)

(c) I have a path toward solving (b), but piggybacking it on this PR without #1103 as a baseline is unnecessary clutter.

blerner:
@jpolitz I'm ok with this PR; merge when you're satisfied too.

PR #1107, opened 2017-07-28
Typing getMaker

mkolosick:
This returns the desugaring to getMaker and fixes a bug in the type checker that prevented it from type checking.


blerner:
@jpolitz I've already done a preliminary code review on this, and it looks good.  Basic idea: split the prim-app variadic getMaker into getMaker0--getMaker5 and getMaker, give them individual type signatures (even though they share a common implementation in runtime.js), and then type checking basically Just Works.

src/js/base/runtime.js
```diff
@@ -467,6 +467,19 @@ function (Namespace, jsnums, codePoint, util, exnStackParser, loader, seedrandom
       return thisRuntime.getFieldLoc(obj, field, ["runtime"]);
     }
 
+    function getMaker(obj, makerField, exprLoc, constrLoc) {
+      var maker = isObject(obj) && obj.dict[makerField];
+      if (!isFunction(maker)) {
+        if (thisRuntime.ffi === undefined || thisRuntime.ffi.throwFieldNotFound === undefined) {
```
jpolitz: Note https://github.com/brownplt/pyret-lang/issues/1126 (doesn't need fixing for this PR, but notice that the pattern has been copied).

Also, the pattern has been copied incorrectly; I'm pretty sure this should say `throwConstructorSyntaxNonConstructor`, not `throwFieldNotFound`

PR #1115, opened 2017-07-31
[Emacs] Autoload pyret-mode on .arr files.

aatxe:
Added autoloading of pyret-mode to .arr files to the emacs mode. It's a small change, but is convenient.


blerner:
Thanks :)  While you're at it, you should also add
```
(add-to-list 'file-coding-system-alist '("\\.arr\\'" . utf-8-unix))
```
in there, to ensure files are Unicode-aware.

aatxe:
Done. :)

PR #1130, opened 2017-08-07
occurrence typing in cases

mkolosick:
Added occurrence typing in case branches.


blerner:
Well that's a nice, tiny change for a feature that neat 😀

sorawee:
Yes!!!

PR #1134, opened 2017-08-14
Contract improvements: weaving contract annotations onto relevant definitions

### LINKED ISSUE #1114: Proposal for weaving contracts onto functions
We've left this as a TODO item for ages, and it prompted a discussion about how contract lines are/aren't comments...  We've left this open because it's not immediately clear what it means to weave annotations in, so here's a concrete proposal.

1. *Where are annotations allowed?*  The challenge here is matching the identifiers shown in the contracts to the resolved `Name`s in definitions, especially since they appear in the wrong order, appear before name resolution occurs, and we have to handle something to do with shadowing.  I suggest we allow two positions: for any given file, contracts should appear _either_ consecutively in a bunch of contracts at the top of the file (just below any provide/import declarations), _or_ we allow contracts to appear immediately before the definition group in which their name appears.  (I have to quibble about "definition group" in order to accommodate let-rec'ed functions.)  This solves the name-resolution problem: if presented just before a definition group, then it's' "whatever `Name` comes next".  If presented at the top of the file, then it's "whatever the final binding of that `Name` is", exactly how `provide` works.

2. *What to do about imports?*  I think we should _wrap_ imported functions with a new lambda that does two things: it enforces the annotation as specified here, and intercepts any general-type-or-arity-mismatch errors from the underlying and complains about an annotation-mismatch (a new kind of error) between the purported contract and the underlying signature of the function.  In other words, if a student writes
```
circle :: Number -> Image
```
which is blatantly wrong, then this contract should desugar to (a pseudocode mishmash of Pyret and JS)
```
shadow circle = lam(n :: Number) -> Image:
  try {
    ans = circle(n)
    if (ans is-a Image) { return ans }
    else { raise contract-return-mismatch(ans, Image, srcloc of contract) }
  } catch(e) {
    raise contract-mismatch(e, srcloc of contract);
  }
end
```
The idea being, the underlying imported function is "trusted" as ground-truth, and the annotation here is more likely to be wrong.  The `contract-mismatch(e, srcloc)` idea is to be rendered as "The actual function expected an arity ### or arguments of type ###, but the contract here stipulated something else: double-check your contract."  The `contract-return-mismatch(ans, Image, srcloc)` idea is to handle the anonymous return value more flexibly.

3. *What to do about redundant contracts?* If a student writes
```
square-num :: Number -> Number
fun square-num(n :: Number) -> Number:
  n * n
end
```
should that be a well-formedness error because the parameter is explicitly annotated anyway?  What if the contract and the annotation disagree?  I'm inclined to say that specifying both a contract and an annotated function definition should be a well-fordmedness error, *or* to make it be a wrapped-lambda like the imported case above, with a preference for the well-formedness error.

Thoughts?

### LINKED ISSUE #1131: Enhancing contract syntax to include parameter names
Promoting an email thread to a feature request, and related to #1114 --- and I'd really like to come to consensus on this concurrently with that other issue, so that we change the behavior all at once and in a predictable way:

Currently, our syntax for contracts doesn't allow students or teachers to include the parameter names in the definition.  Concretely, we can write

```
num-exp :: Number, Number -> Number
circle :: Number, String, String -> Image
```
but we can't write "The first number to num-exp is a base and the second is the exponent" or "The first string is outline/solid and the second is a color"  Ironically, we don't even follow this in our documentation, and we write

```
num-exp :: (base :: Number, exp :: Number) -> Number
circle :: (radius :: Number, mode :: String, color :: String) -> Image
```
which is a nice syntax, but is incompatible with what we currently have implemented.  We have several potential syntaxes here, differing in the placement and necessity of parens:

1. `circle :: Number, String, String -> Image` as a top-level unparenthesized annotation
1.5. `circle :: (Number, String, String -> Image)` as a top-level parenthesized annotation
2. `circle :: (radius :: Number, mode :: String, color :: String -> Image)`
3. `circle :: (radius :: Number, mode :: String, color :: String) -> Image` as in our documentation
4. `circle :: radius :: Number, mode :: String, color :: String -> Image`

I'm not quite confident that (4) is unambiguously parseable (in the higher-order or refinement-annotation cases), so I'd really like to rule that out as not-worth-the-risk.  (1) and (1.5) are our current implementation, and so (2) is the closest to that...but (3) matches our docs and seems to be the most "natural" for people who've been writing them this week.

Thoughts?

blerner:
Fixes #1114 and fixes #1131

### DIFF
--- a/src/arr/compiler/resolve-scope.arr
+++ b/src/arr/compiler/resolve-scope.arr
@@ -218,12 +218,12 @@ fun weave-contracts(contracts, rev-binds) block:
                           acc)                        
                       end
                     else:
-                      errors := link(C.contract-non-function(c.l, id-name, l), errors)
+                      errors := link(C.contract-non-function(c.l, id-name, l, true), errors)
                       link(fun-to-lam(bind), acc)
                     end
                   | else =>
                     if A.is-a-arrow(c.ann) or A.is-a-arrow-argnames(c.ann) block:
-                      errors := link(C.contract-non-function(c.l, id-name, bind.value.l), errors)
+                      errors := link(C.contract-non-function(c.l, id-name, bind.value.l, false), errors)
                       link(bind, acc)
                     else:
                       link(rebuild-bind(bind, A.s-bind(l, shadows, id, c.ann), bind.value), acc)
@@ -675,7 +675,7 @@ where:
   id = lam(s): A.s-id(d, A.s-name(d, s)) end
   checks = A.s-app(d, A.s-dot(d, U.checkers(d), "results"), [list: ])
   str = A.s-str(d, _)
-  ds = lam(prog): desugar-scope(prog, C.standard-builtins).visit(A.dummy-loc-visitor) end
+  ds = lam(prog): desugar-scope(prog, C.standard-builtins).ast.visit(A.dummy-loc-visitor) end
   compare1 = A.s-program(d, A.s-provide-none(d), A.s-provide-types-none(d), [list: ],
         A.s-let-expr(d, [list:
             A.s-let-bind(d, b("x"), A.s-num(d, 10))
@@ -1419,7 +1419,7 @@ fun check-unbound-ids-bad-assignments(ast :: A.Program, resolved :: C.NameResolu
 where:
   p = PP.surface-parse(_, "test")
   px = p("x")
-  resolved = C.resolved(px, empty, [SD.mutable-string-dict:], [SD.mutable-string-dict:], [SD.mutable-string-dict:])
+  resolved = C.resolved-names(px, empty, [SD.mutable-string-dict:], [SD.mutable-string-dict:], [SD.mutable-string-dict:])
   unbound1 = check-unbound-ids-bad-assignments(px, resolved, C.standard-builtins)
   unbound1.length() is 1
 end
--- a/src/arr/compiler/resolve-scope.arr
+++ b/src/arr/compiler/resolve-scope.arr
@@ -6,6 +6,7 @@ import ast as A
 import srcloc as S
 import parse-pyret as PP
 import string-dict as SD
+import lists as L
 import file("compile-structs.arr") as C
 import file("ast-util.arr") as U
 import file("gensym.arr") as G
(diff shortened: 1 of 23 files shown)


src/arr/compiler/resolve-scope.arr
```diff
@@ -118,23 +119,147 @@ fun desugar-toplevel-types(stmts :: List<A.Expr>) -> List<A.Expr> block:
...
+    if      A.is-s-let-bind(bind): A.s-let-bind(bind.l, new-b, new-v)
+    else if A.is-s-var-bind(bind): A.s-var-bind(bind.l, new-b, new-v)
+    else if A.is-s-letrec-bind(bind): A.s-letrec-bind(bind.l, new-b, new-v)
+    end
+  end
+  fun names-match(funargs :: List<A.Bind>, annargs :: List<A.AField>):
+    if is-empty(funargs) and is-empty(annargs): true
+    else if is-empty(funargs) or is-empty(annargs): false
+    else:
+      (funargs.first.id.toname() == annargs.first.name) and names-match(funargs.rest, annargs.rest)
```
jpolitz: This makes me wonder about underscore names

src/arr/compiler/resolve-scope.arr
```diff
@@ -118,23 +119,147 @@ fun desugar-toplevel-types(stmts :: List<A.Expr>) -> List<A.Expr> block:
...
+          | none => link(fun-to-lam(bind), acc)
+          | some(c) =>
+            contracts-sd.remove-now(id-name)
+            if A.is-a-blank(ann) block:
+              if not(c.l.before(bind.value.l)) block:
+                errors := link(C.contract-bad-loc(c.l, id-name, bind.value.l), errors)
+                link(fun-to-lam(bind), acc)
+              else:
+                cases(A.Expr) bind.value:
+                  | s-fun(l-fun, name, params, args, ret, doc, body, _check-loc, _check, blocky) =>
```
jpolitz: This seems weird to me (as in, I probably just need an explanation). What is a `s-fun` doing on the RHS of a binding?

src/arr/compiler/resolve-scope.arr
```diff
@@ -231,8 +387,10 @@ fun desugar-scope-block(stmts :: List<A.Expr>, binding-group :: BindingGroup) ->
           add-letrec-bind(binding-group, A.s-letrec-bind(
               l,
               A.s-bind(l, false, A.s-name(l, name), A.a-blank),
-              A.s-lam(l, name, params, args, ann, doc, body, _check-loc, _check, blocky)
-            ), rest-stmts)
+              # NOTE(Ben): deliberately keeping this as an s-fun;
+              # it'll get turned into an s-lam in weave-contracts
```
jpolitz: This comment says _what_ is happening, but not _why_. I think the reason is that we want to give errors for contract-ing anything that isn't an `s-fun`, right?

src/arr/compiler/well-formed.arr
```diff
@@ -892,7 +896,10 @@ well-formed-visitor = A.default-iter-visitor.{
   end
 }
 
-top-level-visitor = A.default-iter-visitor.{
+top-level-visitor = well-formed-visitor.{
+  # Most behaviors are delegated to the non-toplevel visitor,
+  # but these expression forms are handled specially at the toplevel:
```
jpolitz: I'm suspicious of this change, but can't point to the (regression) test that would fail because of it. Something about toplevel vs. non-toplevel checks about `data` is flashing in my memory.

src/arr/compiler/well-formed.arr
```diff
@@ -992,228 +999,25 @@ top-level-visitor = A.default-iter-visitor.{
...
-  method s-method-field(_, l :: Loc, name :: String, params :: List<A.Name>, args :: List<A.Bind>, ann :: A.Ann, doc :: String, body :: A.Expr, _check-loc :: Option<Loc>, _check :: Option<A.Expr>, blocky):
-    well-formed-visitor.s-method-field(l, name, params, args, ann, doc, body, _check-loc, _check, blocky)
+    # TODO
+    true
   end,
-  method s-for-bind(_, l :: Loc, bind :: A.Bind, value :: A.Expr):
-    well-formed-visitor.s-for-bind(l, bind, value)
+  method s-app(self, l :: Loc, _fun :: A.Expr, args :: List<A.Expr>) block:
+    if (A.is-s-dot(_fun) and A.is-s-id(_fun.obj)
+        and (_fun.obj.id.toname() == "builtins") and (_fun.field == "trace-value")):
```
jpolitz: So many abstraction-crossing string checks!

src/arr/trove/lists.arr
```diff
@@ -887,6 +887,30 @@ fun distinct(l :: List) -> List:
...
+        if pred(first) block:
+          link(first, help(rest))
+        else:
+          tail := l
+          empty
+        end
+    end
+  end
+  { help(lst); tail }
+where:
```
jpolitz: These should probably go in test-lists.arr; the status of tests within trove files is somewhat murky.

tests/pyret/tests/test-contracts.arr
```diff
@@ -319,52 +295,120 @@ check "should work in cases positions":
...
+    ```) is%(output) compile-error(CS.is-contract-inconsistent-names)
+  run-str(
+    ```
+    is-even :: Number -> Boolean
+    fun is-odd(n): true end
+    is-odd :: Number -> Boolean
+    fun is-even(n): true end
+    ```) is%(output) compile-error(lam(e): CS.is-contract-bad-loc(e) and (e.name == "is-odd") end)
 end
-|#
```
jpolitz: 🥇 💯 👍 🎉 

Hooray for reviving old tests!!!

tests/pyret/tests/test-contracts.arr
```diff
@@ -319,52 +295,120 @@ check "should work in cases positions":
...
-  run-str("x :: {String; String} = {\"sewhwr\"; 4124}") satisfies is-contract-error
-  run-str("x :: {Number; String; {Number; Number; {String}}; String} = {4124; \"frwfq\"; {5123;531;{5351}}; \"fqwfq\"}") satisfies is-contract-error
-  run-str("x :: {Number; Number; {Number; String}} = {412; 5412; {412; \"fgwdef\"; 5135}}") satisfies is-contract-error
+  run-str("x :: {Number; String; {Number; Number; {String}}; String} = {4124; \"frwfq\"; {5123;531;{\"fqswf\"}}; \"fqwfq\"}") is%(output) success
+  run-str("x :: {String; String} = {\"sewhwr\"; 4124}") is%(output) contract-error
+  run-str("x :: {Number; String; {Number; Number; {String}}; String} = {4124; \"frwfq\"; {5123;531;{5351}}; \"fqwfq\"}") is%(output) contract-error
+  run-str("x :: {Number; Number; {Number; String}} = {412; 5412; {412; \"fgwdef\"; 5135}}") is%(output) contract-error
+end
+
+check "standalone contract statements":
```
jpolitz: I don't see a test with _multiple_ named arguments in a contract, which could be useful for making sure reverse errors aren't introduced.

src/arr/compiler/resolve-scope.arr
blerner: Underscores don't matter here: we're weaving the annotations onto the argument names before we desugar away underscores.

src/arr/compiler/resolve-scope.arr
blerner: See your next comment.  This is to address [your comment](https://github.com/brownplt/pyret-lang/issues/1114#issuecomment-321822575) about wanting to only weave annotations onto explicit functions, rather than lambdas.  Since functions get desugared away *very* quickly, I needed to reinstate them for a little while longer, until I could check for them in `weave-contracts`.  Then, once weave-contracts is finished, all functions will be turned into lambdas.

src/arr/compiler/resolve-scope.arr
blerner: Not quite; that's too harsh.  We want to be able to say
```
G :: Number
G = 12345
```
and ensure that `G` turns out ok.  But for arrow contracts, you'd wanted there to be a distinction between `lam` and `fun` in terms of how thoroughly the annotation is woven in.  Currently, we weave the annotation onto the let binding, meaning "we'll check that it's a Function, but nothing more detailed than that".  But for literal `fun` declarations, we need to preserve that information until `weave-contracts`

src/arr/compiler/well-formed.arr
blerner: Yeah, it's broken.  Reverting.

src/arr/compiler/well-formed.arr
blerner: Yeah :(  This one's necessary because we allow `type-let` as a block-ending expression, but only at the top level.  But `trace-value` will wrap all top-level, block-ending expressions...which makes the type-let no longer be a top-level expression!  (At least we had a regression test for this...)

src/arr/trove/lists.arr
blerner: Separate issue.  I was following the exemplars in the existing lists file; we should clean it all up at once.

tests/pyret/tests/test-contracts.arr
blerner: Good point.  It's not likely, since I'm walking over the lists of arguments and annotations in lockstep, rather than via fold2 or something (because I need to efficiently check whether they're the same length, too), but I'll add a test.

PR #1135, opened 2017-08-14
Allow remainder to take non-int args

ds26gte:
Fix for brownplt/code.pyret.org#215.

`jsnums.remainder` is a purely a utility function. There is no corresponding `num-remainder` in Pyret to test. `rotate` in `code.pyret.org` appears to be its sole user at this point, where the restriction to integer values was clearly harmful.


blerner:
Two questions: why should non-integer inputs immediately devolve to roughnums?  And, are there any tests for correctness of this, for non-integer or negative inputs (both inputs!)?

ds26gte:
- The sign of the result follows JavaScript's `remainder` operator (which is called `%`, but it's really remainder, not modulo, as we may have discussed some time back), and the sign semantics (if not the argument signature) is identical to the Racket version.

- Yes, it might be worth it to do some fine-tuning for arguments that are not rough but are non-integral. Will try. Again, `rotate`/`canonicalize-angle` is really the only client, and it will fixnum-ize the final number of degrees anyway, so we won't see a "more perfect" rotation even after the mod.

blerner:
Right, and num-modulo will restrict the inputs to be integers, so the roughness shouldn't be an issue at the Pyret surface level.  

Still, would like to see some tests for negative numbers so that we document our expectations rather than just say "whatever JS does".

ds26gte:
The expectation is this: `remainder(x,y)` always takes the sign of `x`, and the absolute values of all four `remainder(±x, ±y)` are identical.

Re the tests: As I said, Pyret doesn't provide a `num-remainder` to run tests on. Where would tests of utility functions such as `jsnums.remainder` go? 

(I'll be updating the PR later today to get precise (rational) remainders for rational arguments, stooping to roughnum only when contagion demands it. FYI, my proposed approach will find the LCM of the arguments, and do integer remainder on the re-calculated numerators, unless you can think of a better way.)

blerner:
`tests/runtime/runtime-numbers.js` seems like the best place.

I'm a little surprised still by the sign expectation -- sorry if this is a rehash; I just don't remember the discussion.  The equations are supposed to look like this?

```
5 % 3 == 2 because 5 == 1 * 3 + 2
-5 % 3 == -2 because -5 == -1 * 3 + -2
5 % -3 == 2 because 5 == -1 * -3 + 2
-5 % -3 == -2 because -5 == 1 * -3 + -2
```

Probably worth writing that example down somewhere, to remind forgetful folks like me how to math ;-)

ds26gte:
Pretty much. The idea is given `M` and `N` we define their integer quotient `Q` to be the truncation of `M/N`, i.e., the zero-directed floor. Then the remainder `R` is whatever you get that satisfies `M == N*Q + R`.

This definition of quotient is slightly arbitrary so I wouldn't beat myself up for not internalizing it, but it is apparently the most common definition (there are others) according to https://en.wikipedia.org/wiki/Modulo_operation#Remainder_calculation_for_the_modulo_operation.

Anyway, this is the definition I've seen in all the languages that I've seen (!) that do have a remainder op. As the article says, Knuth has a different notion of quotient, but I just quickly confirmed that even he settled for the mainstream definition for TeX.

ds26gte:
I've added tests to `tests/runtime/runtime-numbers.js`, but `make runtime-test`, being unused for long, appears to have rotted considerably...

jpolitz:
Proposal: let's add `num-remainder`, which is the same as `num-modulo` but with the domain being all numbers. It will be easier to test from Pyret-land that way, and both functions are useful.

blerner:
Joe's got a good point, that adding code to a JS file that doesn't run isn't a good idea...

ds26gte:
Sounds good, but please note that `remainder` and `modulo` are not necessarily identical when both converge (and it's not just a matter of signs).

```
num-remainder(5, 3) => 2
num-remainder(5, -3) => 2
num-remainder(-5, 3) => -2
num-remainder(-5, -3) => -2

num-modulo(5, 3) => 2
num-modulo(5, -3) => -1
num-modulo(-5, 3) => 1
num-modulo(-5, -3) => -2
```

blerner:
Gah.  Ok.  Proposal: `num-remainder` works on any numbers, and `num-modulo` works only on non-negative integers -- because if you're trying to do modular arithmetic on negative numbers, please stop and reconsider your mathematical life choices.  Or something 8-)

Joe points out there's at least one more well-defined function here, such that `num-joes-remainder(m, n)` should never return a negative number at all, because "remainder" implies "is left over", which doesn't feel like a negative quantity.

@schanzer, @shriram, do either of you have a pedagogical reason to favor one of these specs over the others?

schanzer:
I believe Euclidean division requires that the remainder always be positive. [Here's an answer from The Math Forum](http://mathforum.org/library/drmath/view/75805.html), which covers our simpler numeric case as well as a discussion of dividing polynomials.

ds26gte:
That example only shows the 17 ÷ 5 is better resolved as {quotient 3, remainder 2} rather than {quotient 4, remainder -3}. No one would prefer otherwise (3 is the floor of 17/5; 4 isn't.) The choice here has no bearing on keeping the remainder positive when the dividend is negative.

I don't think we should change the results produced by remainder and modulo lightly. Integer-remainder is used internally a lot, and changing its behavior would break things within jsnums, and the rewriting to solve the breakage will not be clean. 

Extending the domain of math functions to negative (or fractional, or irrational, or imaginary) inputs only has to work with the rest of math -- that's the whole point of domain extension. It doesn't have to be prima facie intuitive; rather, we use a playing-nice solution to build up our intuition after the fact. (That said, one can certainly build intuition that tolerates negative remainders. Think of debts: a left-over debt is still a debt.)

src/js/base/runtime.js
```diff
@@ -4739,6 +4739,16 @@ function (Namespace, jsnums, codePoint, util, exnStackParser, loader, seedrandom
         [thisRuntime.NumInteger, thisRuntime.NumInteger]);
       return thisRuntime.makeNumberBig(jsnums.modulo(n, mod, NumberErrbacks));
     }
+
+    var num_remainder = function(n, m) {
+      if (arguments.length !== 2) { var $a=new Array(arguments.length); for (var $i=0;$i<arguments.length;$i++) { $a[$i]=arguments[$i]; } throw thisRuntime.ffi.throwArityErrorC(["num-remainder"], 2, $a, false); }
+      thisRuntime.checkArgsInternal("Numbers", "num-remainder",
+        [n,                      m],
+        [thisRuntime.Number, thisRuntime.Number]);
+      console.log('doing num_remainder', n, m);
```
blerner: Stray console.log -- please delete

ds26gte: Oops; thanks; done.

PR #1139, opened 2017-08-16
Method to the tabular madness

jpolitz:
A bunch of new method-based infrastructure for tables.

Docs are at https://www.pyret.org/docs/horizon/tables.html#%28part._s~3atables~3amethods%29

@jswrenn I could use some advice – there are a lot of placeholder `throwMessageException`s in the methods for this PR. What is the best way for me to turn them into good errors. For example, should I be creating new error variants for each new kind of error, or is there a more generic form I should be using in many cases? An error-review would be helpful for me.


jswrenn:
I'll take a close look at the errors that are here tomorrow. Providing detailed error reports for library routines is an underdeveloped area of Pyret's error reporting. To answer your question: the best thing to do is to implement a distinct variant for each error.

src/js/trove/load-lib.js
```diff
@@ -333,7 +333,23 @@
           // NOTE(joe): This is the place to add checkAll
           var currentChecker = otherRuntime.getField(checker, "make-check-context").app(otherRuntime.makeString(main), checkAll);
           otherRuntime.setParam("current-checker", currentChecker);
-        }
+        },
+        "builtin://table": function(table) {
```
blerner: Crud.  I'm noticing that this set of postLoadHooks is diverging from handalone.js (and probably cpo-standalone) too.  Already, handalone mentions "throwInvalidTableColumn" and "toArray" in the ffi section, and this doesn't; this doesn't include builtin://data-source or builtin://reactors.  We need to make sure these three loaders stay in sync.

src/js/trove/table.js
```diff
@@ -13,10 +13,87 @@
...
+          if(headers1[hk1[i]] !== headers2[hk1[i]]) {
+            return neq(self, other);
+          }
+        }
+        return runtime.raw_array_fold(runtime.makeFunction(function(ans, val1, j) {
+          if (ffi.isNotEqual(ans)) { return ans; }
+          return runtime.safeCall(function() {
+            return rec.app(val1, rowData2[j]);
+          }, function(eqAns) {
+            return get(EQ, "equal-and").app(ans, eqAns);
```
blerner: Hoist the call to `equal-and` out of the loop?

src/js/trove/table.js
```diff
@@ -13,10 +13,87 @@
...
+      for(var i = 0; i < rawArrayOfTuples.length; i += 1) {
+        var colname = rawArrayOfTuples[i].vals[0];
+        if(headerIndex["column:" + colname] !== undefined) {
+          return runtime.ffi.throwMessageException("Duplicate column name in row: " + colname);
+        }
+        headerIndex["column:" + colname] = i;
+        rowData[i] = rawArrayOfTuples[i].vals[1];
+      }
+
+      return makeRow({ headerIndex: headerIndex }, rowData);
```
blerner: I see: this `{headerIndex}` object is just enough of an `$underlyingTable` to make the row work successfully?

src/js/trove/table.js
```diff
@@ -227,17 +308,52 @@
...
+          runtime.checkString(colname);
           return order(false, colname);
         }),
+        'order-by': runtime.makeMethod2(function(_, colname, increasing) {
+          ffi.checkArity(3, arguments, "order-by", true);
+          runtime.checkString(colname);
+          runtime.checkBoolean(increasing);
+          return order(increasing, colname);
+        }),
+        'order-by-columns': runtime.makeMethod1(function(_, specs) {
```
blerner: How is this method different from `multi-order`?  Multi-order takes an array of 2-element arrays, and this takes a list of 2-tuples?  If so, I'd rather
* remove `multi-order` as a Pyret-visible method on tables, in favor of `order-by-columns`,
* keep the implementation of `multiOrder` available as a JS-callable function on tables, and
* change the use of `multi-order` in table-desugaring to call that helper function directly (to keep the desugaring efficient)

src/js/trove/table.js
```diff
@@ -248,6 +364,114 @@
...
+
+        'column': runtime.makeMethod1(function(_, colname) {
+          ffi.checkArity(2, arguments, "column", true);
+          runtime.checkString(colname);
+          var lookupName = "column:" + colname;
+          if(!(lookupName in headerIndex)) {
+            throw runtime.ffi.throwMessageException("no-such-column");
+          }
+          var results = rows.map(function(r) {
+            return r[headerIndex[lookupName]];
```
blerner: Can hoist `headerIndex[lookupName]` out of the loop too, since `lookupName` is invariant

src/js/trove/table.js
```diff
@@ -248,6 +364,114 @@
...
+          runtime.checkNumNonNegative(row);
+          runtime.checkNumInteger(row);
+          var rowFix = runtime.num_to_fixnum(row);
+          if(rowFix >= rows.length) {
+            throw runtime.ffi.throwMessageException("row-n-too-large");
+          }
+          return makeRow({ headerIndex: headerIndex }, rows[rowFix]);
+        }),
+
+        'all-rows': runtime.makeMethod0(function(_) {
```
blerner: Interesting that this method returns a list of rows, whereas all-columns returns a list of lists.  It makes sense from one angle, and is inconsistent from another.  Worth calling out in the documentation.

src/js/trove/table.js
```diff
@@ -366,11 +596,39 @@
...
+            for(var j = 0; j < colnamesList.length; j += 1) {
+              var colIndex = headerIndex['column:' + colnamesList[j]];
+              collists[i].push(rows[i][colIndex]);
+            }
+          }
+          return makeTable(colnamesList, collists);
+        }),
+
         
         '_column-index': runtime.makeMethod3(function(_, operation_loc, table_loc, col_name, col_loc) {
```
blerner: Does this need to remain a Pyret-visible method?  Feels awkward, and is only used by desugaring.  (Also, it sure ain't a `makeMethod3`, when I see five arguments in it...)  Ditto for `_no-column` below.

src/js/trove/table.js
```diff
@@ -435,15 +691,37 @@
...
+        }),
+
+        'row': runtime.makeMethodN(function(self, ...args) {
+          if(headers.length !== args.length) {
+            throw runtime.ffi.throwRowLengthMismatch(makeTable(headers, []), args);
+          }
+          return makeRow({ headerIndex: headerIndex }, args);
+        }),
+
+        'new-row': runtime.makeObject({
```
blerner: Is this different from `raw-row`?  It's not documented...

src/js/trove/table.js
jpolitz: Yeah – what I actually want to do (but didn't need to in order to make things work) is to construct a row out of a table and a raw array of values. The table will have an underlying array of headers, header index, and rows (which its methods close over now).

This would be a big performance win for programs that use lots of tables. For this PR, I just did the headerIndex object since that was all that was actually needed to implement row.

src/js/trove/table.js
jpolitz: You answered your question in the sentence that followed, so I won't add to that.

I don't think `multi-order` can be Pyret-invisible if the desugaring needs to call it, because it's not stack-safe because less than could blow the stack.

src/js/trove/table.js
```diff
@@ -248,6 +364,114 @@
...
+          runtime.checkRow(row);
+          var theseKeys = Object.keys(headerIndex);
+          var rowKeys = Object.keys(row.$underlyingTable.headerIndex);
+          if(theseKeys.length !== rowKeys.length) {
+            throw runtime.ffi.throwMessageException("add-row-length");
+          }
+          for(var i = 0; i < theseKeys.length; i += 1) {
+            if(headerIndex["column:" + headers[i]] !== 
+                row.$underlyingTable.headerIndex["column:" + headers[i]]) {
+              throw runtime.ffi.throwMessageException("row-name-mismatch: " + headers[i]);
```
jpolitz: Names must match up in the correct positions, so if you append an `a,b,c` row to a `a,c,b` table, it's an error.

src/js/trove/table.js
jpolitz: Yeah it's interesting to ask if there should be corresponding constructors/datatypes for `column`, just like are added her for `row`.

src/js/trove/table.js
```diff
@@ -366,11 +596,39 @@
...
           return runtime.ffi.makeList(getColumn(col_name));
         }),
+
+        'select-columns': runtime.makeMethod1(function(_, colnames) {
+          ffi.checkArity(2, arguments, "get-column", true);
+          runtime.checkList(colnames);
+
+          var colnamesList = ffi.toArray(colnames);
+          if(colnamesList.length === 0) {
+            throw ffi.throwMessageException("zero-columns");
```
jpolitz: This is actually a well-formedness error on tables right now. But since you can't drop columns, and can't select zero columns, it can be entirely enforced syntactically.

src/js/trove/table.js
jpolitz: Good point on not documenting it.

See the tests:

```
t = table: a, b
    row: 1, 2
  end

  [t.new-row: 3, 4] is [raw-row: {"a"; 3}, {"b"; 4}]
```

src/js/trove/load-lib.js
jpolitz: Good catch! I've made a note to sync these before merging.

src/js/trove/table.js
```diff
@@ -596,11 +596,39 @@
...
         'get-column': runtime.makeMethod1(function(_, col_name) {
           ffi.checkArity(2, arguments, "get-column", true);
+          runtime.checkString(col_name);
           if(!hasColumn(col_name)) {
             ffi.throwMessageException("The table does not have a column named `"+col_name+"`.");
           }
           return runtime.ffi.makeList(getColumn(col_name));
         }),
+
+        'select-columns': runtime.makeMethod1(function(_, colnames) {
```
jswrenn: Why does `select-columns` consume a `List` of column names, whereas `row` consumes a variadic number of values?

jpolitz:
@jswrenn asks:

> Why does select-columns consume a List of column names, whereas row consumes a variadic number of values?

Good question! The answer might be that it should be more like `row`. Let's start with a simpler question, since I don't think this argument is written out anywhere:

> why does `row` consumes a variadic number of values?

There's no other functions or methods in Pyret like it! Is that OK? Why try it here?

The answer is that, while we're intentionally loosening our type story for the table methods, I still want to have _some_ way to describe types for these heavily dependently-typed operations. I see two main contenders for row construction:

```
t = table: a, b, c end
[t.new-row: 1, 2, true]
t.row(1, 2, true)
```

(Note that the former can't be `t.row:` due to syntactic constraints.)

The current type system can't express the type of the `new-row` version of the row construction, because collection constructors are intended to have homogeneous elements provided. By adding the notion of a positional collection constructor, we could do this.

The way the documentation is written for `raw-row` shows this – there's a call-out that “note that the values in the columns may have different types”

The type system _can_ express the method type for `t.row`. It would have to synthesize the type from each syntactically appearing `table` expression, and there would be additional complications for programmatic construction of tables (e.g. after `build-column`). But, the type is expressible as a regular method type. This is useful for documentation, and I claim pedagogically, because we can re-use knowledge about contracts in describing `row` creation.

I don't actually know which of these two is going to see more use, going to be more sensible to type-check, etc. So I want to provide both in this version of the library to see what's most effective.

Back to the original question:

> Why does select-columns consume a List of column names, whereas row consumes a variadic number of values?

I had it in my head during the design that construction of new rows was the really special thing here, and that operation alone deserved this special treatment.

However, thinking forward to type-checking this, one other requirement for the type-checker will be added support for reasoning about individual strings, and finite collections of strings, as types. That's just what's necessary to statically check something like `column`, whose list contents can only be statically known by reasoning about the string argument.

So if row types are going to be synthesized from table construction, and singleton string types are going to be added to the type language, then it's no harder to construct the right type for `select-columns` than it is to construct the type for `row`.

There is a good deal of syntactic convenience in the difference, as well as avoiding the dependency on lists:

```
t.select-columns("city", "population")
```

over

```
t.select-columns([list: "city", "population"])
```


I'm fairly convinced by this argument that `select-columns` should get special treatment as well.

blerner:
> One other comment: these examples are slightly less strong than full variadic functions. If we consider any individual row method on a particular table, it has a specific arity. Different tables may have row methods with different arities, but the arity is fixed once the table is created.

Completely agree.  So the `row` method really is a _set of methods_, one per table, each of which has a perfectly standard method type.  But `select-columns` is different: making it arity-parametric is a completely new thing for Pyret.

Also -- if we *really, really* want to, we can hack the syntax for `[t.row: ...]` to make `row:` work there.  It'll be a gross hack, (the method is named `row`, followed by a colon, both of which will be recognized by the single token `row:`...), but it could be forced to work.

jswrenn:
Of interest to me: using arity-parametric methods is a godsend for providing good error reports. If `select-columns` needs to display an error that refers to some n<sup>th</sup> column, I can provide a much better error message if `select-columns` is arity parametric (since I can highlight the n<sup>th</sup> argument); if `select-columns` consumes a `List`, it's generally impossible to highlight the n<sup>th</sup> element of the list.

src/js/trove/table.js
blerner: Right -- I'm saying we should ditch the syntactic restriction, since the well-formedness error is a lot more polite.

jpolitz:
Oh @blerner's comment convinces me that I was wrong when I said:

> it's no harder to construct the right type for select-columns than it is to construct the type for row.

It is harder, because `select-columns` is truly variadic, while `row` isn't.

So that's actually an argument for not doing variadic `select-columns`, since it's _truly_ variadic. It would introduce the need to talk about, document, and express methods that can take different numbers of arguments on different invocations.

src/js/trove/table.js
jpolitz: Got it; I missed the difference b/w `table:` and `select` syntax vs. wf errors

jswrenn:
@jpolitz: The argument about error message here still stands. If `select-columns` is variadic, the error message it displays can actually highlight the problematic columns. Disadvantage: this would be _less_ general than the version that consumes a list, since Pyret lacks `apply`; @blerner's suggestion could provide this more general interface.

jswrenn:
Providing good error reports for some of these is going to be difficult. For instance, for this program:
```
[raw-row: {"a"; 1}, {"a"; 2}]
```
...we can present the error message purely as a matter of syntactic terms, because the column names are given as literals.

If they are _not_ literals, e.g.:
```
a = "a"
b = "a"
[raw-row: {a; 1}, {b; 1}]
```
...then we can and should still highlight the occurrences `a` and `b` on line three, but now the prose of the report needs to comment on the fact that these terms _evaluated_ to _"a"_ and _"b"_, respectively.

Similarly, if this is the program:
```
a = {"a"; 1}
b = {"a"; 2}
[raw-row: a, b]
```
...then we should still probably highlight the occurrences of `a` and `b` on line three, but now the prose of the report needs to comment on the fact that these terms evaluated to tuples, of which only the first component of each is actually important.

And these are only the common cases. A well-formed error report still needs to be given even if the the program is this:
```
raw-row.make([raw-list: {"a"; 1}, {"a"; 2}])
```

Similar challenges exist for `select-columns`, unless we make `select-columns` variadic.

blerner:
Adding variadic functions to Pyret is not something to be taken lightly, and I'm not inclined to do so for the sake of error messages -- we'd need to need a linguistic or pedagogical reason to want that to be exposed to programmers.  (Because every "I've coded in C for five minutes!" novice programmer is going to jump on that ability, for no particularly good reason, and then you'll have a huge variety of other errors to deal with.)

Additionally, remember that in all the examples you give above, you do not get srclocs for the subterms because these are *just* normal function calls, not customized syntax (unlike `table` and query syntax).  You have to make the messages work given the same information you get for normal function calls.

jswrenn:
> You have to make the messages work given the same information you get for normal function calls.

Well aware.

> Adding variadic functions to Pyret is not something to be taken lightly, and I'm not inclined to do so for the sake of error messages

To be super clear: I _am_ not arguing we add variadic method definitions to the Pyret _language_, just that we use our existing ability to make javascript-implemented variadic methods on `select-columns`, just as we did for the `row` method.

There are pedagogical reasons for doing this:
 - removes the dependency on `list` (https://github.com/brownplt/pyret-lang/pull/1139#issuecomment-323111057)
 - improves the error report that's provided for a common operation ("for the sake of error messages" _is_ a pedagogical reason)

@jpolitz wrote:
> select-columns is truly variadic, while row is [arity parametric].

While I think this is little overly-subtle (at the end of the day, we're trying to provide an ergonomic interface), I think a similar argument extends to `select-columns`. The `select-columns` method is _bounded_ arity parametric: the upper bound on its arity is exactly equal to the number of columns present in the table it's called on.

blerner:
It may be subtle, but it's a crucial distinction.  The `row` method *for any given table* has a unique, fixed arity.  Each table has its own customized method (which knows the appropriate arity and the appropriate column names).  The proposed `select-columns` method does not; it is therefore variadic and not part of Pyret's current design.  There are no variadic functions in Pyret, except for `[constructor: ...]` syntax, which implicitly builds a list around its arguments, and still presents as a single-arity function to implementors of new convenience-constructor objects.

jswrenn:
@jpolitz Is this intended behavior?
```
include tables

data Foo: foo(bar) end

a = foo(1)
b = foo(2)

# Both tests pass.
check:
  a is-not b
  [raw-row: {a; 1}, {b; 2}] raises "Duplicate column name"
end
```

jpolitz:
Nope, that's a bug. The `raw-row` constructor should make sure the columns are strings.

jswrenn:
> The row method for any given table has a unique, fixed arity.

@blerner, given that criterion, should `add-column` _also_ have parametric arity? Currently it consumes a list, but the length of that list for any given instance of table is always fixed.

blerner:
As a semantically pristine approach, perhaps yes. As a practical matter, not at all. I'd almost prefer an API `add-columns` that took a table as input, and checked for lack of duplicate column names, and then adjoined the given table to the right of the current one. That would have nice duality with `add-rows` that took a table, checked for exactly identical column names, and adjoined the given table beneath the current one.

But the point of this API is to allow programmatically creating a table dynamically from its contents, rather than syntactically. And to that end, I think the current API is probably the right choice.

jpolitz:
Yup, it should prevent it, that's a bug, thanks!

With the latest pushes do you think the errors are pretty much covered, or is there significant work to do?

I'm assuming we'll come up with a reasonable solution for https://github.com/brownplt/pyret-lang/issues/1144

jpolitz:
@jswrenn I'm going to merge this into horizon so we can more actively play with it. Still curious if you think these are finished or not.

jpolitz:
Also @shriram, it looks like @jswrenn requested feedback above, but hasn't gotten it yet.

Issue #1147, opened 2017-08-25
obiwan error on num-to-roughnum(<bignum>)


ds26gte:
`num-to-roughnum(1e100)` returns `~1e99`. Tests apparently didn't catch it because we were converting ratios of bignums (so the `-1`s cancel), though it's not clear why
```
check "comparing rough MAX_VALUEs":
  ~1.7976931348623157e+308 - ~1.7976931348623157e+308 is%(within-abs(0.001)) ~0
   1.7976931348623157e+308 - ~1.7976931348623157e+308 is%(within-abs(0.001)) ~0
end
```
(which contains an implicit `num-to-roughnum` of a bignum) passed during `make test` but not when explicitly tested in the IDE.

blerner:
Looks like this is fixed; closing.

PR #1148, opened 2017-08-28
Fixed off-by-1 error in exponent while converting bignum to roughnum.

ds26gte:
Fix for #1147. (Pure bugfix -- no functionality added or deleted. Test updated.)


jpolitz:
Any responses to @blerner's comments here? Seems like this is wanting one more test, maybe?

sorawee:
This is really important to be fixed for the plot library to be functional. I need to patch this locally when I developed it, or else the domain and range inferring will be completely incorrect.

PR #1150, opened 2017-08-31
pri[VI]teer - vim + ctags config 

rachitnigam:
From the heretical lands where the church of emacs does not exist, I bring to you, VIM:

- Significantly enhance vim highlighting support for .arr files. Automatic detection for .arr files.
- Code snippets insertion for common code blocks.
-  ctags support for jumping to function / type definitions and imports.

### DIFF
--- a/tools/ctags/pyret.ctags
+++ b/tools/ctags/pyret.ctags
@@ -1,0 +1,11 @@
+--langdef=pyret
+--langmap=pyret:.arr
+
+--regex-pyret=/^[ \t]*fun[ \t]+([a-zA-Z0-9_-]+)/\1/f,function/
+--regex-pyret=/^[ \t]*data[ \t]+([a-zA-Z0-9_-]+)/\1/d,data/
+--regex-pyret=/^[ \t]*\|[ \t]*([a-zA-Z0-9_-]+)[^=]*$/\1/d,data/
+--regex-pyret=/^[ \t]*\|[ \t]*([a-zA-Z0-9_-]+)[^=]*$/is-\1/d,data/
+--regex-pyret=/^([a-zA-Z0-9_-]+)[ \t]*=/\1/v,value/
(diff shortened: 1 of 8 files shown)


blerner:
I don't know of any of us are capable of reviewing this particular PR for you... 😎

rachitnigam:
@blerner Fixed the issues you pointed out.

ds26gte:
Curious what the addition of the two comment lines (the only change) to `indent/pyret.vim` is for. It seems to merely identify the original location of the file as itself; why? (And none of the other files have this kind of self-reference.)

rachitnigam:
@ds26gte that's because I had a separate repo where I was pushing all this code: https://github.com/rachitnigam/pyret-lang.vim

Keeping it in a separate repo helps with vim plugin managers that can directly update the plugin.

I didn't think this repo would want that repo as a submoodule so I moved the latest code here.

ds26gte:
Ah. Sounds good. And since you seem to be way more versed in VimL than I, please do feel free to improve `indent/pyret.vim` at your convenience: it was a very naive effort on my part and isn't comprehensive.

jpolitz:
I've had this on for a week now. Is there any reason it could cause searching to become SLOW? Like 10-20 seconds lag when I press enter after a search like `/<<<<<`?

rachitnigam:
I haven't noticed a slowdown compared to JS Files of similar sizes. It's pretty lightweight for a syntax highlughter (most of it is just color coding keywords)

jpolitz:
I still see extreme slowdowns with this in terminal vim on OSX with iterm2, bad enough that I can't navigate around files over a few 100s of lines reasonably. Interestingly, the same doesn't happen in GUI mode Macvim.

rachitnigam:
That's curious. I haven't had problems with the newest neovim and vim versions either.

PR #1159, opened 2017-09-07
Type checker builtins

### LINKED ISSUE #1158: importing built-in modules fails with type-checker in REPL
image

The same thing doesn't seem to happen when running with `-type-check` offline, but if we add, in repl-tests:

```

  tc-result = restart("import lists as L", true)
  tc-result.v satisfies L.is-success-result
  should-fail-map-use = next-interaction("L.map(4, 5)")
  should-fail-map-use.v satisfies L.is-failure-result
  should-succeed-constructor = next-interaction("[L.list: 1, 2, 3]")
  should-succeed-constructor.v satisfies L.is-success-result

```

We see the same failure at the first line.

mkolosick:
fixes #1158

### DIFF
--- a/src/arr/compiler/type-check.arr
+++ b/src/arr/compiler/type-check.arr
@@ -167,7 +167,12 @@ fun type-check(program :: A.Program, compile-env :: C.CompileEnvironment, module
     else:
       uri = globvs.get-value(g)
       # TODO(joe): type-check vars by making them refs
-      context.set-global-types(context.global-types.set(A.s-global(g).key(), compile-env.mods.get-value(uri).values.get-value(g).t))
+
+      if (g == "_"):
+        context
+      else:
+        context.set-global-types(context.global-types.set(A.s-global(g).key(), compile-env.mods.get-value(uri).values.get-value(g).t))
+      end
     end
   end, context)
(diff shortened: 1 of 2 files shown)


src/arr/compiler/type-check.arr
```diff
@@ -167,27 +167,36 @@ fun type-check(program :: A.Program, compile-env :: C.CompileEnvironment, module
     else:
       uri = globvs.get-value(g)
       # TODO(joe): type-check vars by making them refs
-      context.set-global-types(context.global-types.set(A.s-global(g).key(), compile-env.mods.get-value(uri).values.get-value(g).t))
+
+      if (g == "_"):
```
jpolitz: It is odd to me that this is necessary. It feels like something didn't get desugared correctly if we're getting here. But I understand that it's because the global environment is somewhat special, and the only context that deals in raw, un-atomized names.

mkolosick: This seems to be a weird interaction between the desugaring of `include` and how the repl treats its code as globals. It is probably worth looking into, but I wanted to get at least a quick (but correct) fix in, so I didn't do too much digging tonight.

Issue #1156, opened 2017-09-06
type checker bug?


shriram:
![image](https://user-images.githubusercontent.com/75030/30113475-d0e4d714-92e2-11e7-8a80-5e0a4272c1a9.png)

What am I missing here? This works just fine in the untyped language.

Since this example is critical to my first lecture, I can't actually show students the type checker, which means we're probably going to miss out on any data collection…

If you can identify the problem in the next 35 minutes, please email me directly (or even text).

sorawee:
This needs to be changed: https://github.com/brownplt/pyret-lang/blob/master/src/arr/compiler/type-defaults.arr#L130

blerner:
Go for it. Should be String -> list<String> right?

Issue #1125, opened 2017-08-04
redundant list-to-raw-array


sorawee:
There exists `raw-array-from-list` which checks for arity mismatch, etc. already. `list-to-raw-array` does exactly same thing, but has no arity check.

blerner:
Ironically, your plot libraries are the only client of `list-to-raw-array`... ;-)

sorawee:
Yup. I took those out from my local version already. When that get merged, I will submit another PR to clean this up. This issue is just a reminder.

sorawee:
Related: `raw-array-from-list` checks that the arity is 2 instead of 1....

blerner:
@sorawee is this finished, now?

schanzer:
Confirming that `list-to-raw-array` is no longer defined on CPO. Closing.

PR #1165, opened 2017-09-10
Fix #1164, #1156, #1125

### LINKED ISSUE #1164: rudimentary program doesn't type-check
A cs019 student wasted a bunch of time unable to do homework because he was unable to run

```
fun test(str :: String) -> List<Number>:
string-to-code-points(str)
end

check:
test("abc") is [list: 97, 98, 99]
end
```

(and various variants of it). The problem is that `string-to-code-points` has type `Any`, and so results in a type error. Once we got him out of the Type Checker mode, he's fine, and finally able to start on his homework.

Color me not happy. At all.

sorawee:


mkolosick:
This should also be cherry-picked onto master so it can be released.

src/js/base/runtime.js
```diff
@@ -5211,7 +5211,7 @@ function (Namespace, jsnums, codePoint, util, exnStackParser, loader, seedrandom
       'raw-array-from-list': makeFunction(raw_array_from_list, "raw-array-from-list"),
       'raw-array-join-str': makeFunction(raw_array_join_str, "raw-array-join-str"),
       'get-value': makeFunction(getValue, "get-value"),
-      'list-to-raw-array': makeFunction(function(l) { return thisRuntime.ffi.toArray(l); }, "list-to-raw-array"),
+      'list-to-raw-array': makeFunction(raw_array_from_list, "raw-array-from-list"),
```
blerner: I don't understand this line: I thought you were trying to get rid of this function entirely?  Why reimplement it, and change its name, if that's going to happen?

src/js/base/runtime.js
```diff
@@ -285,7 +285,7 @@ function (Namespace, jsnums, codePoint, util, exnStackParser, loader, seedrandom
         }
         s += " }";
         return s;
-        
+
```
blerner: Realistically, all this whitespace tweaking shouldn't be part of this PR, since it just increases the likelihood of merge conflicts when we try to cherry-pick it

src/js/base/runtime.js
sorawee: The plot library current uses `list-to-raw-array` because `raw-array-from-list` in `master` is not usable (buggy). Once this is merged to `master`, we can get rid of it.

sorawee:
Really sorry for all whitespaces. I seem to have a plugin in both git and atom that make these invisible :( I will be more careful next time.

PR #1168, opened 2017-09-11
Fixes #1167

### LINKED ISSUE #1167: `+` on lists not recognized by type checker
`[list: 1, 2] + [list: 3, 4]` works fine in untyped but not in typed code. Is that intentional? If so, I would like to request a better error message: the current error highlights the `list` and says "Unable to infer the type of the expression. Please add an annotation", which is baffling.

Worse, it happened to a student who reported the following:
```
Unable to infer the type of the expression at
splitter(r,[list: ""] + titles)
Please add an annotation.
I have an annotation in the first line of the function, so I am not sure how to proceed:

splitter(chars :: List<String>, titles :: List<String>)-> List<String>:
```
Sure enough, this is a pretty baffling response to fix.

mkolosick:
This fixes a substitution bug in the constraint solver as discussed in #1167.

### DIFF
--- a/src/arr/compiler/type-check-structs.arr
+++ b/src/arr/compiler/type-check-structs.arr
@@ -431,8 +431,8 @@ sharing:
   method solve-level-helper(self, solution :: ConstraintSolution, context :: Context) -> FoldResult<{ConstraintSystem; ConstraintSolution}>:
     solve-helper-constraints(self, solution, context).bind(lam({system; shadow solution}, shadow context):
       solve-helper-refinements(system, solution, context).bind(lam({shadow system; shadow solution}, shadow context):
-        solve-helper-fields(system, solution, context).bind(lam({shadow system; shadow solution}, shadow context):
-          solve-helper-examples(system, solution, context)
+        solve-helper-examples(system, solution, context).bind(lam({shadow system; shadow solution}, shadow context):
+          solve-helper-fields(system, solution, context)
         end)
       end)
     end)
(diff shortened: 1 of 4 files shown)


src/arr/compiler/type-check-structs.arr
```diff
@@ -431,8 +431,8 @@ sharing:
   method solve-level-helper(self, solution :: ConstraintSolution, context :: Context) -> FoldResult<{ConstraintSystem; ConstraintSolution}>:
     solve-helper-constraints(self, solution, context).bind(lam({system; shadow solution}, shadow context):
       solve-helper-refinements(system, solution, context).bind(lam({shadow system; shadow solution}, shadow context):
-        solve-helper-fields(system, solution, context).bind(lam({shadow system; shadow solution}, shadow context):
-          solve-helper-examples(system, solution, context)
+        solve-helper-examples(system, solution, context).bind(lam({shadow system; shadow solution}, shadow context):
```
blerner: Why the swap in order, here?  Is there a semantic change as a result?

src/arr/compiler/type-check-structs.arr
```diff
@@ -756,7 +754,7 @@ fun solve-helper-refinements(system :: ConstraintSystem, solution :: ConstraintS
             end, {system; empty-tree-set})
 
             solve-helper-constraints(temp-system, constraint-solution(empty-tree-set, [string-dict: ]), context).bind(lam({shadow temp-system; temp-solution}, shadow context):
-              cases(ConstraintSolution) temp-solution:
+              cases(ConstraintSolution) temp-solution block:
```
blerner: Why the `block`?

src/arr/compiler/type-check-structs.arr
mkolosick: `solve-helper-fields` checks to make sure that all types required to have certain fields actually have those fields. So it has to come after all the types are solved for. Therefore it should come after `solve-helper-examples`. I think with how they actually are used it makes no difference to the order, but this is the more correct order.

src/arr/compiler/type-check-structs.arr
mkolosick: I just missed removing that from my debugging.

mkolosick:
You are correct on needing regression tests. Thank you for pointing that out.

mkolosick:
I have now added tests and fixed an error they picked up.

jpolitz:
I'm happy with this when @blerner closes his review.

blerner:
I'd specifically like to see a test or two on Srcloc's _plus, and on ConcatList's _plus, which are the current data-definitions that have existing custom _plus implementations -- one of them is monomorphic, and the other is polymorphic, and those should be some helpful regression tests too.

mkolosick:
`Srcloc` is not currently working properly in typed code so I have those tests commented out at the moment. I am working on this in issue #1169. I have added a monomorphic plus test and will uncomment the `Srcloc` tests as well once the other issue is handled.

jpolitz:
@blerner @mkolosick I don't quite understand the state of testing for this. Is this a demonstrable improvement independent of #1169, even if a test we'd like to write fails for unrelated reasons?

blerner:
I'm happier with it; I'll approve my changes so this can land, though I'd still like to see more tests eventually.

PR #1185, opened 2017-09-21
Type defaults

mkolosick:
This commit adds a way to write bottom types in the javascript type notation (this required a new bootstrap). This allows most of the global type definitions to now reside in `global.js`. Additional changes were made to add more builtin modules to type-defaults where needed (for builtin .arr modules) and to the `provides` object for .js modules.

There are more builtins to add types for, but I wanted to get the new bootstrap merged in sooner rather than later.


build/phase0/js/amd_loader.js
```diff
@@ -2,7 +2,9 @@ var define, requirejs;
 
 // NOTE(joe): this is defined for use in dependency bundles.  Even though it
 // isn't used in this file, it's an important definition
-var nodeRequire = require;
+if(typeof require === "function") {
+  var nodeRequire = require;
+}
```
jpolitz: I can imagine a few reasons for this – what made it be necessary?

mkolosick: Hmmm, I'm unsure. That came from running new-bootstrap, so it's not from any code that I wrote.

jpolitz: Oooh I misread. Got it. Going to have to be careful with reviewing this new phase0 "standard library" stuff.

PR #1186, opened 2017-09-21
Horizon -> master

jpolitz:
Main features:

- method-based tables
- spies
- fixed error messages using a reasonably efficient checkArgsInternal
- browserified standalones
- numerous bug- and code-organization-fixes to the type-checker


jpolitz:
FYI (especially @blerner) as you're reading this; some of these commits show up on master already, so this is reading to me as an overly expansive number of commits. For example, all the source-map stuff is deployed on master already, yet it's being listed here as well.

All of the commits from way back in 2016 are the ones to do with spies, which we merged into horizon this summer. So it's not like this is a merge of a branch that diverged last summer; they just carried their metadata from the spies branch.

PR #1197, opened 2017-09-25
Allow roughnums to have optional ellipsis

ds26gte:
This PR supersedes #1071.

Differences are:

1. The optional ellipsis is placed before the exponent for such roughnums as are in scientific notation.,

2. `pyret-toikenizer.js` adds many regexps that are recognizably bad numbers, involving illegal dots, or incorrect number of dots. Tests for these are included in `parse.js` and `test-parse-errors.arr`. Note that bad numbers of the type `2.` and `~2.` are not treated as bad numbers as they're currently treated as methods called on a non-object.


jpolitz:
Just a note, the build failure above is because of cross-repository pulls not working with the benchmarking infrastructure, not necessarily any failure in the PR. I can try and figure out how to only run benchmarks for stuff that's on an within-repo PR.

@ds26gte if you want to push this to a roughnum-ellipses-2 branch on the main pyret-lang repo and make the PR from there, it will more obviously run the tests cleanly.

Overall, I'm quite happy with this PR.

ds26gte:
I have pushed a branch `horizon+roughnum-ellipses-2` into `brownplt/pyret-lang`. I don't seem authorized to create a PR from there though.

PR #1199, opened 2017-09-26
Added roughnum tokens that allow optional trailing ellipses (before e…

jpolitz:
This PR supersedes #1071.

Differences are:

The optional ellipsis is placed before the exponent for such roughnums as are in scientific notation.

pyret-tokenizer.js adds many regexps that are recognizably bad numbers, involving illegal dots, or incorrect number of dots. Tests for these are included in parse.js and test-parse-errors.arr.

Note that bad numbers of the type 2. and ~2. are not treated as bad numbers as they're currently treated as methods called on a non-object.


jpolitz:
@ds26gte I don't think it was a permission issue – I think Github just won't let a PR be created if there's another PR of the same commit already existing (I just saw a link to the cross-repo PR when I looked at the branch).

jpolitz:
@blerner close or keep?

blerner:
I don't remember this terribly well, and I can't say as I've missed having this ability, either. Did we intend to permit `~2....` with four dots as legal syntax?

jpolitz:
Something like that. I think I'll close it. No one seems to be desperately missing it, the conflicts seem nontrivial, and it adds complexity.

PR #1208, opened 2017-09-29
list cleanup

sorawee:
1. a lot of methods are still not exploiting raw-list-*. Now they makes use of them.
2. rename join-str2 which was added experimentally and left accidentally to the real
   join-str since it's evidently much much more better in Firefox and have about the
   same performance in Chrome. Since it's undocumented, there's no need to retain join-str2


sorawee:
@jpolitz @blerner this now passes the test (I also add more tests for the new function I add)

sorawee:
This should target horizon instead of master. I don't know how to change the target branch. Will close this and open a new PR

blerner:
It's easy to change the target branch -- I've done so, so I'm reopening this PR.

blerner:
Now, there's a merge conflict with lists.arr, so check into that and clean it up, but it should be easier than recreating the PR.

sorawee:
@blerner as this PR deals with `join-str`, I think I will convert `join-str-last` to a similar style. For performance, is it a good idea to implement `raw-array-join-str-last` like this:

```
function raw_array_join_str_last(arr, sep, lastSep) {
  if (arr.length === 0) return '';
  var lastElement = arr.pop(); // use mutation here
  var result = arr.join(sep) + lastSep + lastElement;
  arr.push(lastElement); // also use mutation here
  return result;
}
```

A more functional way would be to use `slice` instead of mutating the array directly, but this will copy elements in the array unnecessarily.

sorawee:
Also, a question for @blerner and @jpolitz:

If I want to call `raw_array_map(tostring, arr)` in JS at a tail call position, do I need to use `safeCall`? My thought is no since there's nothing else to do in that stack frame, so the reconstructed stack could skip it safely. Is that accurate?

blerner:
I'm ok with the mutation for now, though leave a comment saying this is explicitly assuming an exclusive access to that array.  (Pyret is single-threaded, yes, but once it gets stopified, it's possible we might pause in the middle of this function, and the array will be in an inconsistent state.)

For scenarios like this, you would call `safeTail` instead.  It does nothing special, but it's a good syntactic marker for us that this is a tail call of a stack-safe function.

shriram:
Exactly. As I believe our Whalesong paper was the first to point out,
single-threadedness in JavaScript does not imply absence of race conditions
[see section 6.2 and figure 16 of
http://cs.brown.edu/~sk/Publications/Papers/Published/yk-whalesong-racket-browser/paper.pdf
].

sorawee:
@jpolitz and @blerner: please review.

@blerner and @shriram: I'm curious why I need to be aware of this race-condition. If using Stopify produces an incorrect output, isn't it because the program transformation is unsound rather than the input program is faulty? And as the Whalesong paper stated, the solution is to fix the program transformation (e.g., adding locks in the transformation).

blerner:
Robustness.  "Adding locks to the transformation" implies a whole-program transformation.  That's subtle, and we'd need to remember to uphold it in any future code we write.  This "please give me exclusive-write access to this array" assumption is *easy* for us to forget, and so we'd be baffled if something sporadically breaks because of it.

jpolitz:
> I'm curious why I need to be aware of this race-condition. If using Stopify produces an incorrect output, isn't it because the program transformation is unsound rather than the input program is faulty?

I agree with @sorawee here.

I would use safeTail only because it "counts" the frame so the runtime knows how many active JS frames there are, and can capture the stack at a reasonable limit before blowing the JS stack.

Issue #1175, opened 2017-09-17
bad "shadowing" error message


shriram:
[Running on CLI.]

If you have an identifier named `sets` in your program, you will get weird "shadowing" errors: eg,
```
stelvio ~/Desktop/CodePyret> pyret mst.arr 
1/1 modules compiled (mst.arr)      
Cleaning up and generating standalone...
The declaration of the identifier named `sets` at file:///Users/sk/Desktop/CodePyret/mst.arr:80:16-80:20 shadows the declaration of an identifier also named `sets` at file:///Users/sk/Desktop/CodePyret/mst.arr:1:0-109:70
There were compilation errors
```
This error is meaningless because `sets` is not "declared" anywhere in the program. The program doesn't even include/import the sets module! So a user has no way of making sense out of this.

blerner:
A little more context for triaging this would be helpful.  What does the program say?  Does the same error happen for other libraries (like lists, or option, or whatever)?

shriram:
Here's a complete file:
```
sets = 3
```
This produces
```
Cleaning up and generating standalone...
The declaration of the identifier named `sets` at file:///Users/sk/Desktop/CodePyret/sd.arr:1:0-1:4 shadows the declaration of an identifier also named `sets` at file:///Users/sk/Desktop/CodePyret/sd.arr:1:0-1:8
```

shriram:
Same happens if I replace `sets` with `lists` (with the error message updating, of course).

blerner:
This no longer occurs; closing.

Issue #1066, opened 2017-07-05
baffled by partial import of `lists`


shriram:
This has come up before, but with no satisfactory answer.

Say I'm trying to do some basic stats stuff. I get things from the `statistics` and `math` libraries with
```
include statistics
include math
```
and all is good. Then I realize I also need the function `distinct`, which is in the `lists` library.

I try `distinct` at the top-level. It's not bound.

Aha, so I `include lists`. I now get a ton of duplicate binding errors — but _not_ one for `distinct`. After these errors, I still can't use `distinct`.

So now I have to learn a new include form, `import`, and to say `L.distinct`, even though (say) `L.filter` is the same as `filter`, which is already present.

It would at least be better if _all_ of `lists` were already included, so then the documentation could just say so, and maybe even give a good error for anything already included previously (rather than the stream of 30 errors, which is so darn intimidating, and sends a "don't bother reading error output" signal). But this half-inclusion, which prevents subsequent inclusion, and breaks a pattern that works for other libraries, is really frustrating to impose on a user.

blerner:
Three comments: first, you were the one who had pushed for not including everything by default, so as not to pollute the global namespace. If you're changing your mind on that, ok, but let's record that for posterity.

Second, "fixing" this is as easy as making the remaining imports, in the relevant part of compile-structs.arr. The lists module already tries to export everything it defines.

Third, I wonder if there's another solution, where we split the lists module into a lists-base module that implicitly gets included everywhere, and the actual trove lists module that defines extra stuff and can be explicitly imported or included when desired.

jpolitz:
Yeah, this sucks.

1. It should probably not be an error to include a name that's already
included if it comes from the same module.
2. We know we need to make decisions about the global namespace. It is
unfortunately never a good time to do this. Leading up to the summer we are
busy with the features we need for summer. After trainings at the end of
the summer we can't change things. So this needs to get as high a
prioritization as actual features if it's going to happen.

shriram:
>  first, you were the one who had pushed for not including everything by default

Well, to be honest, this was at a time when I had high hopes that Justin would create a rough equivalent of `#lang` for Pyret. It has become clear that that will never happen (not the `#lang` part — maybe someone else will do that — but not from Justin). So I voiced some opinions under a different regimen that no longer applies and has no real hope of applying in the near future, which means some of those things need to be revisited.

[...]

Issue #1010, opened 2017-04-22
Namespacing and Shadowing


belph:
Currently (on `master`, anyway), the following program compiles:
```
data Foo:
  | bar(x)
end

Foo = bar(1)
```
Then, in the REPL, the following is possible:
```
> Foo
bar(1)
> a :: Foo = bar(2)
> a
bar(2)
```
In other words, `Foo` is now bound in the value and module namespaces at the same time. Shouldn't this be a shadowing error? In terms of pedagogy, it seems like we wouldn't want to explain that "this name means two disjoint things at the same time." My inclination is to update `resolve-scope` to maintain a separate 'unified' namespace to handle shadowing errors. Does this seem reasonable?

See Also:
- #571 (removal of `Foo` as a function defined by `data:`)
- #1009 (similar namespacing issue, with modules)

jpolitz:
I approve the idea of a unified namespace with shadowing errors across names and types.

I will note that we currently don't have syntax for `shadow`ing a type, so this is potentially more annoying than our current shadowing rules, which are easier to circumvent when you want to.

jpolitz:
@shriram / @blerner I'd like an opinion on if we're all on board with this change. I think it's an improvement, but it is (a) backwards incompatible and (b) does lead to a “complex things are impossible rather than hard” situation where if the value namespace takes a type name, there's no way at all to define a type with that name.

There's actually a nice module/namespacing branch that is minorly blocked on a decision here, so it would be good to have consensus.

I remembered this because of the discussion on https://github.com/brownplt/pyret-lang/issues/1221 today.

blerner:
In broad, I think I agree with having a unified namespace.  (Ironically, on the Racket mailing list right now, there's a discussion about the ability of having separated namespaces, to support individual and disjoint value and type bindings...)  I'm concerned about the implementation, about fixing up name-resolution, and I'm concerned about shadowing.  We'll now have four cases to consider:

* Existing value binding, shadowed value binding ==> allowed
* Existing value binding, shadowed type binding ==> currently syntactically impossible
* Existing type binding, shadowed value binding ==> ????
* Existing type binding, shadowed type binding ==> currently syntactically impossible

We should figure out what we actually want the behavior of all four of these cases to be, before we decide whether this is a uniformly good idea or not.  My hunch is we want to allow type/type shadowing, but prohibit value/type or type/value shadowing on semantic rather than syntactic grounds, but I'm not sure...

shriram:
[...]

PR #1222, opened 2017-10-12
Module ids on horizon

jpolitz:
Making a PR for this so we can have a place for discussion. This is a long-time-coming PR whose main effect can be summarized by the fact that this program:

```
import lists as L
L.mepty # intentional typo
```

becomes a static error instead of a runtime error. A large number of changes go into this in shipping around static information in various ways.

It currently fails because the command-line-parsing library defines `String` and `Number` as identifiers:

image

I think other than this issue this PR is a strict improvement.


jpolitz:
Just for context, this PR is relevant for several issues:

https://github.com/brownplt/pyret-lang/issues/1175
https://github.com/brownplt/pyret-lang/issues/1010
https://github.com/brownplt/pyret-lang/issues/678

And most important for:

https://github.com/brownplt/pyret-lang/issues/1066

This will make it so we can more easily aggregate together modules that may re-export the same names for the same functions.

src/arr/trove/ast.arr
```diff
@@ -464,12 +499,12 @@ sharing:
 end
 
 data DefinedValue:
-  | s-defined-value(name :: String, value :: Expr) with:
+  | s-defined-value(name :: String, provided-name :: Option<Name>, value :: Expr) with:
```
jpolitz: `Expr` should almost certainly be a `Name` here.

src/arr/trove/ast.arr
```diff
@@ -480,7 +515,7 @@ sharing:
   end
 end
 data DefinedType:
-  | s-defined-type(name :: String, typ :: Ann) with:
+  | s-defined-type(name :: String, provided-name :: Option<Name>, typ :: Ann) with:
```
jpolitz: `Ann` should almost certainly be a `Name` here.

src/arr/trove/ast.arr
```diff
@@ -490,6 +525,34 @@ sharing:
     self._match(visitor, lam(val): raise("No visitor field for " + self.label()) end)
   end
 end
+data DefinedModule:
+  | s-defined-module(name :: String, provided-name :: Option<Name>, id :: Name, mod :: String) with:
```
jpolitz: Is `mod` intended to be a URI?

jpolitz: In this and the preceding examples, it's not at all clear what the difference between `name` and `provided-name` are. Some description/example would be helpful near these types.

src/arr/trove/ast.arr
```diff
@@ -490,6 +525,34 @@ sharing:
...
+    method label(self): "s-defined-module" end,
+    method tosource(self):
+      PP.infix(INDENT, 1, str-colon, PP.str(self.name), PP.str(self.id), PP.str(self.mod))
```
jpolitz: This looks like it might be an arity bug.

src/arr/trove/ast.arr
```diff
@@ -945,6 +1015,19 @@ data Expr:
   | s-str(l :: Loc, s :: String) with:
     method label(self): "s-str" end,
     method tosource(self): PP.str(torepr(self.s)) end
+    # Base should be a name that refers to a module import, like import lists as L
+    # path is generalized to list because eventually this should support
+    # Lib.Mod.x kinds of references
+    #
+    # NOTE: The following invariants should hold for any s-module-dot 'smd':
+    #   is-s-mref-by-name(smd.base) => smd.path.length() >= 1
+    #   is-s-mref-by-uri(smd.base)  => smd.path.length() == 1
```
jpolitz: I think there's a different invariant here. Pre-scope-resolution, this could work either way.

Post-scope-resolution, everything should have a path of length 1, and a base that is a `mref-by-uri`.

This seems complicated! Why isn't this just an inductively collapsing case in resolve-scope that _always_ uses a `mref-by-uri`?

src/arr/trove/cmdline.arr
```diff
@@ -2,11 +2,11 @@
...
   file-name: file-name,
-  args: other-args,
+  args: args,
   Number: read-number,
-  Bool: read-bool,
-  String: read-string,
-  Custom: read-custom,
+  Bool: Bool,
+  String: String,
+  Custom: Custom,
```
jpolitz: This seems weird – is this _required_ by this commit or was it done for other reasons? It seems like this commit should be making this kind of renaming less necessary.

src/js/base/runtime.js
```diff
@@ -2802,8 +2802,10 @@ function (Namespace, jsnums, codePoint, util, exnStackParser, loader, seedrandom
     function getDotAnn(loc, name, ann, field) {
       checkString(name);
       checkString(field);
-      if(ann.hasOwnProperty(field)) {
+      if(field in ann) {
         return ann[field];
+      } else if (ann.dict && field in ann.dict) {
+        return ann.dict[field];
```
jpolitz: I hope this isn't permanent.

src/js/base/runtime.js
```diff
@@ -5057,12 +5059,20 @@ function (Namespace, jsnums, codePoint, util, exnStackParser, loader, seedrandom
         return m.jsmod;
       }
       else {
+        var modules;
+        if (thisRuntime.getField(m, "provide-plus-types").dict["modules"]) {
+          modules = thisRuntime.getField(thisRuntime.getField(m, "provide-plus-types"), "modules");
+        } else {
+          modules = thisRuntime.makeObject({});
```
jpolitz: We should make sure this can be removed after bootstrapping

src/js/base/runtime.js
```diff
@@ -5177,13 +5187,15 @@ function (Namespace, jsnums, codePoint, util, exnStackParser, loader, seedrandom
...
+    function makeModuleReturn(values, types, modules, internal) {
       return thisRuntime.makeObject({
         "defined-values": values,
         "defined-types": types,
+        "defined-modules": modules || {},
         "provide-plus-types": thisRuntime.makeObject({
           "values": thisRuntime.makeObject(values),
-          "types": types,
+          "types": thisRuntime.makeObject(types),
+          "modules": thisRuntime.makeObject(modules || {}), // FIXME: Should this be required?
```
jpolitz: Responding to comment, this does seem useful & required.

src/js/base/runtime.js
```diff
@@ -5314,7 +5326,12 @@ function (Namespace, jsnums, codePoint, util, exnStackParser, loader, seedrandom
         }
       });
       typeFields.forEach(function(tf) {
-        newns = newns.setType(tf, getField(moduleObj, "types")[tf]);
+        if(hasField(moduleObj, "defined-types")) {
```
jpolitz: This needs to be resolved before merging – when does this happen and why?

src/arr/compiler/compile-structs.arr
```diff
@@ -133,13 +153,14 @@ end
...
-        mods :: StringDict<Provides> # map from dependency key to info provided from module
+        mods :: StringDict<Provides>, # map from dependency key to info provided from module
+        uri-map :: StringDict<String> # map from URI to dependency keys
       )
 end
 
 # The strings in globals should be the appropriate dependency (e.g. in mods)
 data Globals:
-  | globals(values :: StringDict<URI>, types :: StringDict<URI>)
+  | globals(values :: StringDict<String>, types :: StringDict<String>, modules :: StringDict<String>)
```
jpolitz: Why aren't these URIs anymore? It seems like we may want to know that.

src/arr/compiler/ast-util.arr
```diff
@@ -1133,11 +1145,17 @@ fun get-named-provides(resolved :: CS.NameResolution, uri :: URI, compile-env ::
             exp = resolved.datatypes.get-value-now(d.d.key())
             data-typs.set-now(d.d.key(), data-expr-to-datatype(exp))
           end
+          module-typs = SD.make-mutable-string-dict()
+          for each(m from modules):
+            mod = resolved.modules.get-value-now(m.name.key())
+            module-typs.set-now(m.name.toname(), CS.m-resolved(mod))
```
jpolitz: This should probably be `CS.m-uri(mod.from-uri)` to avoid copying exports everywhere.

src/arr/compiler/ast-util.arr
```diff
@@ -1203,9 +1221,17 @@ fun canonicalize-value-export(ve :: CS.ValueExport, uri :: URI, tn):
...
-    other-uri = compile-env.mods.get-value(depkey).from-uri
-    other-uri == uri
+  compile-env.uri-map.get(uri)
+end
+
+fun transform-module-dict(d :: SD.StringDict<CS.ModuleExport>, uri :: URI, transformer, compile-env):
+  for fold(s from [SD.string-dict:], v from d.keys-list()):
+    me = d.get-value(v)
+    canonicalized = cases(CS.ModuleExport) me:
+      | m-uri(mod-key) => CS.m-resolved(find-mod(compile-env, mod-key))
```
jpolitz: Another dangerous place where we're expanding a module to its Provides.

src/arr/compiler/resolve-scope.arr
```diff
@@ -762,29 +1113,31 @@ fun resolve-names(p :: A.Program, initial-env :: C.CompileEnvironment):
...
-    for SD.each-key(name from initial.globals.types):
+    for each(name from initial.globals.types.keys-list()) block:
       mod-info = initial.mods.get-value(initial.globals.types.get-value(name))
-      acc.set-now(name, C.type-bind(C.bo-module(none, mod-info.from-uri), C.tb-type-let, names.s-type-global(name), none))
+      tbind = nb-type(C.type-bind(C.bo-module(none, mod-info.from-uri), C.tb-type-let, names.s-type-global(name), none))
+      acc.set-now(name, tbind)
+    end
+    for each(name from initial.globals.modules.keys-list()) block:
+      mod-info = initial.mods.get-value(initial.globals.modules.get-value(name))
+      mbind = nb-module(C.module-bind(C.bo-module(none, mod-info.from-uri), C.mb-resolved(mod-info), names.s-global(name)))
```
jpolitz: This line can't be a solution to a real problem.

jpolitz:
@blerner and I had a long look at this today. We are rejecting this PR as not at all the right approach.

We need to admit that a global URI->Provides mapping is the right way to do this, and stop faffing around with local dependency keys. We don't _really_ have separate compilable units right now (we regenerate the whole DAG of cached information when we recompile), and quasi-preserving that is getting in the way.

We should coalesce a lot of the datatypes here into:

- A map from URI -> Provides
- A function that takes a URI and a dependency (probably implemented as a map) and produces a Provides

These two functions can be used to resolve all names to the URI they are provided from directly when they are encountered. It removes the idea of having a _path_ that goes through multiple modules (we could store the residue of the path for error information if we really want, but it shouldn't be part of the datatype. Since paths aren't necessarily unique, this may not be useful anyway).

One issue we saw in the PR was also that provides could be embedded in the serialized representation of modules, effectively duplicating them when they were re-exported. This should be avoided because it could lead to an explosion of re-exports.


We think, overall, this change needs to be driven by the design of `provide` syntax, and the kinds of operations we want to be expressible on names across modules. There are issues of the original identity of a definition (including the "same" thing from multiple places shouldn't cause a shadowing error), and re-exporting definitions cleanly in syntax to package things together.

This PR doesn't give a clear answer for why or how those things should work, and since they are design goals for the module system, we should address them head-on before making large changes.

PR #1232, opened 2017-10-22
fix #1230

### LINKED ISSUE #1230: I've confused with fibonacci function using tail recursive.
Hi,

I'm a Racket user, and learning the Pyret lang.

I've tried Pyret online and compiled by phase0, but I got wrong answer of fibonacci function using tail recusion:

```
#lang pyret

fun iter(a :: Number, b :: Number, count :: Number) -> Number:
  if count == 0:
    b
  else:
    iter((a + b), a, (count - 1))
  end
end


fun fib(n :: Number) -> Number:
  iter(1, 0, n)
where:
  fib(10) == 55
end

print(fib(10))
```

but its result is `512`.

How could I fix my program?

sorawee:
Fix the bug by assigning the values to a temporary variable before assigning it again to the real variable, avoiding the dependency issue.

Note that the temporary variable `compiler-name(arg.toname())` is in scope at the beginning of the function, but is effectively useless (no more reference to the variable) at the point where we assign the value to the variable. That's why I decide to use it for as a temporary variable. But if anyone objects, I can also change it to other names.

### DIFF
--- a/src/arr/compiler/anf-loop-compiler.arr
+++ b/src/arr/compiler/anf-loop-compiler.arr
@@ -896,6 +896,37 @@ fun compile-split-method-app(l, compiler, opt-dest, obj, methname, args, opt-bod
   end
 end
 
+fun is-id-occurs(target :: A.Name, e :: J.JExpr) block:
+  var found = false
+  e.visit(J.default-map-visitor.{
+    method j-id(self, name :: A.Name):
+      when name == target:
+        found := true
+      end
+    end
+  })
+  found
+end
+
+type Asgn = {A.Name; J.JExpr}
+
+fun get-assignments(lst :: List<Asgn>, limit :: Number) -> {List<Asgn>; List<Asgn>}:
+  if limit == 0:
+    {empty; lst}
+  else:
+    cases (List) lst:
+      | empty => {empty; empty}
+      | link(arg, rest) =>
+        if for any({_; e} from rest): is-id-occurs(arg.{0}, e) end:
+          get-assignments(rest + [list: arg], limit - 1)
+        else:
+          {good; bad} = get-assignments(rest, rest.length())
+          {link(arg, good); bad}
+        end
+    end
+  end
+end
+
 fun compile-split-app(l, compiler, opt-dest, f, args, opt-body, app-info, is-definitely-fn) block:
   ans = compiler.cur-ans
   step = compiler.cur-step
@@ -907,6 +938,13 @@ fun compile-split-app(l, compiler, opt-dest, f, args, opt-body, app-info, is-def
      compiler.allow-tco and
      compiler.options.proper-tail-calls and
      (compiled-args.length() == compiler.args.length()): # if it's an arity mismatch, use non-TCO to handle the error
+    args-list = for map2(formal from compiler.args, actual from compiled-args.to-list()):
+      {formal; actual}
+    end
+    {no-dep-lst; dep-lst} = get-assignments(args-list, args-list.length())
+    shadow dep-lst = for map({formal; actual} from dep-lst):
+      {formal; fresh-id(compiler-name('tmp_asgn')); actual}
+    end
     c-block(
       j-block(
         [clist:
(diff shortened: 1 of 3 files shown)


sorawee:
Does this also need bootstrap?

blerner:
A few thoughts:
1. In terms of code complexity, a simpler change would be to synthesize an activation record here and then reuse the logic for restoring activation records.  But that might negate some of the benefits of tail calls.  (You'd also wind up with unpaired calls to `traceEnter` and `traceExit`, unless you also passed along `elidedFrames`, so that might be more complicated than it's worth.)
2. I _think_ you could avoid the need for temporary variables by checking whether any argument in `compiler.args` appears at a position with higher index in `compiled-args.to-list()` -- the former being the variables being written, and the latter being the variables being read, so this condition would check whether you overwrite the variables before they've been used.  (For instance, the reason your suggested fix in #1230 "worked" is because you changed the dependency ordering among the arguments, and so it happened to work correctly there.)  I think this check might be worth doing, since several key small functions are tail recursive, and I'd like to try to keep their stack frames small...
3. I don't think this requires a bootstrap to fix student-compiled code.  It would need a bootstrap to fix the pyret compiler itself, if this bug bit us anywhere.

jpolitz:
Merging this as-is for now, we should make an issue for @blerner 's 2nd suggestion above, which is quite good.

src/arr/compiler/anf-loop-compiler.arr
```diff
@@ -919,15 +919,18 @@ fun compile-split-app(l, compiler, opt-dest, f, args, opt-body, app-info, is-def
               j-expr(j-dot-assign(RUNTIME, "EXN_STACKHEIGHT", j-num(0))),
               j-expr(j-assign(ans, rt-method("makeCont", cl-empty)))]))] +
         CL.map_list2(
-          lam(compiled-arg, arg): j-expr(j-assign(arg, compiled-arg)) end,
+          lam(compiled-arg, arg): j-expr(j-assign(compiler-name(arg.toname()), compiled-arg)) end,
```
jpolitz: Should this be `j-var` of this new name? I worry that this will make a bunch of global variables.

jpolitz: Also, do we have a guarantee that these names won't conflict with other existing names in scope (or in future modifications to the compiler?) These seem a little brittle if they just use the arguments names with a consistent prefix appended, which is what `compiler-name` does, and `toname` just gives back a human-looking name.

blerner:
Yes, we need a `j-var` here. As for brittle renaming, compiler-name ensures that we don't conflict with user written names, but gives no guarantees about our own. If we come up with a naming convention, eg `$...$foo` where the dots can be any short string we want, then we can easily avoid accidental name collisions.

jpolitz:
`anf-loop-compiler` has the function `fresh-id` defined (I believe) expressly for this purpose. Shouldn't this use that for each argument's temporary?

sorawee:
We don't need `j-var` because these variables are declared already. More precisely, the function's formal arguments are actually:

```
function ($a, $b, $count) {
  if ($a is activation record) {
    a = $a[0];
    b = $a[1];
    count = $a[2];
  } else {
    var a = $a;
    var b = $b;
    var count = $count;
  }
  ...
  // $a, $b, $count never appears again
}
```

I simply reuse the name `$a`, `$b`, `$count`.

sorawee:
> I'd like to try to keep their stack frames small...

Can you tell me why my change would not make the stack frames small?

blerner:
Assigning to the formal parameters of a function is one of the behaviors that seem to pessimize jit performance. That's why there *are* the copies of formal parameters at all (read the comments above that snippet in the compiler code). So don't reuse those parameters directly. Which means you need another set of variables instead, and that will bloat the stack frame size.

sorawee:
So the only objection here is about performance? Can we get this merged for now then? I'm not really free until next week, but students are using Pyret right now.

jpolitz:
@blerner are you satisfied that the current use of variables is _correct_? I'm totally weirded out by the use of `toname` still.

blerner:
Am I convinced they're correct?  Not really.  The better thing to do here would be to use `compiler.args.map(formal-shadow-name)`, and then change line 528 to be `args: args.map(_.id)` without the call to `.map(js-id-of)`.  `formal-shadow-name` is how we derive the actual-arguments-we-stuck-in-the-function-header from the-names-we-use-in-the-function-body, so it's tautologically correct.  I don't want to be doing this assignment at all, though, but I'll accept it for now.

blerner:
Could we dodge this question entirely by using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Swapping_variables instead, and let the js engine take care of the dependencies?

jpolitz:
Given the reports we're getting of students using older Safari on iPads, my confidence that this will work everywhere we need it to is low. And since it's in compiled output, it's not easy to apply Babel

sorawee:
Feel free to take this over from me. I'm really not free this week.

sorawee:
@jpolitz @blerner  The latest commit naively implements topological sort which solves the issue. It's not big-oh efficient, but the arg list should be short and this should be faster practically.

sorawee:
I spent 3 hours thinking how to solve an NP-complete problem (https://en.wikipedia.org/wiki/Feedback_vertex_set). Sigh.

blerner:
Regression test we ought to add:
```
fun oops(a, b):
  if a == 1: b
  else:
    oops(1, a)
  end
end
fun working(a, b):
  if a == 1: b
  else:
    working(1, a) + 0
  end
end

examples:
  oops(4, 5) is working(4, 5)
end
```

src/arr/compiler/anf-loop-compiler.arr
sorawee: The newer version generates a new id using `fresh-id` and now uses `j-var`.

jpolitz:
@blerner – I think I'm happy with this at this point. At the very least, it fixes a serious issue and doesn't obviously hurt anything. Any objections to merging?

sorawee:
Out of curiosity, how is this traditionally implemented? I devised the above algorithm by myself, so I don't know how good or bad it is compared to the traditional method.

jpolitz:
I don't know, actually. My guess is that the tail call would be converted to just nodes + edges in a CFG representation, and it would be handled by whatever register allocation strategy was in use. That could be graph coloring or some other heuristic.

shriram:
Yep. It gets turned into a "loop" and then register allocation treats it
like any other loop.

blerner:
I think it's ok for now, yes.  I want to understand it better, and I'd like to eliminate some inefficient allocations of lists that we probably could improve...but since most arglists are short, we're talking small constants here, so it doesn't matter much.

PR #1242, opened 2017-11-03
Fixes #1233

### LINKED ISSUE #1233: type checker has wrong type for `raises`
The type-checker consistently uses a function's return type as the expected type on the RHS of a `raises`, but that should instead be a string. Eg:

```
fun func() -> Number:
  0
where:
  func() raises ""
end
```

says

    The type constraint Number was incompatible with the type constraint String

Note that the following program does _not_ produce a type error:

```
fun f(x) -> Number:
  1
where:
  f(0) is 1
  f(1) raises 'sorry'
end
```

It'd be great if we could fix this soon, since it's hurting students on the current assignment.

[Credit: Zachary Espiritu.]

mkolosick:
Adds type-checker support for almost all of the test comparisons. Doesn't yet handle `s-op-raises-satisfies` or `s-op-raises-violates`.

### DIFF
--- a/src/arr/compiler/type-check.arr
+++ b/src/arr/compiler/type-check.arr
@@ -11,6 +11,7 @@ import file("ast-util.arr") as AU
 import file("type-structs.arr") as TS
 import file("type-check-structs.arr") as TCS
 import file("compile-structs.arr") as C
+import file("type-defaults.arr") as TD
 
 type Type = TS.Type
 type TypeMembers = TS.TypeMembers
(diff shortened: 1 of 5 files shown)


blerner:
Looks good, but what are the remaining difficulties with raises-satisfies and raises-violates?  Easy enough to implement now, or will that be an entirely separate PR?

mkolosick:
Actually I can do those now in this pull request. Those two depend on types for error.arr but I appear to have done that already. So I'll add those now.

PR #1243, opened 2017-11-03
Fixes #1240

### LINKED ISSUE #1240: type checker internal error on `!`
This program works fine in the untyped version but effectively gives an internal error in the typed version:
```
data Box<T>:
  | box(ref v :: T)
end

n1 = box(1)
n1!v
```
This is problematic because it means we can't use the type checker on any of the remaining assignments in 019.

(I'm curious how this got past the test suite, because it's such a rudimentary program.)

Note that there are two problems:

- the fact that this produces an error at all
- the output of which says "One or more internal errors prevented us from showing the best error message possible. Please report this as a bug."

mkolosick:
This fixes a typo in reference type substitution.


blerner:
LGTM.  (Once again the refrain, if only we could run the type checker on the compiler, we'd have caught this!)

PR #1247, opened 2017-11-08
Emacs mode: Fix shorthand lambda/tuple disambiguation and 'constant fields'

belph:
@blerner: I saw that `horizon` is currently trailing `master`, so I opened this PR for this branch. If you want me to reopen on `horizon`, just let me know.

This PR also fixes cases like this:
```
data A:
  | a with:
    b : 2
end
```


blerner:
This seems not to be quite fixed yet: the following code breaks in emacs but doesn't break in CPO:
```
fun foo():
  constructor-loc =
    destructured-definition.and-then({(v) block:
          v.{1}})
        .or-else(self.constructor-loc)
      bad
    end
```
Things go wrong starting at the `v.{1}` line

belph:
Got it; I'll take a look

blerner:
I think this works; I'm willing to merge.

PR #1250, opened 2017-11-10
Tuple fix

mkolosick:


jpolitz:
Regression test?

mkolosick:
Can someone take a look at `tests/type-check/main.arr`? How should I actually run the compiled code?

jpolitz:
https://github.com/brownplt/pyret-lang/blob/master/tests/pyret/test-compile-helper.arr

Has `run-to-result`, which could be parameterized with `type-check: true`

Alternatively, we could add a new file, called `type-run.arr`, that is
alongside regression.arr and main.arr, which runs a set of tests that
compile and execute with the type checker on. That would just mean copying
and tweaking a make rule.

I'm happy with either solution; the latter might be easier to get going,
especially because I don't know that all the type-check tests are intended
to be meaningfully _runnable_ even if they are _checkable_.

mkolosick:
I'm getting a `Key builtin(global) not found file:///Users/mkolosick/dev/pyret-lang/src/arr/compiler/resolve-scope.arr:749:17-749:79` when I try to use `run-to-result`. I'm very confused about all of the different ways to run the compiler.

mkolosick:
Solved that issue. But I still need help. The biggest thing I can't figure out is how to run the program and let it use imports from the same folder that it is in (see `tests/type-check/good/local-import.arr` and `tests/type-check/good/export.arr`.

tests/type-check/main.arr
```diff
@@ -30,6 +31,13 @@ fun string-to-locator(name, str :: String):
   }
 end
 
+fun run-typed-file(base-path, filename) block:
+  file = FL.open-input-file(base-path + filename)
+  file-contents = FL.read-file(file)
+  FL.close-input-file(file)
+  TCH.run-to-result-typed(string-to-locator(filename, file-contents), base-path)
```
blerner: This should be `base-path + filename`, right?

mkolosick: Are you referring to the `filename` in the `string-to-locator` call?

tests/type-check/main.arr
```diff
@@ -58,38 +66,38 @@ end
...
 check "These should all be good programs":
   base = "./tests/type-check/good/"
-  good-progs = FL.list-files(base)
+  # good-progs = FL.list-files(base)
+  good-progs = [list: "_plus.arr"]
   for each(prog from good-progs):
     when is-arr-file(prog) block:
-      filename = base + prog
-      result = compile-str(filename)
+      result = run-typed-file(base, prog)
```
blerner: You're not using `compile-file` above now, you're using `run-typed-file` which seems to be unrelated...

mkolosick: The issue is that `compile-file` doesn't actually run the compiled code. `run-typed-file` is my attempt to actually run the good type checker tests.

tests/type-check/main.arr
blerner: yes

mkolosick: That shouldn't (and doesn't) make a difference. The first argument is just used for the `uri()` and `name()` in the locator.

jpolitz:
I see the tests passing – does this work for running the tests that also run files, @mkolosick ?

mkolosick:
No it doesn't as the tests should actually be failing right now.

blerner:
I'm digging into this right now.  It *is* in fact running the programs, and the problem seems to be in how `run-to-result-typed` gets the actual result of the program back out.  (It seems to be stopping at some `Result` value, rather than the value contained inside it.)

blerner:
I've got it; fix coming in a minute

blerner:
Note that with the d9c59e9 commit, the *expected* output is for `_plus.arr` to fail (since it was deliberately broken to check whether output was happening), but I'm surprised to see that `good-reactor.arr` is failing also.  @mkolosick over to you now, this should get you unstuck,

blerner:
Ah.  @jpolitz the good-reactor file is failing because load-lib's notion of `postLoadHooks` is impoverished compared to handalone's notion.  It's missing stuff for `builtin://data-source` and `builtin://reactors`.  I really want to refactor this into one centralized place, and it feels like `runtime-lib.js` is the right place for it...but then handalone can't use it because it can't load Pyret modules yet.  For now, I can copy the missing bits from handalone into load-lib, but this is getting ugly :(

My latest push should make this PR pass now, so if it does, it's fair to merge.

PR #1254, opened 2017-11-12
Refactor the postLoadHook creation so that it's only done once...

blerner:
This de-duplicates the code in handalone.js and load-lib.js so that the common postLoadHooks are written uniquely.  Once this merges, a similar change can be made to CPO.

### DIFF
--- a/src/js/base/post-load-hooks.js
+++ b/src/js/base/post-load-hooks.js
@@ -1,5 +1,5 @@
 define("pyret-base/js/post-load-hooks", function() {
-  function makeDefaultPostLoadHooks(runtime, main) {
+  function makeDefaultPostLoadHooks(runtime, hookOptions) {
     return {
       "builtin://srcloc": function(srcloc) {
         runtime.srcloc = runtime.getField(runtime.getField(srcloc, "provide-plus-types"), "values");
@@ -123,9 +123,9 @@ define("pyret-base/js/post-load-hooks", function() {
         runtime.setParam("makeReactor", runtime.getField(r, "make-reactor").app);
       },
       "builtin://checker": function(checker) {
-        checker = runtime.getField(runtime.getField(checker, "provide-plus-types"), "values");
-        // NOTE(joe): This is the place to add checkAll
-        var currentChecker = runtime.getField(checker, "make-check-context").app(runtime.makeString(main), true);
+        var checker = runtime.getField(runtime.getField(checker, "provide-plus-types"), "values");
+        var currentChecker = runtime.getField(checker, "make-check-context").app(runtime.makeString(hookOptions.main),
+                                                                                 hookOptions.checkAll);
         runtime.setParam("current-checker", currentChecker);
       }
     };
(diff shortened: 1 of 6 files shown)


src/js/base/post-load-hooks.js
```diff
@@ -0,0 +1,136 @@
...
+          runtime.getField(ds, "is-CellContent").app, "CellContent");
+      },
+      "builtin://reactors": function(reactor) {
+        var r = runtime.getField(runtime.getField(reactor, "provide-plus-types"), "values");
+        runtime.setParam("makeReactor", runtime.getField(r, "make-reactor").app);
+      },
+      "builtin://checker": function(checker) {
+        checker = runtime.getField(runtime.getField(checker, "provide-plus-types"), "values");
+        // NOTE(joe): This is the place to add checkAll
+        var currentChecker = runtime.getField(checker, "make-check-context").app(runtime.makeString(main), true);
```
jpolitz: The last parameter is a variable called `checkAll` in `load-lib`, and is unconditionally `true` here

src/js/base/post-load-hooks.js
```diff
@@ -0,0 +1,136 @@
...
+        };
+        runtime["checkCellContent"] = runtime.makeCheckType(
+          runtime.getField(ds, "is-CellContent").app, "CellContent");
+      },
+      "builtin://reactors": function(reactor) {
+        var r = runtime.getField(runtime.getField(reactor, "provide-plus-types"), "values");
+        runtime.setParam("makeReactor", runtime.getField(r, "make-reactor").app);
+      },
+      "builtin://checker": function(checker) {
+        checker = runtime.getField(runtime.getField(checker, "provide-plus-types"), "values");
```
jpolitz: Does this `checker` variable have a declaration? It looks like this is a variable _assignment_ in handalone and a variable _declaration_ in load-lib.

src/js/base/post-load-hooks.js
```diff
@@ -0,0 +1,136 @@
+define("pyret-base/js/post-load-hooks", function() {
+  function makeDefaultPostLoadHooks(runtime, main) {
```
jpolitz: In general, I bet this second argument is going to need to be a `runtimeOptions` datatype of some sort.

blerner: Yeah I had thought of that, but didn't notice the other points of variation, so didn't think it was warranted...

src/js/base/post-load-hooks.js
blerner: I'm not sure, because I can't tell whether it was even _used_ as a variable or not in the original code!  I think it was a copypasta error that was meant to be a variable declaration.

src/js/base/post-load-hooks.js
blerner: Fixed, thanks.

src/js/base/post-load-hooks.js
jpolitz: I see no `var` declaration for it. Hooray for refactoring fixing bugs!

Issue #835, opened 2016-08-22
unexpected output when updating field name


shriram:
I was not expecting this last line of REPL output (this is on pyret-dev):
```
> data Foo: f(a, b) end
> f(2, 3)
f(23)
> f(2, 3).{a: 4}
{a4b3$fieldNames[raw-array: "a""b"]_match<method:f>}
```

shriram:
Note: the actual output looks a bit different, but copying-and-pasting REPL output still does not work (per a previously filed report, and that's a serious one).

jpolitz:
See https://github.com/brownplt/pyret-lang/issues/307.  This should be an error, but has a more complicated story long-term.

jswrenn:
This is a surprising output:
```
> data Foo: f(a, b) end
> f(2,3).{c:4}
{c: 4, a: 2, b: 3, $fieldNames: [raw-array: "a", "b"], _match: <method:f>}
```

PR #1255, opened 2017-11-14
Adding update and extend type support for data types

mkolosick:
Allows extend and update syntax to be used on data types.
Errors if the fields do not occur in the data type, e.g.
```
data Foo:
  | bar(a :: Number)
end

bar(1).{b: 3}
```

### DIFF
--- a/tests/type-check/bad/data-as-object-extend-missing.arr
+++ b/tests/type-check/bad/data-as-object-extend-missing.arr
@@ -1,0 +1,5 @@
+data Foo:
+  | bar(a :: Number, ref b :: String)
+end
+
+bar(1, "a").{a: "a"}
--- a/src/arr/compiler/type-check.arr
+++ b/src/arr/compiler/type-check.arr
@@ -1858,8 +1858,8 @@ end
 
 fun synthesis-extend(update-loc :: Loc, obj :: Expr, obj-type :: Type, fields :: List<A.Member>, context :: Context) -> TypingResult:
   collect-members(fields, false, context).typing-bind(lam(new-members, shadow context):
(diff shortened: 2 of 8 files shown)


tests/type-check/bad/data-as-object-extend-missing.arr
```diff
@@ -0,0 +1,5 @@
+data Foo:
+  | bar(a :: Number, ref b :: String)
+end
+
+bar(1, "a").{a: "a"}
```
blerner: Are these test cases named correctly?  This doesn't seem to be a missing field, it's a wrong type.  In general, could you add a comment to each bad test, above the failing expression, explaining what the error should be?

mkolosick: Oops I flipped two of the test files for both update and extend.

blerner: This looks almost right...but you forgot to fix this specific file.  After that, I'm good to merge this.

mkolosick: I have fixed it now.

blerner:
Thanks -- just to confirm: it is a design goal of the type checker to disallow adding fields to data values, right?

If so, I'm good to merge...

mkolosick:
That is correct.

jpolitz:
Wow, we got static support for this feature before dynamic support is fully baked.

Currently, update of _existing_ fields does weird things dynamically, as does adding new fields.

https://github.com/brownplt/pyret-lang/issues/835 is currently an error.

See:

https://github.com/orgs/brownplt/teams/pyret/discussions/8

I think we need a decision on the desired dynamic semantics of this before we can start blessing programs as typed.

PR #1266, opened 2017-11-25
Adding simple function-entry/function-exit profiling information

blerner:
* Using the `-profile` command-line argument will instruct the compiler to add profiling information to the main module.  (This is easily generalized: the `should-profile` method added to `build-program` in cli-module-loader.arr could easily return true always, or for some subset of modules, etc.)
* The profiling information consists of a triple, [srcloc, boolean, timestamp]:
  - The srcloc is in raw form (i.e. a 7-tuple JS array).
  - Boolean true means function entry; false means function exit.
  - The timestamp is the highest-resolution we can get per platform, meaning process.hrtime, falling back to performance.now, with Date.now as last resort.  (The process.hrtime output is a pair of integers meaning seconds and nanoseconds; the performance.now output is a double meaning milliseconds, and Date.now is an integer meaning milliseconds.)
* Any profiling output is printed to process.stderr (at least for now; this could easily be changed in handalone.js), one line per profile entry, in JSON format.
* This commit doesn't add any interpretation of the output (in particular, it does no runtime processing of the hrtime output, so that no time is wasted during actual running of the program -- save that for postprocessing).  Separate scripts could easily be written to analyze the data and produce interesting statistics...

### DIFF
--- a/src/js/base/post-load-hooks.js
+++ b/src/js/base/post-load-hooks.js
@@ -75,10 +75,21 @@ define("pyret-base/js/post-load-hooks", function() {
           empty_only : runtime.getField(ds, "empty-only")
         };
 
-        runtime["makeCStr"] = runtime.getField(ds, "c-str").app;
-        runtime["makeCNum"] = runtime.getField(ds, "c-num").app;
-        runtime["makeCBool"] = runtime.getField(ds, "c-bool").app;
-        runtime["makeCCustom"] = runtime.getField(ds, "c-custom").app;
+        function reCache(src, constructorName, target, cachedName) {
+          return function() {
+            try {
+              var constr = runtime.getField(src, constructorName);
+              return constr.app.apply(constr, arguments);
+            } finally {
+              // after jitting, this will be faster than the call above
+              target[cachedName] = runtime.getField(src, constructorName).app;
+            }
+          };
+        }
+        runtime["makeCStr"] = reCache(ds, "c-str", runtime, "makeCStr");
+        runtime["makeCNum"] = reCache(ds, "c-num", runtime, "makeCNum");
+        runtime["makeCBool"] = reCache(ds, "c-bool", runtime, "makeCBool");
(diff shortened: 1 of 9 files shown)


src/js/base/post-load-hooks.js
```diff
@@ -75,10 +75,21 @@ define("pyret-base/js/post-load-hooks", function() {
           empty_only : runtime.getField(ds, "empty-only")
         };
 
-        runtime["makeCStr"] = runtime.getField(ds, "c-str").app;
-        runtime["makeCNum"] = runtime.getField(ds, "c-num").app;
-        runtime["makeCBool"] = runtime.getField(ds, "c-bool").app;
-        runtime["makeCCustom"] = runtime.getField(ds, "c-custom").app;
+        function reCache(src, constructorName, target, cachedName) {
```
jpolitz: This seems unrelated to the point of the PR, unless I'm missing something. But this code is already on horizon, so maybe this is irrelevant.

blerner:
Not sure I follow what you mean about reCache being added twice. And yes,
strictly speaking it's not connected to the point of this pr, but I added
it either as a cherry-picked commit from elsewhere, or added it fresh
because it seemed horrible that a simple performance-measuring pr should
have such a noticeable performance regression. (It was a bad interaction
with the tables implementation and data values that hadn't been evaluated
even once yet... And wasn't something I wanted to let linger:
reconstructing later why the performance was bad would be tedious and
redundant.)

jpolitz:
> Not sure I follow what you mean about reCache being added twice.

The same code is on both `master` and `horizon` right now that is added in this commit.

I don't think it would be a _regression_ if this PR didn't add that code, because that's not how commits work.

In part, I'm pointing this out because, reviewing past PRs, one thing that _really_ holds them up is when unrelated changes are lumped together into one PR. This is about adding profiling code, and there's a completely unrelated change in it (sure, it improves performance, but that's a tenuous connection at best).

I'm going to merge this, because I think it's fine in this case, but it seems really odd to have the same code added from multiple different PRs, and it makes the review much less clear.

PR #1274, opened 2017-11-26
Generate visitors automatically

sorawee:
This PR separates AST visitors from `ast.arr` to another file named `ast-visitors.arr`. This change is done by @jpolitz in the v-hull branch: https://github.com/brownplt/pyret-lang/blob/v-hull/src/arr/trove/ast-visitors.arr

However, the file `ast-visitors.arr` can also be automatically generated by reading `ast.arr` and type annotations on each AST node constructors. This is what this PR does. Concretely:

1. It collects all variants from data definitions that contain `visit` method.
2. For each member of a variant, if the type annotation indicates that it can be recursively visited, make it so in the visitors.

To generate the visitor file, run

```
$ node build/phaseA/pyret.jarr \
    --build-runnable src/arr/compiler/ast-transformer.arr \
    --outfile ast-transformer.jarr \
    --builtin-js-dir src/js/trove/ \
    --builtin-arr-dir src/arr/trove \
    --require-config src/scripts/standalone-configA.json
$  node ast-transformer.jarr \
      -default-map-visitor \
      -default-iter-visitor \
      -dummy-loc-visitor \
      --in-file src/arr/trove/ast.arr \
      --out-file src/arr/trove/ast-visitors.arr \
      --builtin-import ast \
      -strip-annotation
```

This `ast-transformer` can potentially by used with `js-ast` and `anf-ast` too.

The benefit is that it will keep AST visitors in sync with AST better. In fact, this PR discovers some bugs (at least 10). 

1. https://github.com/brownplt/pyret-lang/blob/horizon/src/arr/trove/ast.arr#L3162 causes `s-lam`'s name to be empty string, but this should not be its job. Unfortunately, this bug is essential for the tests to pass, so I am passing a "patch" parameter to make the generated visitor slightly wrong but pass the tests. I will need help on this for the best action to do.
2. https://github.com/brownplt/pyret-lang/blob/horizon/src/arr/trove/ast.arr#L3119 has arity mismatch
3. https://github.com/brownplt/pyret-lang/blob/horizon/src/arr/trove/ast.arr#L1941 doesn't recursively visit `params`
4. https://github.com/brownplt/pyret-lang/blob/horizon/src/arr/trove/ast.arr#L3452 should not recur on `name` since it's a string
5. https://github.com/brownplt/pyret-lang/blob/horizon/src/arr/trove/ast.arr#L3448 forgot to recur on the `obj`
6. https://github.com/brownplt/pyret-lang/blob/horizon/src/arr/trove/ast.arr#L2871 forgot to recur on the `id`
7. Missing visitor branch for `a-checked`
9. Missing visitor branch for `s-tuple-get` and friends for dummy-loc-visitor
8. Missing all visitors from `VariantMemberType`
10. https://github.com/brownplt/pyret-lang/blob/horizon/src/arr/trove/ast.arr#L3231 forgot loc strip

To review for the content of the AST visitor change, here's the diff outputted by Python's ndiff algorithm: https://gist.github.com/sorawee/db06472e9b227c690e51857df8aa1ed1


sorawee:
@jpolitz and @blerner: should `ProvidedValue`, `ProvidedAlias`, and `ProvidedDatatype` have the visit method?

sorawee:
Please read the following as if you have not seen my PR:

> I don't like the idea of checking in to our repository a file which is now auto-generated, because it will inevitably bitrot and fall out of sync, the way that our racket/scribble documentation has fallen out of sync. This should either be generated as a pre-build step, or it should not be an autogenerated file.

> If we autogenerate this, then we should also rethink and improve how we autogenerate the _match method on every single data definition. I'd rather consider building a JS library that takes a data value and a visitor, and does the double-dispatch in one step and one library, rather than add bloat to every data definition.

I proposed the following:

We write a file named `ast-core.arr` in `compiler`. We then autogenerate `ast.arr` with `visit`, `_match`, and visitors in `ast.arr` in trove without checking them in. Other stuff in the current `ast.arr` will be moved to `ast-util.arr` in trove. With this way, we don't need to touch JS, and the bloat part is in autogenerated file and not something we write.

> I do not like the use of and-then and or-else in the option cases of the visitor: they add an extra lambda allocation to visiting every option in an AST (for and-then) and evaluate their argument needlessly (for or-else), which is overhead that we don't need. That's why the original visitors had an option method...

Sure. Is it then a good idea to add `method list` in every visitor too?

> What is this only-serializable stuff, and what are you asking for when you ask for a serialize-visit?

Let me ask this question instead: can I add `visit` method to every data definition that is a part of AST node (except primitives and Loc). In particular, add `visit` method to `ProvidedValue`, `ProvidedAlias`, etc. If so, I won't need only-serializable stuff anymore.

I simply want to maintain the old visitor while supporting the new desugar system which is a work @justinpombrio  and I are working on, and that work needs a visitor on every AST node (including `ProvidedValue`, etc.). `only-serializable` is simply a flag whether the normal visitors should recursively visit it or not. But if I can add the `visit` method to those anyway, we won't need these flags anymore.

> I really don't like the idea of an appendix file.

Right, so please tell me what to do with the issue I asked above:

> https://github.com/brownplt/pyret-lang/blob/horizon/src/arr/trove/ast.arr#L3162 causes s-lam's name to be empty string, but this should not be its job. Unfortunately, this bug is essential for the tests to pass, so I am passing a "patch" parameter to make the generated visitor slightly wrong but pass the tests. I will need help on this for the best action to do.

sorawee:
My another proposal on generated file. 

I think generated files should be checked because they are files in trove. I.e., people will _import_ it, and it is important that they can see what is the content inside the file. What we can do to avoid the "fall out of sync" is simply changing the build script to diff the autogenerated file with the checked file, and if the diff is not empty, fail the build process.

sorawee:
> I really don't like the idea of an appendix file -- that feels like a kludgy hack, and it makes the dummy-loc-visitor twice as expensive for visiting everything except lambdas (because the function call has to jump through two JS objects...)

I don't understand this.

blerner:
Ok, looking at the tests that fail in resolve-scope.arr, this can be fixed by passing a name parameter into `thunk` -- that's *easy* to fix, and the tests aren't reliant on that name parameter.  The test that fails in desugar.arr is slightly more tedious to fix, because it's being cute and relying on surface-parsing on both sides of the test, but it's not hard to fix that either.  I'd rather have a `remove-lambda-names-visitor` created just for this test, than have an appendix file.  Are those the only reasons you needed the appendix?

blerner:
Sure, add a visit method to everything -- that seems more uniform to me, and you can get rid of the only-serializable thing.

blerner:
Sure, add a `map-visit` method to the visitors too; that would actually be more efficient than our current `the-list.map(_.visit(self))` stuff that we do far too much...

blerner:
I'd like to see at least a Makefile rule make sure that the ast-utils file is never out of date.  I just don't trust ourselves to manually keep it in sync.

jpolitz:
RE: generated files, I really don't want them in-tree.

We import `parse-pyret`, which is generated from JGLR, and I don't see any problem with that not being checked in. This repo already has enough issues with huge histories for `pyret.jarr` causing longer-than-necessary checkout times. I'd rather not make it worse.

sorawee:
Ben, yes the only reason I added the appendix is to work around the failing tests. Note however that generating another visitor will take significantly more time to compile. Since you're objecting hopping over two objects which only affects the compiling phase and not the runtime (as the dummy location visitor is only used in that phase), I'm not sure if it's really worth generating another visitor.

sorawee:
@jpolitz `parse-pyret` is not automatically generated. I think you meant `pyret-parser.js`.

But I disagree with your analogy. `pyret-parser.js` is used internally. On the other hand, my `ast-visitors.arr` will be in `trove`, used by users. For this reason, it should be checked in.

blerner:
@sorawee re "How will users know otherwise...?" -- that's what documentation is for.  If you can only figure out how to use something by reading the source, then we've failed at documentation utterly.  It is *not* important for people to read the code of the visitors themselves -- if anything, your PR here makes it even less important because users will know that it's entirely autogenerated, uniform code.

I'm happy with this PR overall (though I'd love to see the `lists.all(_.visit(self), _)` and `lists.map(_.visit(self), _)` idioms turn into helpers here rather than allocate the extra closures), and modulo whether we check this generated file into the repo or not, I'm good to accept this.  @jpolitz ?

sorawee:
> though I'd love to see the lists.all(_.visit(self), _) and lists.map(_.visit(self), _) idioms turn into helpers 

This was done!

blerner:
d'oh so it was -- Github "helpfully" didn't show the large diff where this was visible :)

jpolitz:
I'm happy to merge this if we keep generated code out-of-tree. This is a really good idea.

If the code is going to be in `trove` and we're worried about users knowing what it does, I agree with @blerner that it should be handled by documentation.

sorawee:
@jpolitz @blerner the problem is that the AST visitors generator uses AST visitors itself. If I do not check it in, it will result in a "cannot find dependency" error.

Does that mean we need a bootstrap? And when I compile the AST visitors generator, what command should I use? Currently I have:

```
node -max-old-space-size=8192 build/phase0/pyret.jarr --outfile build/phaseA/ast-visitors-generator.jarr \
  --build-runnable src/arr/compiler/visitors/ast-visitors-generator.arr \
  --builtin-js-dir src/js/trove/ \
  --builtin-arr-dir src/arr/trove/ \
  --compiled-dir build/phaseA/compiled/ \
  -no-check-mode \
  --require-config src/scripts/standalone-configA.json
```

But of course, compiling in this way will try to find the source file of AST visitors.

jpolitz:
This is certainly a bootstrapping problem.

How did you build ast-visitors-generator the first time? Did you use the visitors from `ast.arr`, a handwritten ast-visitors file, or something else?

I see two options here:

1. Make the visitor-generator script be implemented as a command-line option of the compiler, rather than a completely separate script.
2. Alternatively, check in  _built_ versions of libraries like the visitors, available in `phase0` or similar. `server-dev` does this with a release directory so that clients don't spend time compiling lists, options, sets, etc, and just see pre-built versions.

If the visitor-generator were its own standalone project, it would be a compiler that would need a bootstrapping process and it would need to be "installed," probably from a pre-built, already-bootstrapped "binary," in order to get started.

sorawee:
> How did you build ast-visitors-generator the first time? Did you use the visitors from ast.arr, a handwritten ast-visitors file, or something else?

Currently `ast-visitors-generator.arr` has `import ast-visitors as AV` where `ast-visitors.arr` was initially hand-written. But when I ran `ast-visitors-generator.arr`, `ast-visitors.arr` was overwritten with the autogenerated one. To address your concern about not checking in `ast-visitors.arr`, I removed the file, resulting in the problem.

I would rather separate this from the Pyret compiler, since Pyret compiler is to compile to JS, while my script doesn't have that functionality.

I'm lost on your last point. I was looking at https://github.com/brownplt/pyret-lang/compare/server-dev and didn't see any "checked-in built version". I also looked at the `Makefile` but didn't see any rule that will populate the `release` directory... And how is this process different from normal bootstrapping again?

PR #1280, opened 2017-12-02
Disable TCO when arguments are closed over a lambda/method

sorawee:


src/arr/compiler/anf-loop-compiler.arr
```diff
@@ -520,6 +520,37 @@ var total-time = 0
 
 show-stack-trace = false
 fun compile-fun-body(l :: Loc, step :: A.Name, fun-name :: A.Name, compiler, args :: List<N.ABind>, opt-arity :: Option<Number>, body :: N.AExpr, should-report-error-frame :: Boolean, is-flat :: Boolean, is-method :: Boolean) -> J.JBlock block:
+  var in-lam = false
+  var arg-used-in-lambda = false
+  arg-names = args.map(_.id)
+  dummy-anf-lettable = N.a-obj(A.dummy-loc, empty)
+  body.visit(N.default-map-visitor.{
```
blerner: This would be better done with `default-iter-visitor`, so that you don't have to fake up a `dummy-anf-lettable` here.

src/arr/compiler/anf-loop-compiler.arr
```diff
@@ -899,12 +930,14 @@ end
 
 fun is-id-occurs(target :: A.Name, e :: J.JExpr) block:
   doc: "Returns true iff `target` occurs in `e`"
+  dummy-js-expr = j-num(0)
   var found = false
   e.visit(J.default-map-visitor.{
```
blerner: Ditto.  If you're throwing away the result of the visit, then this can change to be the default-iter-visitor also.

src/arr/compiler/anf-loop-compiler.arr
```diff
@@ -520,6 +520,37 @@ var total-time = 0
...
+      in-lam := saved-in-lam
+      dummy-anf-lettable
+    end,
+    method a-id(self, shadow l, id) block:
+      when in-lam and not(arg-used-in-lambda) and arg-names.member(id):
+        arg-used-in-lambda := true
+      end
+      N.a-id(l, id)
+    end
+  })
```
blerner: I _think _you might need another case here, to handle `s-case-branch`es.  We sometimes compile those to be lambdas rather than inlined switch cases, and those might close over arguments too.  Definitely should test this situation to find out.

sorawee:
The default map visitor is the only thing that is available for JS AST and ANF AST for now. Once my auto generated visitors are merged, we can generate another kind of visitors easily and improve these. For now I'll stick with default map visitor.

sorawee:
@blerner `s-cases-branch` (or rather `ACasesBranch` since we are in the ANF'ed world) doesn't cause an issue. Consider the following code:

```
data Test:
  | variant(field)
end

fun test(a, b):
  v = cases (Test) b:
    | variant(x) =>
      shadow z = x + 0
      shadow z = z + 0
      shadow z = z + 0
      shadow z = z + 0
      shadow z = z + 0
      shadow z = z + 0
      a - 1
  end
  if a == 1:
    b.field
  else:
    test(1, variant(v))
  end
end

test(10, variant(20))
```

This will generate something along the line:

```
$step51 = 4;
$al56 = L[12];
var $temp_branch71 = function _d81b6581907580f5b54edb71fbe061b2d96bfd3f994d1b7956c1a378dd086301__13($x64) {
var $step72 = 0;
var $elidedFrames76 = 0;
var $ans73 = D;
var $al74 = L[11];
if(R.isAR($x64)) {
$step72 = $x64.step;
$al74 = $x64.from;
$ans73 = $x64.ans;
$elidedFrames76 = $x64.elidedFrames;
x64 = $x64.args[0];
} else {
var x64 = $x64;
}
if(--R.GAS <= 0 || --R.RUNGAS <= 0) {
R.EXN_STACKHEIGHT = 0;
$ans73 = R.mC();
}
while(!R.isContinuation($ans73)) {
switch($step72) {
case 0: $step72 = 1;
$al74 = L[10];
// omitting isFunction check;
$ans73 = _plus4.app(x64,(0));
break;
case 1: var z65 = $ans73;
$step72 = 2;
$al74 = L[9];
// omitting isFunction check;
$ans73 = _plus4.app(z65,(0));
break;
... [ELIDED] ...
case 6: $step72 = 7;
$al74 = L[4];
// omitting isFunction check;
$ans73 = _minus3.app(a53,(1));
break;
case 7: ++R.GAS;
return $ans73;
default: R.ffi.throwSpinnakerError(L[11],$step72);
}
}
if($step72 !== 7) {
$ans73.stack[R.EXN_STACKHEIGHT++] = R.mAR($al74,$temp_branch71,$step72,[x64],[],$elidedFrames76);
}
return $ans73;
};
$ans55 = cases59.$app_fields($temp_branch71,[false]);
break;
case 3: $step51 = 4;
$al56 = L[13];
$ans55 = R.throwNoCasesMatched(L[13],cases59);
break;
case 4: var v60 = $ans55;
$step51 = 5;
$al56 = L[3];
if(typeof(equal$always2.app) !== "function") {
R.ffi.throwNonFunApp($al56,equal$always2);
}
$ans55 = equal$always2.app(a53,(1));
break;
case 5: var anf_if63 = $ans55;
$step51 = R.checkPyretTrue(anf_if63)?6:7;
break;
... [ELIDED] ...
case 7: // caller optimization;
$ans55 = variant50.$var.app(v60);
var anf_arg62 = $ans55;
$step51 = 0;
++$elidedFrames58;
if(--R.RUNGAS <= 0) {
R.EXN_STACKHEIGHT = 0;
$ans55 = R.mC();
}
a53 = (1);
b54 = anf_arg62;
continue;
```

Note that it is true that `a53` gets mutated and that we have a `function` over `a53` when an `ACasesBranch` can't be inlined. However, the lambda gets applied right away (`$ans55 = cases59.$app_fields($temp_branch71,[false]);`) and this always happens before the tail recursion (since by definition, tail recursion is the last thing that happens), so the answer is still correct in this case.

I will merge this right now since it affects several students.

jpolitz:
Can we get that cases example turned into a test? I can imagine future optimizations might touch that pattern, and we'd like to know if it breaks!

sorawee:
Sure I will add.

sorawee:
Done

PR #1295, opened 2018-01-01
Annotation flatness

jpolitz:
Basic support for annotation flatness. Should be sufficient to fix issues (not all on Github) that were related to refinements screwing up stack safety with constructors and flat functions.


blerner:
Do we have a collection somewhere of the flatness-related broken programs we've seen, that we could add as regression tests in addition to the new tests you've added?  (I'm thinking of the various programs that crashed with weird stack-safety errors because an annotation was misconstrued as flat when it wasn't, or something like that...)

jpolitz:
Good point. We do have `regression/flat-refinement.arr` which summarized the function annotation position.

I realize now that I'm revisiting this, though, that there is likely a need to pass on flatness information to the compiler, so that it can know whether to compile an annotation check with a step or not. This is especially true for constructors. A good next step is to write some tests that trigger these issues.

Basically, this annotation flatness environment/annotation flatness function should be used in more places, particularly in the compiler when deciding whether or not to introduce steps for checks, and that isn't being done.

Issue #1269, opened 2017-11-25
(Type checker) type mismatch with raise and polymorphism


sorawee:
```
app = lam<a, b>(x :: (a -> b), y :: a) -> b:
  x(y)
end

app(lam(l :: Option<Number>) -> String:
    cases (Option) l:
      | none => "foo"
      | some(x) => raise("bar")
    end
  end, none)
```

produces:

```
Type checking failed because of a type inconsistency.
The type constraint String was incompatible with the type constraint Bot
```

Interestingly, 

```
lam(l :: Option<Number>) -> String:
  cases (Option) l:
    | none => "foo"
    | some(x) => raise("bar")
  end
end
```

alone type checked correctly. It's only when it is passed into a polymorphic function that the problem arises.

PR #1296, opened 2018-01-03
Fixes #1152

### LINKED ISSUE #1152: (Type checker) can't raise if returning type is not provided
```
fun f():
  if true:
    raise("a")
  else:
    "a"
  end
end
```

results in 

```
Type checking failed because of a type inconsistency.
The type constraint Bot was incompatible with the type constraint String
```

(Please note that if I change to `fun f() -> String:`, then I get no error)

mkolosick:
The subtyping order was backward in meet-branch-types


mkolosick:
A note that this also fixes #1269

blerner:
Cool -- can you add the program from #1269 as another regression test?  Then I'm happy to merge this.

blerner:
LGTM

Issue #1288, opened 2017-12-21
and-then's type is incorrect


sorawee:
```
some(1).and-then(lam(x :: Number): x + 1 end)
```

results in 

```
Type checking failed because of a type inconsistency.

The type constraint Number was incompatible with the type constraint Option<B>
```

Note that this is easy to fix. The type defaults in `type-defaults.arr` are simply wrong.

PR #1297, opened 2018-01-04
Fixing type defaults

### LINKED ISSUE #1265: (Type checker) error on spy statement
```
spy:
  a: 1
end
```

produces

```
The type checker rejected your program because the object type
{current-checker :: ( -> {check-raises-satisfies :: Bot, check-satisfies-not :: Bot, check-is-not-refinement :: Bot, check-is-refinement :: Bot, run-checks :: Bot, check-is :: Bot, check-raises-str :: Bot, check-raises-violates :: Bot, check-raises-not :: Bot, check-raises-other-str :: Bot, check-is-not :: Bot, check-satisfies :: Bot}), raw-list-filter :: forall a . ((a -> Boolean), List<a> -> List<a>), has-field :: ({}, String -> Boolean), raw-array-join-str :: forall a . (RawArray<a> -> String), raw-array-from-list :: forall a . (List<a> -> RawArray<a>), trace-value :: forall a . (Any, a -> a), raw-list-fold :: forall a, b . ((b, a -> b), b, List<a> -> b), list-to-raw-array :: forall a . (List<a> -> RawArray<a>), raw-list-map :: forall a, b . ((a -> b), List<a> -> List<b>)}

does not have a field named spy
```

mkolosick:
Fixes #1265 and #1288. Both are issues with types being not ascribed or incorrect.


tests/type-check/good/option.arr
```diff
@@ -0,0 +1 @@
+some(1).and-then(lam(x :: Number): x + 1 end)
```
blerner: This confirms that `and-then` type-checks, but it doesn't confirm the return type.  Could you add a `num-opt :: Option<Number> = ...` binding, to confirm the return type as well?

blerner:
@jpolitz does this PR meet our approval?  Can we merge it, and do we need to keep the branch referenced in #1298?

PR #1298, opened 2018-01-04
New bootstrap.

### LINKED ISSUE #1292: type-checker doesn't know simulate-trace's type
image

As a result, it's impossible to do any typed work with trace output.

mkolosick:
Fixes #1292

The bootstrapping was needed in order to add "Table" to the prims in type-util.js. This is required in order to use the "Table" annotation in JavaScript exports.


blerner:
I'm not sure exactly how to review this, because the relevant changes are tied up with the bootstrapping along the way.

I think I'd rather not merge a PR that has the change *and* the bootstrap in one piece, and *certainly* not in just a single commit, unless we're confident that the bootstrap has pulled in all the changes we want to have (and you've got a few more type fixes that are pending...).  @jpolitz your thoughts?  How divergent are horizon and master right now, and how many bootstraps do we think will be needed to merge them?

jpolitz:
Why does Table need to be a prim in type-util? Couldn't it use one of the long-form type names, or a local abbreviation, rather than requiring it to be a primitive? e.g. something like

```
 { origin: { "import-type": "uri", uri: "builtin://global" }, name: "Table" }
```

That could be in the shorthands list, and we could _also_ add it to type-util, but avoid bootstrapping.

Agreed that it's nice to have the commits separate from the bootstrap for better git control

mkolosick:
That's a good point I'll try that.

mkolosick:
I have made this fix on PR #1297

PR #1299, opened 2018-01-06
Fixes #1268 and fixes #1289

### LINKED ISSUE #1268: (Type checker) wildcard type mismatch
```
x = [list: 1, 2, 3]
x.map(lam(_): 1 end)
```

results in 

```
Type checking failed because of a type inconsistency.
The type constraint Any was incompatible with the type constraint Number
```

### LINKED ISSUE #1289: (Type checker) _ should be compatible with whatever it requires
```
fun f() -> (Number -> Number):
  lam(_): 1 end
end
```

results in

```
Type checking failed because of a type inconsistency.

The type constraint Number was incompatible with the type constraint Any
```

Ignoring input can't possibly go wrong. This should not result in an error.

mkolosick:
Sets underscore argument type as an existential so it becomes a polymorphic variable.

### DIFF
--- a/src/arr/compiler/type-check.arr
+++ b/src/arr/compiler/type-check.arr
@@ -1733,12 +1733,7 @@ fun collect-bindings(binds :: List<A.Bind>, context :: Context) -> FoldResult<SD
         | none =>
           cases(A.Name) binding.id:
             | s-atom(base, _) =>
-              if base == "$underscore":
-                # t-top(binding.l, false)
-                new-existential(binding.l, true)
-              else:
-                new-existential(binding.l, true)
-              end
+              new-existential(binding.l, true)
             | else =>
               new-existential(binding.l, true)
           end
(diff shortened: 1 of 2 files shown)


blerner:
Is it worth adding a few more tests for underscores (in cases patterns, let bindings, etc) to confirm that you've caught all possible places where you intend to infer an existential?

shriram:
That looks right. Thanks!

jpolitz:
@blerner @mkolosick I think this is another case where we probably need to bite the bullet and use some kind of `synth` or `artificial` tag on terms, right? This is going to be a continually larger `or` as we use existentials for more and more things (which I'm sure will happen).

I'm OK merging this for now, but I want to note the issue, because similar things come up with error reporting in desugar, especially with tables.

jpolitz:
@blerner are you satisfied with the explanation?

PR #1300, opened 2018-01-15
Shared Google Drive

ess476:
Here's the version of `shared-gdrive` that I've been working on.

I left a sample program in the root directory of the project. It _**should**_ work "out of the box".
Test via: `node src/server/pyret.js sample-program.arr`

**Issues:**

1. For some reason, the NPM modules `request` and `deasync` aren't being included in the standalone's produced. I assume this is because I didn't configure the standalone-config properly. This means that node will look for a node_modules folder in the directory where the program to be run is located. 
So `node src/server/pyret.js sample-program.arr` works,
but `node src/server/pyret.js ../sample-program.arr` doesn't.

2. This currently always downloads the source of included programs even if there is a compiled and up to date version stored locally (though it won't bother to compile it unnecessarily).

3. Some of the error messages (and exception types?) could be cleaned up.


schanzer:

Issue #1301, opened 2018-01-20
Non-associative roughnum equality error


jpolitz:
These should certainly produce the same result, not one be `false` and the other be `true`.

![image](https://user-images.githubusercontent.com/119891/35179011-6b5a277c-fd46-11e7-9bc3-ab201e83d39c.png)

blerner:
Yikes.  That's a sloppy mistake to have allowed through!  Is this a 
recent regression, or did we screw this up for quite some time now?

ds26gte:
Is it ok that both return `false`? (They can't return `true`. The only other option is a `can't compare` error.)

SethPoulsen:
Should this issue be closed since its a duplicate and the other identical issue has a pull request for it?

PR #1302, opened 2018-01-20
1 == ~1 (like ~1 == 1) should raise error

ds26gte:
This is a fix for #1301.

`equalAlways()` (in `runtime.js`) was checking `typeof` of its first arg only to decide if it wanted to do a plain JS equality check. This fix disallows this if first arg is `number` but second is not.  (In general, I've made `equalAlways()` attempt a basic JS equality only if both args are of the same primitive JS type. When not,  `equal3` takes over as usual, raising an exception as needed.)

Added tests to `test-equality.arr`. (Also fixed typo in test that was supposed to try equality between methods.)

### DIFF
--- a/src/js/base/runtime.js
+++ b/src/js/base/runtime.js
@@ -2282,7 +2282,9 @@ function (Namespace, jsnums, codePoint, util, exnStackParser, loader, seedrandom
     // JS function from Pyret values to Pyret booleans (or throws)
     function equalAlways(v1, v2) {
       if (arguments.length !== 2) { var $a=new Array(arguments.length); for (var $i=0;$i<arguments.length;$i++) { $a[$i]=arguments[$i]; } throw thisRuntime.ffi.throwArityErrorC(["equal-always"], 2, $a, false); }
-      if(typeof v1 === "number" || typeof v1 === "string" || typeof v1 === "boolean") {
+      if (typeof v1 === 'number'  && typeof v2 === 'number' ||
+          typeof v1 === 'string'  && typeof v2 === 'string' ||
+          typeof v1 === 'boolean' && typeof v2 === 'boolean') {
         return v1 === v2;
       }
       return safeCall(function() {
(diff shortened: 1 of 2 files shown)


src/js/base/runtime.js
```diff
@@ -2282,7 +2282,9 @@ function (Namespace, jsnums, codePoint, util, exnStackParser, loader, seedrandom
     // JS function from Pyret values to Pyret booleans (or throws)
     function equalAlways(v1, v2) {
       if (arguments.length !== 2) { var $a=new Array(arguments.length); for (var $i=0;$i<arguments.length;$i++) { $a[$i]=arguments[$i]; } throw thisRuntime.ffi.throwArityErrorC(["equal-always"], 2, $a, false); }
-      if(typeof v1 === "number" || typeof v1 === "string" || typeof v1 === "boolean") {
+      if (typeof v1 === 'number'  && typeof v2 === 'number' ||
```
blerner: Stylistically, it bothers me that this test is mixing `&&`, `||` and `===` without any parens -- I think my Pyret is showing :)

tests/pyret/tests/test-equality.arr
```diff
@@ -139,14 +139,22 @@ check "eq-none is equal nothing never always":
...
-check "error on bare fun and meth":
+check "error on bare fun and meth and roughnum":
   identical(f, f) raises f-func-err
   equal-always(f, f) raises f-func-err
   equal-now(f, f) raises f-func-err
 
   identical(m, m) raises f-meth-err
-  equal-always(f, f) raises f-func-err
-  equal-now(f, f) raises f-func-err
+  equal-always(m, m) raises f-meth-err
```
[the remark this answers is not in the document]
blerner: D'oh, I misread the diff.  Thanks for the clarification.

blerner:
I approve this diff; I'd like it with more parens but I'm fine without them.

src/js/base/runtime.js
ds26gte: I can change it, but please note that JS follows precedence conventions commonly followed in every written English text on logic that don't even bother to mention operator precedence, i.e., it is not arbitrary.  Arithmetic binds tighter than conjunction binds tighter than disjunction...

ds26gte: Parens added now.

Issue #969, opened 2017-01-13
Cannot compile file via absolute path


awstlaur:
Command
```
node build/phaseA/pyret.jarr 
  --build-runnable /home/alex/pyret-examples/example.arr
  --outfile /home/alex/pyret-examples/example.jarr
  --builtin-js-dir src/js/trove/
  --builtin-arr-dir src/arr/trove
  --require-config src/scripts/standalone-configA.json
```

Output
```
Gathering dependencies...The run ended in error:
Cannot find import dependency("file", [list: "/home/alex/pyret-examples/example.arr"])

Pyret stack:
  file:///home/alex/repos/pyret-lang/src/arr/compiler/cli-module-loader.arr: line 254, column 10
  file:///home/alex/repos/pyret-lang/src/arr/compiler/cli-module-loader.arr: line 245, column 4
  file:///home/alex/repos/pyret-lang/src/arr/compiler/cli-module-loader.arr: line 337, column 9
  file:///home/alex/repos/pyret-lang/src/arr/compiler/cli-module-loader.arr: line 384, column 18
  file:///home/alex/repos/pyret-lang/src/arr/compiler/pyret.arr: line 100, column 10
  file:///home/alex/repos/pyret-lang/src/arr/compiler/pyret.arr: line 69, column 4
  file:///home/alex/repos/pyret-lang/src/arr/compiler/pyret.arr: line 175, column 4
```

I am able to do this if I instead specify the same file with a relative path.

SethPoulsen:
I have the same problem.

schanzer:
@SethPoulsen @jpolitz did this eventually get merged into master?

SethPoulsen:
@schanzer looks like it did: 

```
seth@seth:~/git/pyret-lang$ git co master
Branch 'master' set up to track remote branch 'master' from 'origin'.
Switched to a new branch 'master'
seth@seth:~/git/pyret-lang$ git log | grep "Handle absolute" -B 5
Author: Joe Politz <joe@cs.brown.edu>
Date:   Thu Aug 30 07:50:06 2018 -0700

    Merge pull request #1317 from SethPoulsen/absolute-paths
    
    Handle absolute as well as relative paths (issue #969)
--

commit a7dcd40d6d2ffd5ea6463272ae76b75ade25ff94
Author: Seth Poulsen <poulsenseth@gmail.com>
Date:   Mon Mar 19 16:59:38 2018 -0600

[...]

PR #1312, opened 2018-03-12
Compliler can now handle absolute as well as relative paths (issue #969)

SethPoulsen:
I didn't rebuild phase0 because I didn't think this constituted a "major change" that would merit a whole new bootstrap.


SethPoulsen:
For what its worth, on my machine all tests pass and phaseB and phaseC build identically. I assume the failing build is due to some sort of authentication issue.

jpolitz:
Thanks for making this change! Would you mind opening a new PR that targets the `horizon` branch instead of `master`? We generally flow new changes through that branch, sorry that isn't clearer. I also made a change to the `.travis.yml` on `horizon` just now that should skip the credentials-requiring part of the build, allowing this to cleanly pass.

Separately, @blerner there's a design decision here – should we have the file locator be able to do relative paths? I think it's OK (it's more annoying to not have available) but we have a choice to lock things down to relative paths or not here, and it's a decision we're making either way.

SethPoulsen:
@jpolitz Sounds good. I knew that actually, I targeted `master` by mistake. It will take me a bit because I accidentally branched from `master` in the first place, but I should be able to sort it out soon.

blerner:
I think we do want to allow relative paths, but I think we need to be careful, because it'll make things a lot harder to abstract away from the machine-on-which-the-jarr-was-made's file system.  We had talked a while ago about canonicalizing paths in locators to be relative to the build directory, so that `/my/private/filesystem/organization/pyret-project/src` just becomes `src/`.  If we allow files outside that space, we'll need to be a bit more precise about what that normalization means.

SethPoulsen:
If you want everything in reference to the build directory internally, rather than allowing absolute and relative paths, we could just change the `get-real-path` function I created to convert all absolute paths to relative paths starting at the build directory.

Issue #1272, opened 2017-11-26
`make` should run `make phaseA`


sorawee:
@mkolosick raised this issue up somewhere but I think it got lost.

blerner:
This has gotten even less useful recently, now that `make` seems to run something related to stopify...

jpolitz:
Huh? It runs `browserify`, but there's no Stopify on horizon.

Agree that this should change, though.

blerner:
Mrph, meant to say "browserify" -- they're both "ify", though! ;-)

PR #1314, opened 2018-03-18
`make` will run `make phaseA` (issue #1272)

SethPoulsen:
`make` by itself without a specified target will just run whatever is the first target described in the makefile, which was `make showpath`. All I did was move `showpath` somewhere else so that `phaseA` would be the first target described in the makefile.


jpolitz:
Thanks, this LGTM and is a small change that's been lingering too long. Much appreciated!

SethPoulsen:
Did you mean to merge this one straight to master? Either way, it would be cool to have this one cherry-picked over to horizon too.

jpolitz:
Thanks for reminding me. This makes sense on both. I just cherry-picked over to horizon as well

PR #1322, opened 2018-03-27
A very rough sketch of using Node's repl library

jpolitz:
We shouldn't merge this in anything remotely resembling this state. This is pure prototype with bad code and missing features all over the place. I just want to have a place to discuss a bit.

### DIFF
--- a/src/arr/compiler/cli-repl.arr
+++ b/src/arr/compiler/cli-repl.arr
@@ -1,0 +1,27 @@
+import runtime-lib as RT
+import load-lib as L
+import string-dict as SD
+
+import js-file("./cli-repl") as CR
+
+import file("repl.arr") as R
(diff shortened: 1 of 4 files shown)


SethPoulsen:
@jpolitz Looks great, performance is actually pretty good, too. I'm just a little jealous you beat me to it ;)

I'm getting a crazy amount of output to the terminal before the actual result pops out, even on a successful computation. Any idea why?

jpolitz:
I figured there was going to be a little bit of magic incantation to get this off the ground, though pleasingly all the APIs except for the last step (plucking the value out of the mass of results generated) were pretty sensible to use.

I left plenty of interesting and valuable work for someone who wants to pick this up and run with it ;-)

PR #1323, opened 2018-03-27
Added .clear handler to Experimental REPL.

SethPoulsen:
A nice little addition to #1322, letting you clear the environment and restart without reloading everything.

### DIFF
--- a/src/arr/compiler/cli-repl.js
+++ b/src/arr/compiler/cli-repl.js
@@ -7,7 +7,7 @@
       var runInteractions = runtime.getField(pyretRepl, "run-interaction");
       function runUsingPyretRepl(src, context, filename, callback) {
         return runtime.runThunk(function() {
-            return runInteractions.app(src); 
+            return runInteractions.app(src);
           },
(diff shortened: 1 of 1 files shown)


src/arr/compiler/cli-repl.js
```diff
@@ -28,6 +28,9 @@
             console.log('Received "exit" event from repl!');
             restarter.resume("repl done");
           });
+          replServer.on('reset', () => {
+            restartInteractions.app("");
```
jpolitz: This should need to be `runtime.runThunk(function() { return restartInteractions.app(""); }, function(result) { /* probably just ignore the result here */ });`

This is because `restartInteractions`, in principle, could capture the stack and need to be restarted in the middle. In this case, I can't seem to make this break, but I think it's a useful comment regardless.

SethPoulsen: Done! Thanks for the tip. I'm still trying to wrap my head around all of the Pyret internals.

PR #1324, opened 2018-03-27
Added a more useful error message.

SethPoulsen:
This error really confused me when I first saw it, I had to dig into the code to see why I was even being given an error message at all.


jpolitz:
Thanks. `pyret.arr` needs general cleanup, sorry you had to dig a bit for this.

PR #1326, opened 2018-03-29
Node repl load definitions file

SethPoulsen:
After 
`make src/arr/compiler/cli-repl.jarr`,  
run 
`node src/arr/compiler/cli-repl.jarr ./definitions.arr` 
to load the file `./definitions.arr` into the REPL session. 

It would be nice if we could feed the remaining command line arguments to the definitions file, but I wasn't able to see a simple way to do it. I assume that the plumbing for doing such a thing hasn't been set up because it hasn't been needed for the CPO repl.

### DIFF
--- a/src/arr/compiler/cli-repl.js
+++ b/src/arr/compiler/cli-repl.js
@@ -7,7 +7,7 @@
       var runInteractions = runtime.getField(pyretRepl, "run-interaction");
       function runUsingPyretRepl(src, context, filename, callback) {
         return runtime.runThunk(function() {
-            return runInteractions.app(src); 
+            return runInteractions.app(src);
           },
           function(result) {
             if(runtime.isSuccessResult(result)) {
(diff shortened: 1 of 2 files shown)


src/arr/compiler/cli-repl.js
```diff
@@ -20,7 +20,9 @@
       }
       var restartInteractions = runtime.getField(pyretRepl, "restart-interactions");
       return runtime.safeCall(function() {
-          return restartInteractions.app("");
+          // need to convert definitions to JS string?
```
jpolitz: This is fine; Pyret strings are JS strings, so this will work without additional changes.

SethPoulsen: Yes, I know it works for now, but I was wondering if I was supposed to wrap it in a conversion function anyway--the docs say something about possible making Pyret strings more complex in the future.

Issue #1278, opened 2017-11-28
lists is not available in the REPL


sorawee:
In the definition window, we can write:

```
lists.sort([list: 1, 2, 3])
```

In the REPL, this results in unbound id.

SethPoulsen:
Should this be a code.pyret.org issue? Or does it fit here? I'm trying to figure out where I should look if I want to have a go at fixing it.

jpolitz:
It's going to need to be fixed in pyret-lang. The file
src/compiler/repl.arr defines the model of scope and modules in the REPL;
CPO is just a client of that library.

I have a feeling this issue is deeper rooted than just a trivial fix. There
are several interesting things that happen with types, scope, modules, and
the REPL.

I won't say don't dive in, since you seem to be having success navigating
and building the codebase! I just want to note that I suspect this could be
a bit of a rabbit hole.​

SethPoulsen:
Ok, sounds good. Is there currently an easy way to start a REPL without the website?

jpolitz:
You can do it programmatically with the support in
tests/pyret/tests/test-repl.arr (which is how pyret-lang tests the REPL
without CPO being present), but the best development workflow is probably
to use a local CPO.

There isn't a good command-line REPL right now. There was one a while ago
but it bitrotted and never got all the way there, sad to say.

[As a medium-sized, fairly separable project, a CLI REPL could be a
terrific contribution, for anyone listening in ;-)]

SethPoulsen:
This can be closed now, right?

sorawee:
This was merged to horizon (http://pyret-horizon.herokuapp.com/) but not CPO (http://code.pyret.org/). Probably should leave open until it's merged to CPO (to avoid people filing the same bug again)

PR #1327, opened 2018-03-30
Ensured that builtins--lists, arrays, etc. are defined in repl.

SethPoulsen:
Fix for #1278.


jpolitz:
This is a pretty clever fix. I pulled and tried this out and it looks good. Would you mind adding a test or three to `test-repl.arr` to show the updated behavior working, and we'll merge?

blerner:
Nitpick: could we use tree-sets here instead of lists, so that searching for is-standard-import isn't `O(#imports)` time?

SethPoulsen:
Done and Done.

Issue #1206, opened 2017-09-28
is-roughly should be recommended by roughnum test failure


jpolitz:
```
examples:
  num-sqrt(5) is num-sqrt(2 + 3)
end
```

Reports:

![image](https://user-images.githubusercontent.com/119891/30988872-9acc903e-a450-11e7-8298-66e241fbb39e.png)

But we have a better solution now – use `is-roughly`. The error message should probably guide the user in that direction as well. Since this error does double duty on `==` and on `is`, it's probably best to mention both.

A docs link would be really helpful here, if we have a sufficiently stable place to point them.

PR #1330, opened 2018-04-02
Updated comparing 2 roughNums message to direct user to is-roughly

doshiro:
Fix for #1206
Couldn't find any links to documentation about is-roughly in the glossary, though I checked here: https://www.pyret.org/docs/latest/Glossary.html

### DIFF
--- a/src/arr/trove/error.arr
+++ b/src/arr/trove/error.arr
@@ -2240,21 +2240,20 @@ data RuntimeError:
       ask:
         | is-number(value1) and is-number(value2) then:
           # one (or both) of them must be a roughnum
-          fun within-error(message):
+          fun roughnum-compare-error(message, function-name):
             [ED.error:
               [ED.para: ED.text(message)],
               [ED.para: ED.embed(value1)],
               [ED.para: ED.embed(value2)],
               [ED.para: ED.text("Consider using the "),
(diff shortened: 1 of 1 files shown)


jpolitz:
Good call on the glossary not linking to this! The relevant doc is at https://www.pyret.org/docs/latest/testing.html#%28part._testing_is-roughly%29, but it isn't indexed into the glossary.

I have a few comments on improving this overall. I think we should make this refer to `is-roughly` in all 3 cases of rough<->exact, rough<->rough, and exact<->rough. I propose this message:

> Attempted to compare a Roughnum to an Exactnum for equality, which is not allowed ... consider using is-roughly as the testing operator to compare them instead.

> Attempted to compare two Roughnums for equality, which is not allowed ... consider using is-roughly as the testing operator to compare them instead.

In general, this guides folks towards the thing they most often want, which is "roughly-equal" checking. If they have finer-grained tolerances, the is-roughly documentation links to `is%(within(...))`, which they can read about from there.

blerner:
I'm going to close this PR, because `is-roughly` isn't a *function*, but rather a *testing operator*.  The proper solution is to fix this in checker.arr.  I'll have a PR for that coming up soon, and open a new issue about the glossary.

PR #1333, opened 2018-04-22
Change Makefile to support multiple parsers.

InnPatron:
Change Makefile to support multiple parsers.

As a side note, I do not know how granular you want changes to be. I can just roll this change in with my error message overhaul.


jpolitz:
@blerner I see no issue merging this on its own, it seems like a pretty reasonable generalization of a few rules. Any objection?

PR #1337, opened 2018-05-01
Filtering and Ignoring Tokens

InnPatron:
While writing specific grammars (i.e. for 'balanced' blocks) for error analysis, I've found that I can ignore most of the data besides a few important tokens (keywords, parentheses, etc). 

**New Proposal**:
Treat the tokenizer as an iterator. Use a "filter" function which chooses what tokens are ignored or passed to the parser. If the filter function is not provided, a default filter is used.

**Original proposal**: Add special "__IGNORE" metatoken handling

Provide a way to discard tokens during tokenization. This pull requests integrates a filter for tokens with type "__IGNORE" (a metatype) which will NOT be passed to the parser while still passing over the string data.

Some alternatives:
1. Change token types and pass "__IGNORE" tokens to the grammar. However, I would like the error analysis grammars to be relatively simple.
2. Integrate productions from the full grammar. Again, I would like to keep the grammars simple, and it kind of defeats the purpose of the finer-grained grammars.
3. Write custom procedures to filter out unnecessary data. Prone to mistakes and may not match up with the expected output.

### DIFF
--- a/lib/jglr/jglr.js
+++ b/lib/jglr/jglr.js
@@ -81,7 +81,7 @@ define("jglr/jglr", ["jglr/rnglr"], function(E) {
               if (this.skippedTokens === undefined) {
                 this.skippedTokens = t.pos;
               } else {
-                this.skippedTokens.combine(t.pos);
+                this.skippedTokens = this.skippedTokens.combine(t.pos);
               }
               break;
             }
@@ -98,7 +98,7 @@ define("jglr/jglr", ["jglr/rnglr"], function(E) {
     if (p === undefined) {
       p = this.skippedTokens;
     } else if (this.skippedTokens !== undefined) {
-        p.combine(this.skippedTokens);
+      p = p.combine(this.skippedTokens);
     }
 
     return p;
--- a/lib/jglr/jglr.js
+++ b/lib/jglr/jglr.js
@@ -71,8 +71,6 @@ define("jglr/jglr", ["jglr/rnglr"], function(E) {
             else
               this.skippedWhitespace = this.skippedWhitespace.combine(newWS);
             break;
-          } else if (tok_type === "__IGNORE") {
-            break;
           }
           var t = this.makeToken(tok_type, s, SrcLoc.make(l, c, p, this.curLine, this.curCol, this.pos));
           this.curTok = t;
--- a/lib/jglr/jglr.js
+++ b/lib/jglr/jglr.js
@@ -1,9 +1,10 @@
 define("jglr/jglr", ["jglr/rnglr"], function(E) {
   const SrcLoc = E.SrcLoc
 
-  function Tokenizer(ignore_ws, Tokens) {
+  function Tokenizer(ignore_ws, Tokens, Whitelist) {
     this.ignore_ws = ignore_ws;
     this.Tokens = Tokens;
+    this.Whitelist = Whitelist;
   }
   Tokenizer.prototype.isEOF = function(tok) { return tok === E.EOF; }
   Tokenizer.prototype.tokenizeFrom = function(str) {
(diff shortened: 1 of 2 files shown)


blerner:
This is a good idea, but the particular implementation is risky, in that it can pretty easily screw up srcloc information for the lexer or parser. That's why the `skippedWhitespace `field exists on the tokenizer.  I'd rather see you generalize that idea to have a list of `skippedTokens` that can be checked by the parser if necessary.

InnPatron:
Do you mean something like this?

InnPatron:
That check in some form is necessary unless you want me to write a whitelist for all parsers (including the main one). That just seems like a maintenance burden.

My bad on `combine` and I will switch to a dictionary.

blerner:
If you initialize the whitelist of ignorable tokens to be empty, then you don't need to check if the whitelist exists. The default case is, "this token isn't ignorable", which is exactly what we want...

blerner:
Also, a friend just commented : congratulations on your first issue being leet 🤘

InnPatron:
Ah, I was interpreting the whitelist of **accepted** tokens rather than ignored tokens.

InnPatron:
Alright, so do we want a list of accepted tokens or a list of ignored tokens?

I am in favor of a list of accepted tokens because:
1. Easier to use in my case. My parsers will be ignoring large parts of the language.
2. Less maintenance burden if tokens are added or replaced. With a list of ignored tokens, if another token is added, it has to be added to each used list or else the parsers will break.

blerner:
Perhaps easier in your case, but the typical use of a tokenizer and parser is to actually read input, not discard it. It should be the job of the parser to ignore input, but that's tedious, so your tokenizer should ignore inputs just as it does whitespace or comments. I don't think we want the default to be "ignore this token".

InnPatron:
But the default is not "ignore this token". It is "ignore this token if the whitelist exists and it is not on the whitelist".

blerner:
Conceptually, it's just wrong though: a typical tokenizer does not ignore its input. So having a list of tokens to be *ignored* is correct: if the list is missing, it's equivalent to an empty lost, which implies that all tokens should be preserved. Pyret's default tokenizer would include just whitespace and comments, and your reparative tokenizer could ignore a bunch of other stuff too.

InnPatron:
Input into the tokenizer is not being ignored. Output of the tokenizer is being filtered. The tokenizer is still advancing its internal state regardless of whether the token is ignored or not.

From Wikipedia: > Tokenization is the process of demarcating and possibly classifying sections of a string of input characters. The resulting tokens are then passed on to some other form of processing. The process can be considered a sub-task of parsing input.

I do not think there is an inherent reason for tokenization to be immediately followed by parsing. The filter is being integrated into tokenizers only for convenience, so it is not necessarily part of the tokenizer _per se_.

Besides, 0tThe grammars I will be writing will be operating on a restricted set of terminals. The alternatie

InnPatron:
Yeah, I am having trouble understanding your reasoning @blerner.

Input into the tokenizer is not being ignore. Output from the tokenizer is being filtered. The filtering is an entirely separate step. I do not buy it is necessarily the job of the parser to ignore tokens. If the token is discarded before the parsing phase, that just means the parser's set of terminals did not include the discarded tokens (i.e. it's not part of the grammar in the first place). Pyret is already doing that in regards to comments and whitespace.

You want a list of tokens to ignore, with an empty or undefined list implying allow all tokens. You argue that this is more correct because tokenizers typically do not ignore input/output. That is correct, but as I've said before, I'm modeling this as an entirely separate step. The filter is just integrated into the tokenizer for convenience.

I want a list of allowed tokens, with an undefined list implying allow all tokens. Whitespace and comments are handled outside of this whitelist (at least for whitespace). The default is still to allow all tokens. As I have said before, maintaining a list of allowed tokens is easier to maintain for the possibly many "sub-parsers" I intend to write.

Alternatively, I can just pass in a function supplied by the tokenizer which acts as the filter, instead of imposing a list-to-ignore or a list-to-accept requirement.

jpolitz:
Closing this for now since we decided on a different approach.

PR #1344, opened 2018-06-13
Fix error that occurs when using implicit version of `spy` with `var`

### LINKED ISSUE #1342: Using `spy` with `var` produces strange error message
Attempting to use `spy` with an assignable variable produces a strange error message:

```
var x = 1
spy: x end
```

screen shot 2018-06-12 at 10 51 59 am

The docs don't say whether or not `spy` should work with assignables, but since `print` works with `var` variables, I expected `spy` to work as well.

The following program also produces a similar error whether or not `f` is applied to anything:

```
fun f(x):
  var y = 1
  spy: y end
end
```

screen shot 2018-06-12 at 10 56 49 am 1

ZacharyEspiritu:
Fixes #1342.

In regards to the updated constructor for `s-spy-expr`, I wasn't sure what the `name` of something that used to fall under the `s-spy-name` case should be (since for those cases, the `name` in the new constructor is now a String instead of an Expr), so I took a guess and in parse-pyret.js set the `name` for things that used to be `s-spy-name` to the empty string and used `value` in places where it was necessary.

### DIFF
--- a/src/js/trove/parse-pyret.js
+++ b/src/js/trove/parse-pyret.js
@@ -239,11 +239,11 @@
         },
         'spy-field': function(node) {
           if (node.kids.length === 1) {
-            return RUNTIME.getField(ast, 's-spy-name')
-              .app(pos(node.pos), tr(node.kids[0]));
+            return RUNTIME.getField(ast, 's-spy-expr')
+              .app(pos(node.pos), RUNTIME.makeString(""), tr(node.kids[0]), RUNTIME.makeBoolean(true));
           } else {
             return RUNTIME.getField(ast, 's-spy-expr')
-              .app(pos(node.pos), symbol(node.kids[0]), tr(node.kids[2]));
+              .app(pos(node.pos), symbol(node.kids[0]), tr(node.kids[2]), RUNTIME.makeBoolean(false));
           }
         },
         'data-with': function(node) {
--- a/src/js/trove/parse-pyret.js
+++ b/src/js/trove/parse-pyret.js
@@ -240,7 +240,7 @@
         'spy-field': function(node) {
           if (node.kids.length === 1) {
             return RUNTIME.getField(ast, 's-spy-expr')
-              .app(pos(node.pos), RUNTIME.makeString(""), tr(node.kids[0]), RUNTIME.makeBoolean(true));
+              .app(pos(node.pos), symbol(node.kids[0].kids[0]), tr(node.kids[0]), RUNTIME.makeBoolean(true));
           } else {
             return RUNTIME.getField(ast, 's-spy-expr')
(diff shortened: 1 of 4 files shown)


blerner:
The name field *must not* be empty, or things will break badly.  The semantics of `spy` is that you provide a "dictionary"-looking collection of field-names and values:
```
spy:
  foo: some expression
  bar: some expression
end
```
and it'll show up as a blue table in the REPL area, with rows named `foo` and `bar`.  The *shorthand* syntax, `spy: some-pyret-name end`, is a contraction that means `spy: some-pyret-name: some-pyret-name end`, i.e., a value to be spied upon whose name is the same as the identifier itself.

src/js/trove/parse-pyret.js
```diff
@@ -239,11 +239,11 @@
         },
         'spy-field': function(node) {
           if (node.kids.length === 1) {
-            return RUNTIME.getField(ast, 's-spy-name')
-              .app(pos(node.pos), tr(node.kids[0]));
+            return RUNTIME.getField(ast, 's-spy-expr')
+              .app(pos(node.pos), RUNTIME.makeString(""), tr(node.kids[0]), RUNTIME.makeBoolean(true));
```
blerner: Here, you know (because of the grammar) that `node.kids[0]` is an identifier, so I think the right thing to do here is `string(node.kids[0])`

ZacharyEspiritu:
Makes sense—pushed some changes that work, though when I tried to use `string(node.kids[0])` I got an error that looks like this:

```
TypeError: Cannot read property 'brands' of undefined
    at hasBrand (cpo-main.jarr:245908)
    at PFunction.app (cpo-main.jarr:245924)
    at isRuntimeError (cpo-main.jarr:278429)
    at getFancyRenderer (cpo-main.jarr:278494)
    at renderError (cpo-main.jarr:278573)
    at _fulfilled (q.js:834)
    at q.js:863
    at Promise.promise.promiseDispatch (q.js:796)
    at q.js:604
    at runSingle (q.js:137)
    at MessagePort.flush (q.js:125)
```

Also switched around to `symbol(node.kids[0])` and I got a similar error. Eventually I changed that to `symbol(node.kids[0].kids[0])` which got the desired behavior, but that feels a little bizarre to me...

blerner:
Oh, wait, weird...Yeah, you do want to use `symbol` here after all -- `string` is for slicing off the quote marks and creating the Pyret value of the string contents, while `symbol` just grabs an id.  And if you look at the uses of it, your overall code of `symbol(node.kids[0].kids[0])` is correct: `node.kids[0]` is the `id-expr` from the grammar, whose first child (i.e., the second `.kids[0]`) is a `NAME`, and that's the token whose contents you're trying to grab.

PR #1359, opened 2018-07-02
Fix 1339

jpolitz:


jpolitz:
The only interesting point for this is that

```
f :: <A> A -> Number
fun f<A>(x):
  0
end
```

is _not_ an error, while

```
f :: (x :: Number) -> Number
fun f(x :: Number):
  0
end
```

_is_ an error. The rationale is that the `A` name could be used inside `f`, and it's natural to want to write the binding around the expression in which it is used.

Other than that, this is straightforward plumbing to make parametric annotations weave in.

PR #1361, opened 2018-07-03
Changing the desugaring of bracket expressions to become a *non-flat*…

### LINKED ISSUE #1307: Bracket access on tables fails badly
From John Clements:
```
# include the DataScience Library
include shared-gdrive("Bootstrap-DataScience-v2.arr", "0BzzMl1BJlJDkZ2NoSnFoQl91NFE")
include tables
include image

# a table of shapes
shapes = table: name, sample, corners
  row: "triangle", triangle(30, "solid", "red"),14
  row: "square",  square(20, "solid", "blue"),4
  row: 34, 47,9
  row: "abc","def",16
  row: "rectangle", rectangle(10, 20, "solid", "red"),5
end

shapes[0]
```
Results in the error
```
Evaluating a field lookup expression in <builtin runtime> errored.

does not have a field named get-value

Please report this as a bug.
```

Notice that there's not even a value shown that "does not have a field named `get-value`".  There's nothing in the console, either.

jpolitz:
… prim-app.  This fixes #1307.

The concept of a non-flat prim-app is of broader utility: it allows us to define operations as primitive runtime functions that nevertheless might fail, and so they requires a step in compiled output.  (Several other things could be refactored to use non-flat prim-apps too, which currently are compiled via clunkier workarounds with methods and such.)

### DIFF
--- a/src/js/base/runtime.js
+++ b/src/js/base/runtime.js
@@ -476,13 +476,17 @@ function (Namespace, jsnums, codePoint, util, exnStackParser, loader, seedrandom
     function getBracket(loc, obj, field) {
       checkArityC(loc, 3, arguments, false);
       if (obj && obj.dict && obj.dict["get-value"]) {
-        return getFieldLoc(obj, "get-value", loc).app(field);
-      } else {
-        raiseJSJS(
-          thisRuntime.ffi.contractFail(
-            makeSrcloc(loc),
-            thisRuntime.ffi.makeBadBracketException(makeSrcloc(loc), obj)));
+        var gV = getColonFieldLoc(obj, "get-value", loc);
+        if (thisRuntime.isMethod(gV)) {
+          return gV.full_meth(obj, field);
+        } else if (thisRuntime.isFunction(gV)) {
+          return gV.app(field);
+        }
       }
+      raiseJSJS(
+        thisRuntime.ffi.contractFail(
+          makeSrcloc(loc),
+          thisRuntime.ffi.makeBadBracketException(makeSrcloc(loc), obj)));
     }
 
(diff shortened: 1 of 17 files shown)


src/js/base/runtime.js
```diff
@@ -475,6 +473,18 @@ function (Namespace, jsnums, codePoint, util, exnStackParser, loader, seedrandom
       return thisRuntime.getFieldLoc(obj, field, ["runtime"]);
     }
 
+    function getBracket(loc, obj, field) {
+      checkArityC(loc, 3, arguments, false);
+      if (obj && obj.dict && obj.dict["get-value"]) {
+        return getFieldLoc(obj, "get-value", loc).app(field);
```
jpolitz: We could call isFunction on the result of the getField here for another `BadBracketException` case. As-is this will apply undefined for `{get-value: "not-a-function"}`

src/js/trove/table.js
```diff
@@ -515,9 +515,9 @@
           return runtime.ffi.makeList(headers);
         }),
 
-        'stack': runtime.makeMethod1(function(_, otherTable) {
+        'stack': runtime.makeMethod1(function(self, otherTable) {
           ffi.checkArity(2, arguments, "stack", true);
-          runtime.checkArgsInternal1("tables", "stack", otherTable, annTable);
+          runtime.checkArgsInternal1("tables", "stack", self, annTable, otherTable, annTable);
```
jpolitz: It occurs to me that we could have a checkMethodArgsInternal1-N that does this check, doesn't check the `self` param (assuming currying worked out), and gets the arity right.

Of course, since in the current compiler the `self` _could_ be the wrong type because of detached currying, this is necessary.

PR #1364, opened 2018-07-06
Fix 383

blerner:


blerner:
I'm not sure why this is failing -- it passes on my machine.  I'm going to experiment with forcing node to flush its output a bit better (we've hit that annoyance before with travis...) so I can figure out what's going wrong.

blerner:
Hmmm, the build is passing but not showing output.  I don't trust it; I'll keep trying.

blerner:
Hmm, well, I found the error in the tests, even if I didn't find why the output is getting truncated.

PR #1367, opened 2018-07-06
New feature: creating a directory tree

blerner:
Implements a new method `create-dir-tree`, which is like `mkdir -p` but enhanced.  Given a base directory, and a string-dict of string-dicts of ..., where the innermost string-dicts' values are irrelevant, build a directory tree whose nesting structure corresponds to the keys of the string-dicts.

### DIFF
--- a/src/js/trove/filelib.js
+++ b/src/js/trove/filelib.js
@@ -1,8 +1,10 @@
 ({
-  requires: [],
+  requires: [
+    { "import-type": "builtin", name: "string-dict" },
+  ],
   provides: {},
   nativeRequires: ["fs"],
-  theModule: function(RUNTIME, NAMESPACE, uri, fs) {
+  theModule: function(RUNTIME, NAMESPACE, uri, SD, fs) {
     function InputFile(name) {
       this.name = name;
       this.fd = fs.openSync(name, "r")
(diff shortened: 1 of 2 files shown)


shriram:
WHOA.

What does this do in the browser?

blerner:
Nothing -- it's in the filelib module, which isn't made available on CPO.

src/js/trove/filelib.js
```diff
@@ -27,6 +30,17 @@
...
       this.fd = fs.openSync(name, (append ? "a" : "w"));
     }
+    var checkISD = RUNTIME.getField(SD, "internal").checkISD;
+    var isISD = RUNTIME.getField(SD, "internal").isISD;
+    function makeDirs(baseDir, tree) {
+      if (!fs.existsSync(baseDir)) fs.mkdirSync(baseDir);
+      if (isISD(tree)) {
+        RUNTIME.ffi.toArray(RUNTIME.getField(tree, "keys-list").app()).forEach(function(sub) {
+          var kid = RUNTIME.getColonField(tree, "get-value").full_meth(tree, sub);
+          makeDirs(baseDir + "/" + sub, kid);
```
jpolitz, line 40: Should we be more judicious about using `path.join` or whatever the blessed `path` library thing is? I know we have other windows issues but that's probably the right thing to do.

src/js/trove/filelib.js
```diff
@@ -27,6 +30,17 @@
       this.name = name;
       this.fd = fs.openSync(name, (append ? "a" : "w"));
     }
+    var checkISD = RUNTIME.getField(SD, "internal").checkISD;
+    var isISD = RUNTIME.getField(SD, "internal").isISD;
+    function makeDirs(baseDir, tree) {
+      if (!fs.existsSync(baseDir)) fs.mkdirSync(baseDir);
+      if (isISD(tree)) {
+        RUNTIME.ffi.toArray(RUNTIME.getField(tree, "keys-list").app()).forEach(function(sub) {
+          var kid = RUNTIME.getColonField(tree, "get-value").full_meth(tree, sub);
```
jpolitz, line 39: Maybe worth a comment that this won't stack capture because of what we know about `immutable-string-dict`'s `get-value` method being flat? Like if this was changed to a generic `.get-value` dictionary interface, this wouldn't be safe.

src/js/trove/filelib.js
blerner: I can't tell, tbh: https://nodejs.org/docs/latest/api/fs.html#fsmkdirpath-options-callback gives an example of calling `mkdir` on `"/"` even on Windows...

src/js/trove/filelib.js
blerner: Yeah, possibly

PR #1374, opened 2018-07-11
Server dev

jpolitz:
This PR implements a bunch of CLI-focused changes for the Pyret server.

This needs some changes before it can merge. For example, package.json and the README were designed to work with this as an npm package, and these changes should NOT be merged into horizon.

It's useful to get the PR up so we can see the aggregate changes to the server code, which include:

- some better options for controlling check block running
- allowing standalones to use installed NPM modules as well as bundled ones
- numerous improvements to handling working directories and relative paths for the client and server

@TheOrangeLego and I have made the https://github.com/brownplt/pyret-npm repository to separate the client connection logic and npm package metadata into a separate repo. This lets us avoid having this repo (`pyret-lang`) be responsible for being installable as an npm package with a runnable entrypoint and so on.


jpolitz:
Ah, I forgot, this _also_ includes `--perilous`, which optionally strips all annotations and several different runtime checks.

SethPoulsen:
Yes! This is exciting.

SethPoulsen:
With this, are we going to be able to have a desktop version of Pyret available through `apt` and `brew`?

jpolitz:
That is absolutely the plan. The stuff I've been prototyping targets `npm` first; it may well be that `npm install -g pyret` ends up being the preferred way to do this, and any apt/brew/yum and so on package would wrap that idea.

A lot of this works. If you check out https://www.npmjs.com/package/pyret-npm (which you can install with `npm install -g pyret`) you get a reasonable `pyret` command installed; the default thing to do is `pyret <some-pyret-file>.arr`. I'm not advertising it at all because it's super-alpha and breaks a lot, but it's on a pretty reasonable trajectory.

If you want the simplest, stupidest thing that is currently broken to push on, in this repo:

https://github.com/brownplt/pyret-npm

I updated some dependencies in npm and I broke all the help text rendering at the console (the "chalk" library having a major version change is to blame). I could downgrade it but there were a bunch of good and related reasons to update several npm dependencies. That repo autodeploys to the npm package when pushes happen on `release`, so it's relatively easy to take PRs and make improvements there.

The idea is that `pyret-npm` will have an implementation of a client for the server side that's implemented in this PR. So fixing the way all the cli args help renders in that repo would unbreak `pyret-npm` in a useful way.

I'd be happy to chat about longer-term plans for this as well, there's a few different directions to push in, both in terms of libraries to expose at the node level, how to integrate well with npm packages, and how to get better performance out of Pyret when we're outside the browser environment.

SethPoulsen:
I've created a homebrew tap [here](https://github.com/SethPoulsen/homebrew-pyret) to (hopefully) make this accessible to even more people. It will automatically install the `node` dependency and everything (use `--ignore-dependencies` if you already have node installed not via brew). Just do

```
brew install SethPoulsen/pyret/pyret
```

The way it works is by pulling from the npm archives at https://registry.npmjs.org/pyret-npm/-/pyret-npm-0.0.16.tgz, which is in parity with how the other brew core taps work when pulling npm packages.

Once @jpolitz is happy enough with the stability of what's there, I can create a PR to homebrew-core, and then it will be as simple as 

```
brew install pyret
```

PR #1379, opened 2018-07-20
Refactor parse for blocks

justinpombrio:
This is a pull request for the refactoring of parsing that Ben suggested.

The idea that `src/js/trove/parse-pyret.js` mixed two things together:

1. Cleaning out all the cruft in the jglr parse tree.
2. Constructing a Pyret AST data structure.

The `codemirror-blocks` block editor wants (1), but not (2), however: it wants to construct its own, blocky, AST. Therefore, this commit separates (1) out into a separate file, `src/js/base/translate-parse-tree.js`. This file provides a function called `translate` that takes a jglr parse-tree and a set of node constructors (as well as some other constructors, like `makeSrcloc`), and produces an AST using those constructors. Pyret passes in the typical constructors from ast.arr, and the block editor will pass in its own constructors.

The tests from `make test` all pass.

I'm happy to tweak the interface if something else would be preferable. (See the top of `translate-parse-tree.js`.) As is, I kept things very close to how they were before, leading to different constructors being used in different ways. For example:
- makeNode('s-num', srcloc, 3)
- opLookup["+"]
- opLookup["is"]\(srcloc)

The hackiest things here are the arguments "getRecordFields", which is used to get the record fields out of a provide (which re-uses record syntax), and "detectAndComplainAboutOperatorWhitespace", which is a heuristic for better parse errors that I don't see how to separate out.


src/js/trove/parse-pyret.js
```diff
@@ -1387,6 +107,48 @@
       "raises-violates":   function(l){return RUNTIME.getField(ast, "s-op-raises-violates").app(l);},
     }
 
+    function translate(node, fileName) {
+      let constructors = {
+        "makeNode": function(con) {
+          let args = Array.prototype.slice.call(arguments, 1);
```
blerner: I'm not a huge fan of this, because gratuitous `arguments` munging is inefficient.  I think I'd rather see `makeNode` be curried over the constructor name, e.g. 
```js
constructors = {
  make: {
    "a-dot": RUNTIME.getField(ast, "a-dot").app
    "s-provide": RUNTIME.getField(ast, "s-provide".app
    ...
  }
}
```
so that you'd call it as `constructors.make["a-dot"](......)`.  And you could eta-expand any actual singleton values, so that e.g. `constructors.make["s-normal"]()` would just call `function() { return RUNTIME.getField(ast, "s-normal"); }`

And yes, I get that it makes the `constructors` object longer.  Perhaps the right way to do this is to iterate reflectively over the provided values out of `ast` and construct the necessary `constructors` that way.  Specifically, for each name `ast["provides"]["datatypes"][*][3][*]`, you'll get a constructor.  The name is at index [0], and if ....[1] is defined then it's a normal constructor, otherwise it's a singleton.

src/js/trove/parse-pyret.js
```diff
@@ -1387,6 +107,48 @@
...
+          if (args.length === 0) {
+            return RUNTIME.getField(ast, con);
+          } else {
+            return RUNTIME.getField(ast, con).app.apply(this, args);
+          }
+        },
+        "opLookup": opLookup,
+        "makeLink": function(head, tail) {
```
blerner: I don't think any of these need to be eta-expanded, do they?

src/js/base/translate-parse-tree.js
```diff
@@ -0,0 +1,1312 @@
...
+        } else {
+          // (provide-stmt PROVIDE stmt END)
+          return makeNode('s-provide', pos(node.pos), tr(node.kids[1]));
+        }
+      },
+      'provide-types-stmt': function(node) {
+        if (node.kids[1].name === "STAR") {
+          return makeNode('s-provide-types-all', pos(node.pos));
+        } else {
+          // will produce record-ann
```
blerner: This was always a bit of a pun and therefore a hack.  Since this is the only reason you need `getRecordFields`, it might be worth rewriting this to not be so punny.

src/js/base/translate-parse-tree.js
```diff
@@ -0,0 +1,1312 @@
...
+    let makeBoolean          = constructors.makeBoolean;
+    let makeNone             = constructors.makeNone; // RT.ffi.makeNone
+    let makeSome             = constructors.makeSome; // RT.ffi.makeSome
+    let getRecordFields      = constructors.getRecordFields; // getField(rec, 'fields')
+    // makeSrcloc: fileName, pos -> srcloc
+    let makeSrcloc         = constructors.makeSrcloc;
+    // combineSrcloc: fileName, pos, pos -> srcloc   -- beg of first to end of second
+    let combineSrcloc      = constructors.combineSrcloc;
+    // To improve error messages in an edge case. See parse-pyret.arr.
+    let detectAndComplainAboutOperatorWhitespace
```
blerner: Not much to be done about this one :(

blerner:
@justinpombrio ping -- please rebase on the latest changes to horizon, and please also revise as we talked about.

src/js/trove/parse-pyret.js
justinpombrio: makeLink has fixed arity, so it's fine. opLookup was a dictionary that returned either values for binops or (srcloc->value) functions for checkops, which was a little clunky so I'm replacing it with two separate dictionaries for binops and checkops.

src/js/base/translate-parse-tree.js
justinpombrio: Well, nothing to be done but give good names and document why it's there :-)

src/js/base/translate-parse-tree.js
blerner: No no -- I'm saying, instead of calling `tr` on the first child, directly, copy the logic from the translation of `record-ann` into here, but *don't* create an actual `s-record-ann` object. Just grab the fields and use them they way you need to.

justinpombrio:
SO says that rebasing in this case [is questionable](https://stackoverflow.com/questions/19016698/git-branch-diverged-after-rebase). I tried it, and it resulted in a "refactor-parse-for-blocks and origin/refactor-parse-for-blocks branches have diverged" message, from which I would have had to force push. Instead, I merged horizon into the branch.

src/js/trove/parse-pyret.js
justinpombrio: Your knowledge of the internal representation seems a bit out of date :-). Here's what I was able to come up with instead:

    let nodes = {};
    const astValues = RUNTIME.getField(astLib, "defined-values");
    for (let i in astValues) {
      let value = astValues[i];
      if (value["$constructor"] !== undefined && value["$name"] === i) {
        // It's a singleton.                                                    
        // If there's a better way to tell, please change the conditional.      
        let name = value["$name"];
        nodes[name] = function() { return value; }
      }
      else if (value["$constrFor"] !== undefined) {
        // It's a normal constructor                                            
        let name = value["name"];
        nodes[name] = value.app;
      }
    }

The test to check if something's a singleton is horrific. I spent a while looking through `defined-values` and `defined-types`, and that's the best I could come up with. If there's a saner way to do this we should.

blerner:
Merging horizon onto this instead of rebasing is fine.

src/js/trove/parse-pyret.js
blerner: Looking directly at the compiled output of ast.arr, the description I gave above is accurate.  It doesn't give you the `.app` closure directly, but it tells you what variants are defined for each datatype, and whether they're singletons or not.  Looking at `$constructor` or `$constrFor` is not really a stable way of doing this...

justinpombrio:
Think this is ready to merge? The one nasty case is my check for whether a value is a singleton or not:

    const astValues = RUNTIME.getField(astLib, "defined-values");
    for (let i in astValues) {
      let value = astValues[i];
      if (value["$constructor"] !== undefined && value["$name"] === i) {
        // it's a singleton
      }

I looked through the fields of `defined-values` and `defined-types`, and didn't see any other hint as to whether something was a singleton or not.

blerner:
Well, the lack of an `.app` field is another good clue, rather than checking the `$name` field.  Any $-field only ever comes from compilation, so that `$constructor` field is a safe thing to use as a proxy for "is this a `data` variant or not".

This is probably ready to merge; I'll look it over tomorrow in more detail.

justinpombrio:
So, I just noticed this open PR I've had :-). Could we either merge or close this?

schanzer:
I would seriously love it if this became a reality

schanzer:

PR #1381, opened 2018-07-27
Add ast-visitors

sorawee:
Note that adding it properly will require bootstrapping. To reduce noises for
the PR request, I have not bootstrap it yet.


sorawee:
To elaborate:

After bootstrapping, `node phase0/pyret.jarr -auto-generate` will be required as a first step in compilation to generate `ast-visitors.arr`. It uses `ast.arr` as a source for generation.

With `ast-visitors.arr`, the usual compilation process can begin normally.

`ast-visitors.arr` will be added to the gitignore file to prevent it entering the repo.

Makefile
```diff
@@ -132,6 +132,9 @@ showpath:
 $(BUNDLED_DEPS): src/js/trove/require-node-dependencies.js
 	browserify src/js/trove/require-node-dependencies.js -o $(BUNDLED_DEPS)
 
+# src/arr/trove/ast-visitors.arr: src/arr/trove/ast.arr src/arr/compiler/autogenerate.arr
```
jpolitz: Should this only depend on ast.arr once things are merged? Autogenerate will be built into the compiler.

src/arr/compiler/pyret.arr
```diff
@@ -123,7 +126,10 @@ fun main(args :: List<String>) -> Number block:
         print-error("Passing command line arguments without compiling standalone no longer supported\n")
         failure-code
       else:
-        if r.has-key("build-runnable") block:
+        if r.has-key("auto-generate") block:
```
jpolitz: I wish this were more parameterized, because others might have a use for this feature as well. For example, why not take in two paths with auto-generate, and make one be a source file and one be a destination file. Then if I wanted visitors for ast-anf, I might be able to generate those without writing any new Pyret code.

schanzer:

PR #1387, opened 2018-08-04
Move Source Line FFI helpers

InnPatron:
Move the Source Line FFI helpers from parse-pyret.js to ffi.js

As far as I can tell, error data types expect a transformed source line data structure. As I am writing other error checking modules partially in Javascript, it would be useful to have these helpers available in general.

### DIFF
--- a/src/js/trove/ffi.js
+++ b/src/js/trove/ffi.js
@@ -16,6 +16,7 @@
   nativeRequires: [],
   theModule: function(runtime, namespace, uri, L, Se, O, E, EQ, ERR, S, CON, /* CH, */ ED, VS) {
     var gf = runtime.getField;
+    var srcloc = runtime.getField(S, "values");
     L = gf(L, "values");
     Se = gf(Se, "values");
     O = gf(O, "values");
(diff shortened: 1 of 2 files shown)


src/js/trove/ffi.js
```diff
@@ -16,6 +16,7 @@
   nativeRequires: [],
   theModule: function(runtime, namespace, uri, L, Se, O, E, EQ, ERR, S, CON, /* CH, */ ED, VS) {
     var gf = runtime.getField;
+    var srcloc = runtime.getField(S, "values");
```
blerner: This is redundant with line 25 below (`S = gf(S, "values")`)

blerner:
This change looks ok to me.  I don't *think* it changes the dependency graph among parse-pyret, ast, srcloc, and ffi in any harmful ways, so this should be fine.

InnPatron:
Updated.

InnPatron:
Hmm... Actually it would make more sense if the transformation occured within the "throw error" functions. And upon further inspection, some of them actually do (ffi.js:435) (with a funciton "makeSrcloc()" in runtime.js) while some don't. So that makes me think I should potentially move "combinePyretPos()" into runtime.js and fix the rest of the error funcitons.

blerner:
No, that won't work, because parse-pyret still needs this functionality to exist on its own.

InnPatron:
Alright then. Anything else?

PR #1388, opened 2018-08-05
Tag "UNTERMINATED-STRING" tokens with the type of quote

InnPatron:
Tag "UNTERMINATED-STRING" tokens with the type of quote for generating better diagnostics. This is a backwards compatible change. Currently, unterminated triple quoted strings, double strings, and single quoted strings produce the same token which may cause some problems trying to guess intents. 

For instance:

```
"foo
bar
baz"
```

would potentially produce a different hint than:

```
\`\`\`foo
bar
"baz
```


blerner:
I don't think this is needed: the first character of the token's value is the opening quote character... Also, I don't see how your second example suggestion is supposed to work. Part of the value of tqs is that they allow other quotes inside them.

PR #1393, opened 2018-08-17
Expose num_to_roughnum from runtime

ZacharyEspiritu:
This PR exposes the `num_to_roughnum` function on the Pyret runtime.

This function is heavily used in the TensorFlow addition (see brownplt/code.pyret.org#287), but should be relatively easy to swap out if there are better alternatives. There might be a better way to create Roughnums from a JavaScript library that I'm missing, so let me know if this is unnecessary.


jpolitz:
This is fine to expose and makes sense to have it conveniently on runtime.

PR #1396, opened 2018-09-01
Add parser-interface, a separate location for defining parser error passes

InnPatron:
Add parser-interface.js, a central location where various parser error passes can be clearly added or removed.

Written in Javascript to avoid the stack traces automatically reported by Pyret's 'raise' keyword.

There are currently no parser error passes implemented.

As a side note, I could not get the FFI helper getField() to work on the module. If there's something I missed let me know.


blerner:
(after talking with Joe, we agreed to close this for now)

PR #1427, opened 2018-11-06
add gcd operation?  are the signatures the way we want?

blerner:
Just adding a `num-gcd` method to Pyret since it's missing.  But do we need the _array_ argument to `jsnums.gcd`, since we never even called that function prior to this PR, and we're only calling it with a single argument here?  @ds26gte did you need that array for spyret or wescheme or some other racket support?


schanzer:
Awesome! In that case, it sounds like Ben can merge once test cases are added

schanzer:
Thanks, @ds26gte ! Ben, close this PR at your leisure.

PR #1432, opened 2018-11-29
Anchor list for-desugaring fix.

InnPatron:
Switch to parameter order for certain functions on lists and immutable lists to match for-desugaring.


InnPatron:
I appear to be desugaring for expression arguments in the wrong order...

PR #1451, opened 2019-05-23
Don't include redundant locs in wf-errors

### LINKED ISSUE #1436: Well-formed errors print redundant srclocs on the command line
The tiny program
```
check:
  nothing
end
```
Produces the following error message on the command line:

> Well-formedness: (at file:///home/blerner/pyret-lang/test1.arr:2:2-2:9) A standalone variable name probably isn't intentional. srcloc(file:///home/blerner/pyret-lang/test1.arr, 2, 2, 9, 2, 9, 16)

The extra `srcloc(...)` isn't needed.  There's a bunch of these `ED.cmcode(self.loc)` expressions in many of the wf-errors' `render-reason` methods.

BenMusch:
Fixes #1436

No tests included because the test suite seems to check that `C.wf-err`s are thrown, but not their specific contents. Happy to find a way include tests if y'all want them

Example of an error message now:

> Well-formedness: (at file:///Users/ben/src/pyret-lang/build/t.arr:2:2-2:9) A standalone variable name probably isn't intentional.

cc @blerner

### DIFF
--- a/src/arr/compiler/compile-structs.arr
+++ b/src/arr/compiler/compile-structs.arr
@@ -428,8 +428,7 @@ data CompileError:
     method render-reason(self):
       [ED.error:
         ED.paragraph([list: ED.highlight(ED.text("Well-formedness:"), [list: self.loc], 0), ED.text(" ")]
-            + self.msg),
-        ED.cmcode(self.loc)]
+            + self.msg)]
     end
   | wf-empty-block(loc :: A.Loc) with:
     method render-fancy-reason(self):


blerner:
You missed the other sentence in the bug report, though: "There's a bunch of these ED.cmcode(self.loc) expressions in many of the wf-errors' render-reason methods."  Take a quick look to spot other redundant ones, if you can.

BenMusch:
Will do. Is there an easy way to test `render-fancy-reason`? I noticed some that appear redundant there, but can't tell if they're potentially used to parse the errors in the UI, so I held off from changing those

blerner:
That's used within CPO online, rather than at the command line.

BenMusch:
As far as I can tell, all of the remaining duplicate loc's are in the `render-fancy-reason` method, so I tested code.pyret.org locally with the duplicate loc's removed and it seems to change the behavior substantially. A local version running my pyret is on the left, production version with both loc's is on the right:

Screen Shot 2019-05-24 at 1 40 25 PM

Maybe it's worth keeping the duplicate locs since it seems like it makes the UI nicer?

jswrenn:
Ideally we would use `cmcode` in the `render-reason` methods, too, and the CLI would display code snippets there too. Unfortunately, CLI error display rendering system doesn't any way to render snippets and it didn't seem trivial to add last time I looked into it, iirc.

In the absence of such a mechanism, any appearance of `cmcode` in `render-reason` (but _not_ `render-fancy-reason`) should be regarded as a bug.

safwansamsudeen:
Why wasn't this merged?

PR #1454, opened 2019-06-18
Add support for units of measure

BenMusch:
PR for the work demo'd [here](http://pyret-units.herokuapp.com/)

This adds support for units of measure in the runtime type system, but not for the static type-checker. There is currently no way to "define" a unit. Instead, the base unit values are effectively treated as strings and checked for literal equality

Most of it is complete to my knowledge, with some notes:
 - Runtime error messages are not complete in the render-fancy-reason method. They all call render-reason right now, but I plan on changing that
 - There are a handful of TODOs about numeric operations I think _might_ be able to support units, but wanted feedback on those before adding it


shriram:
Just wanted to say how very awesome this is. Thanks! (Been discussing it w/ @blerner .)

sorawee:
Totally agreed with Shriram. This is very cool.

Bug report: 

```
1%<u ^ 10000000000000001> == 1%<u ^ 10000000000000000>
```

returns true.

src/js/trove/parse-pyret.js
```diff
@@ -970,20 +970,24 @@
...
-          // TODO: Handle numbers here somehow
           return tr(node.kids[1])
         },
         'unit-atom': function(node) {
-          if (node.kids.length === 1) {
+          if (node.kids.length === 1 && node.kids[0].name === 'NAME') {
             return RUNTIME.getField(ast, 'u-base')
               .app(pos(node.pos), name(node.kids[0]))
+          } else if (node.kids.length === 1 && node.kids[0].name === 'NUMBER') {
+            if (node.kids[0].value !== '1') {
```
blerner: This would probably be better handled as a well-formedness error later, rather than a parse-error now.  Also, `1%<3>` currently triggers an internal error rather than any clean error, so there's some more cleanup needed here.

BenMusch: My only hold-up with making this a wf check is that we have to change `u-one(l)` into `u-num(l, n)`. Not a major issue, something just feels _weird_ about holding onto this value in the AST just for the sake of a wf check. Do you think it's worth holding onto?

BenMusch:
Thanks @shriram and thanks for the bug report @sorawee! Will push a fix for the bug tomorrow

sorawee:
Is there a predicate that distinguishes `1` and `1%<u>`? `is-number` seems to accept both.

BenMusch:
@sorawee You would have to use the `%<u>` annotation there, you can see some examples here: https://github.com/brownplt/pyret-lang/pull/1454/files#diff-31f9ba2796b29ca0fb036e1d44797b4c

There's no function like `has-unit(u)` since units are not first-class values.

src/js/base/pyret-grammar.bnf
```diff
@@ -130,9 +130,18 @@ id-expr: NAME
...
 
-num-expr: NUMBER
-frac-expr: RATIONAL
-rfrac-expr: ROUGHRATIONAL
+dim-expr: (LANGLE|LT) unit-expr (RANGLE|GT)
+# TODO(benmusch): Consider adding | (LANGLE|LT) NUMBER (RANGLE|GT)
+unit-atom: (PARENNOSPACE|PARENSPACE) unit-expr RPAREN
+  | NAME
+  | unit-atom CARET NUMBER
+  | NUMBER
```
blerner: Is this now ToDone? :)

BenMusch: Yep!

sorawee:
IMO, annotation is for assertion, while predicate is for control, so they are not really equivalent. But yeah, non-first class value seems to be the problem here.

blerner:
Having a predicate here would allow programs to be non-parametric in their handling of units. The only two sensible behaviors I'd like are either "this function handles exactly this unit" or "this function is parametric over units". Anything else would allow students to be inconsistent with their units (e.g. "darn it this time it had the wrong units, ok, lemme check for that and cancel them out", instead of "wait why do I have inconsistent units in the first place?"), which seems counterproductive...

BenMusch:
Took some screenshots to showcase what the different runtime errors look like. I hard-coded conditional branches to be true/false to cover all of the statements in the rendering functions. Sorry for the large images.

**Errors when stack locs and asts are availible:

everything-available

**When the ast is not available:**

without-ast


**When no source file is available:**

no-src-available

**When no stack loc is available:**

no-stack-loc


**When 'is-builtin` is true:**

is-builtin

blerner:
Looks good.  The main grammatical complaint is the Test 1 failures, where the last sentence just begins, "unit annotated ...", rather than "*The* unit annotated", or better yet, "The `<...>` unit annotated ..." (if you have the unit available and can render it)

PR #1461, opened 2019-07-17
Fixes 1399: adding arity checks and argument checks everywhere as appropriate

blerner:


blerner:
@jpolitz I think this is good to merge

PR #1466, opened 2019-07-23
Attempting to give a better error message for function calls with spaces in them

blerner:
I.e. `f ()`, `g (x)`, `h (y, z)`, etc.

This is *hacky* and not very robust: we detect unary cases as well-formedness errors, and everything else as "parse errors".  The well-formedness error has to attempt to unwrap the pieces it receives, to undo the builtin["trace-value"] wrapping.  The "parse errors" actually successfully parse, but complain anyway!  The grammar rules for handling 0-arity and 2+arity are carefully rigged to be disjoint from the unary case, so that we avoid ambiguous parses; I'm only *mostly* confident that they really are unambiguous.

image

### DIFF
--- a/src/arr/trove/error.arr
+++ b/src/arr/trove/error.arr
@@ -2467,7 +2467,8 @@ data ParseError:
             ED.text("Pyret thinks this code is probably a function header:")],
           ED.cmcode(self.a),
           [ED.para:
-            ED.text("Function header must not have space before the "),
+            ED.highlight(ED.text("Function headers"), [ED.locs: self.a], -1),
+            ED.text(" must not have space before the "),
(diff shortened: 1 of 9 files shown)


src/arr/trove/error.arr
```diff
@@ -2526,15 +2554,15 @@ data ParseError:
             ED.text("Pyret thinks the string ")],
           ED.cmcode(self.loc),
           [ED.para:
-            ED.text("is unterminated; you may be missing closing punctuation. If you intended to write a multi-line string, use "),
+            ED.text("is not finished; you may be missing closing punctuation. If you intended to write a multi-line string, use "),
```
blerner: Apologies; I didn't mean to include these two lines in this PR; they're unrelated.

schanzer:
🎉

Issue #1470, opened 2019-07-25
Roughnum overflow results in internal error


sorawee:
```
~1e+308 + ~1e+308
```

results in:

```
Non Pyret value: TypeError: Cannot read property 'throwDomainError' of undefined
```

This is the same problem as #819.

blerner:
Jeeeeeeez.  Now that I'm looking at js-numbers again, there are a *lot* more places that need errbacks.  (And because JS allows function calls of incorrect arity, we never noticed......)

blerner:
https://github.com/brownplt/pyret-lang/commit/2bc5e6643c814209feb4c62e78c2c5cbb1c23c2e is a first attempt at fixing this, but it's not thoroughly tested.

sorawee:
@blerner in the chart library I use a lot of these functions that you just add an extra argument. Should they be changed as well?

blerner:
That's quite possible, yes.  I *think* any of these functions could theoretically throw a numeric exception, so we need to start from all the exception-raising sites, and work outward to their callers, and outward from there, etc. until we get to either runtime.js or to another direct client of js-numbers.  It looks like: ide.js, image-lib.js, world.js, the updated image libraries, d3-lib.js, d3-lib-list.js, plot-lib.js, plot-lib-list.js, chart-lib.js, and output-ui.js all potentially need fixing.  (The worst culprit is `toFixnum`, which calls `_integerDivideToFixnum`, which is a `makeIntegerBinop`, which could call `numerator` and `denominator` and `Roughnum.makeInstance`, which each potentially use `errbacks`.  I think none of those uses are live code when *starting* from `toFixnum`, but nevertheless we don't want mixed-arity calls...)

ds26gte:
Are we still pushing fixes into the `horizon` branch these days?

blerner:
Yes.

blerner:
Well, actually, let's make them PRs based on horizon. There's a *lot* of
stuff in flux and I don't want horizon itself to churn too heavily without
Joe or me right now.

ds26gte:
Just saw your PR and tested it for the particular bug reported by @sorawee. You may ignore my other PR as yours subsumes it. 

Here are my initial notes (which are easily fixable):

- Make sure _all_ calls to `Roughnum.makeInstance` use `errbacks`, (except for trivial stuff like `Roughnum.makeInstance(1)` which are known not to error).

[...]

PR #1477, opened 2019-07-29
Roughnum +- should overflow properly. Add tests too.

ds26gte:
This addresses #1470.


blerner:
Dorai, you're missing the work I've already done on this. (The commit should be linked to from the bug report.) There are a *lot* more errbacks missing, and a lot more tests needed.

PR #1478, opened 2019-07-30
more errbacks fixes. Also, num-to-fixnum added.

ds26gte:
- makeIntegerBinop's onBignums arg could throw error, hence should take errbacks parameter.
- _integerDivideToFixnum: its onBignums throws error on fixnum overflow.
- Added errbacks arg for all calls to Roughnum.makeInstance.
- BigInteger’s toFixnum method can throw error on fixnum overflow.
- Added num-to-fixnum as a Pyret function. Added tests.

`num-to-fixnum`, even if not encouraged on a Pyret user, is a useful function to test that internal fixnum conversions throw the right errors as needed.

Converting big-integers with an exponent that is not convertable to fixnum should throw an error, but this is hardly a probable scenario, so I've left it alone. (Exponents much smaller than the max fixnum already bring Pyret to its knees for memory reasons.)


src/js/base/js-numbers.js
```diff
@@ -1080,7 +1080,7 @@ define("pyret-base/js/js-numbers", function() {
       if (typeof(n) === 'number') {
         n = makeBignum(n);
       }
-      return onBignums(m, n);
+      return onBignums(m, n, errbacks);
```
blerner: Adding this is good, but then make sure that every use of `makeIntegerBinop` has 3-argument functions, even if they don't use `errbacks`, so that we don't have multi-arity functions

src/js/base/js-numbers.js
```diff
@@ -1257,11 +1257,14 @@ define("pyret-base/js/js-numbers", function() {
     function(m, n) {
       return m / n;
     },
-    function(m, n) {
+    function(m, n, errbacks) {
```
blerner: E.g. this change is good, but then `_integerEquals` below  should have 3-arg functions

src/js/base/js-numbers.js
```diff
@@ -191,7 +191,7 @@ define("pyret-base/js/js-numbers", function() {
     var match = s.match(scientificPattern), mantissaChunks, exponent;
     if (match) {
       mantissaChunks = match[1].match(/^([^.]*)(.*)$/);
-      exponent = Number(match[2]);
+      exponent = Number(match[2]); //possible, but not probable, fixnum overflow!
```
blerner: This is worth handling.  Would it be easier to try using `fromString` here (which already can handle big numbers) and then checking the size of the result?

ds26gte: If the exponent is not a fixnum, I can make it into a bigint and create intermediate number strings accordingly to construct the larger bigint. This can be done (and errbacks aren't needed, because there is no algorithmically determinable overflow anywhere in the process) and I'll push soon. 

However, the reason I said it was improbable is because Pyret chokes on even moderately large _fixnums_ in the exponent position. So we never get to the point of being able to test if a bigint exponent works.

Try something as "small" as `9e999999` (six 9s in the exponent). The entire IDE (and web page) collapses under the weight of the underlying computation.

blerner: Granted; there are lots of numbers we can produce that collapse the browser due to their computational magnitude.  That *does not mean* that "improbable" is an acceptable failure mode.  If Pyret crashes the browser in its entirety, that's a better user experience than showing the user "errors prevented Pyret from showing the full message" -- the former says we're trying to do our best and the browser can't keep up, while the latter says we screwed up.  If a user types `1e999999999999999999999999999................999`, and we crash because `Number("999............999999999999999999")` screws up in some way (e.g. by returning `Infinity`), then that's not an ok response.  We need to at least create the number representing the exponent, and choose to raise a domain exception instead of letting the page crash.

As for errbacks not being needed: I'm not sure what you're arguing for or against here -- errbacks are needed in any function that transitively may call a function that may throw an error.  And for methods on our number hierarchy, if any of them might use errbacks, then all of them must take errbacks, because that's a part of their signature.  If _this particular_ helper function doesn't need errbacks right now, but also has JS-level failure modes, then maybe this function needs to be redesigned to not crash at the JS level, and that may entail adding errbacks.

ds26gte: I actually fixed the issue in my follow-up push (18f7e6d) to my PR. Exponents that are not fixnum are technically allowed. I am, however, unable to devise a meaningful test of this fix, at least on my computer, because of page collapse. Screwing up because of an `Infinity` is definitely not countenanced anymore.

I can certainly add `errbacks` to `expandExponent` (and that would entail adding it to `makeBignum` too), but right now, none of the functions called within the dynamic extent of either of these functions invokes `errbacks.throw*`.

I would have needed to add `errbacks` if I was having `expandExponent` raise error on a non-fixnum exponent, but the current fix addresses such exponents fully, and will give an answer _if your computer permits_. Please review that it is satisfactory: the change is localized to the body of `expandExponent`, with one helper function called `zfillBignum`

ds26gte: P.S. When I say non-fixnum exponents are now allowed, I meant of course exponents that are strings of digits only, not embedded exponent notation. We do not (intentionally?) address numbers such as `1e999e999` (i.e., with multiple `e`s).

ds26gte:
As old as my PR is, it seems to merge OK with pyret-lang#fix-1470, Pyret builds and tests work. So I'm making the merge (to the official `fix-1470` branch).

May need work to merge into horizon.

blerner:
This broke the build for that PR -- double-check and see what else needs fixing to fix that?

ds26gte:
`.travis.yml` has a `node_js` setting of 8.3.0.

Locally, `fix-1470` builds and tests ok with node version 19.8.1.  

Is it just a matter of updating `.travis.yml`'s `node_js` to 19.8.1?

ds26gte:
Copied `.travis.yml` from `horizon` . That was enough to push through the Travis tests.

sorawee:
This is still not deployed to https://pyret-horizon.herokuapp.com/editor, right? I want to confirm that #1470 is properly fixed, so that I can close it, but it looks like it's still breaking.

sorawee:
Also, from what @blerner said in the issue, it sounds like there should be fixes in the CPO repo as well?

blerner:
No, this isn't merged to horizon yet.  It's merged to https://github.com/brownplt/pyret-lang/tree/fix-1470, which does not yet have a PR for it.

We also don't yet have thorough tests for all of these (at minimum, we need a thorough code-review to make sure we didn't miss one errbacks location, such that all the others just vacuously pass `undefined` along...)

PR #1482, opened 2019-08-26
Module syntax

jpolitz:


jpolitz:
We are pretty satisfied with this. It needs QA so it's not ready to hit master just yet, but the remaining QA is likely best to do on horizon where we can easily test and share.

Issue #1480, opened 2019-08-05
Octal string escapes don't seem to work, should this slice from zero?


gfbee:
https://github.com/brownplt/pyret-lang/blob/2af42d6c8fd44d329f95b83141f5c777df52f44f/src/js/base/pyret-tokenizer.js#L65

blerner:
Thanks for noticing!  (It appears we haven't ever had any uses of these escape sequences to notice that this was busted...)  Yeah, it appears it doesn't need to slice at all.  If you'd like to open a PR to fix it, go ahead: you should base the PR off the horizon branch, and you should add tests to `tests/parse/parse.js`, e.g. `expect(parse("\\101")).toBe("A");` and others, to confirm that the escape sequences are all working properly.

gfbee:
I appreciate the invitation to fix, but will have to leave that for someone else.

jpolitz:
<img width="79" alt="image" src="https://user-images.githubusercontent.com/119891/89936344-59843480-dbc8-11ea-926e-8088f3ed6607.png">

PR #1484, opened 2019-09-09
fix: properly parse octal escape sequences

ayazhafiz:
Fixes #1480.

Currently, octal escape sequences are parsed incorrectly because the
sequence is evaluated only partially. The "\\" delimiter in an octal
sequence like "\101" is already siphoned off during escape sequence
matching, so it is enough to simply parse the match of the numeric
sequence "101".

### DIFF
--- a/src/js/base/pyret-tokenizer.js
+++ b/src/js/base/pyret-tokenizer.js
@@ -62,7 +62,7 @@ define("pyret-base/js/pyret-tokenizer", ["jglr/jglr"], function(E) {
       else if (esc === "\\") { ret += "\\"; }
       else if (esc[0] === 'u') { ret += String.fromCharCode(parseInt(esc.slice(1), 16)); }
       else if (esc[0] === 'x') { ret += String.fromCharCode(parseInt(esc.slice(1), 16)); }
-      else { ret += String.fromCharCode(parseInt(esc.slice(2), 8)); }
+      else { ret += String.fromCharCode(parseInt(esc, 8)); }
       match = escapes.exec(s);
     }
     ret += s;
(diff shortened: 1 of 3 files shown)


ayazhafiz:
> I think this one should be an error, since `\88` is not a valid octal number.

You're right. Should be fixed now.

Right now the test method is rather crude, we check the string representation of the entire AST. Please let me know if you'd prefer a tree visitor or tests specific to the tokenizer rather than the parser; the latter may be more appropriate in this case.

blerner:
Mm, that is slightly awkward, yes.  Seems to me that an easier way to do this would be a pair of tests: one that *only* tested parsing and only tested string literals (rather than let-bound string literals), and a second one that evaluated the string literal and checked that it had the desired value (since Pyret string expressions evaluate to JS strings, in the end).

ayazhafiz:
@blerner is there already infrastructure for testing the tokenizer? If not, is there documentation of setting up new test files? I would like to get some idea of how you all prefer to do that.

blerner:
Yup: look higher in this file, at https://github.com/brownplt/pyret-lang/blob/7656aa31430dabd81cb5e630258477ae23076f99/tests/parse/parse.js#L115-L131.  Basically, call `lex` and then examine the `toks` array.

ayazhafiz:
@blerner please take a look

blerner:
Have you run `make parse-test` on this?  (The parse tests don't run automatically in `make test`.)

ayazhafiz:
> Have you run `make parse-test` on this? (The parse tests don't run automatically in `make test`.)

Yep.

blerner:
Oh my bad; I misread the test code -- I got concerned by the single-quotes in the string literals, since Pyret always prints its string literals with double-quotes...but you're checking the output of the lexer, which preserves the quotes of the actual token.

Yep, this looks good, thanks!

jpolitz:
Thanks @blerner, @ayazhafiz !

PR #1487, opened 2019-09-23
horizon -> master

jpolitz:


src/arr/compiler/ast-util.arr
```diff
@@ -1159,9 +1159,38 @@ fun get-named-provides(resolved :: CS.NameResolution, uri :: URI, compile-env ::
                 vp.set(as-name.toname(), CS.v-alias(corrected-origin, origin-name))
               | s-local-ref(l, name, as-name) =>
                 vb = resolved.env.bindings.get-value-now(name.key())
+                # NOTE(joe/ben): The as-name below has important semantic meaning.
```
sorawee: Racket has features like `free-identifier=?` so `(provide x (rename-out (x y))) (define x 10)` and `(provide x y) (define x 10) (define y x)` are not equivalent. But without these features (which I think is the case for Pyret), I think they are equivalent?

jpolitz: The test we came up with was:

```
file1.rkt
---
#lang racket

(provide x (rename-out (x y)))

(define x 10)


file2.rkt
---
#lang racket

(require "file1.rkt")
(provide y)

file3.rkt
---
#lang racket

(require [rename-in "file2.rkt" (y x)])
(require [only-in "file1.rkt" x])

```

file3.rkt has an error if you define y separately, but no error with this scheme. The underlying identity in the static semantics is the original `define`, not the providing position.

jpolitz: FWIW, this matters for various accidental diamond-import patterns that show up incidentally in teachpacks, and in how we ended up wanting to share struct definitions between the typed and untyped image libraries.

Given Pyret's aggressive avoidance of name shadowing, _and_ the use of the module system to represent REPL entries, we didn't really have an option other than to think about these things. And it's still not perfect :-/

Issue #1525, opened 2020-09-07
type errors are broken


shriram:
This program
```
fun f(x :: Number):
  x / "3"
end
```
reports the following type error:
![image](https://user-images.githubusercontent.com/75030/92342772-ddc6bc00-f08f-11ea-848e-cca995057dd6.png)

Changing it to
```
fun f(x :: String):
  x / "3"
end
```
produces
![image](https://user-images.githubusercontent.com/75030/92342788-ee773200-f08f-11ea-81a2-4aa5ec511283.png)

It seems like somewhere the error-reporting mechanism is failing to recognize structures and is just printing them out literally?

jpolitz:
I think your diagnosis is right, and “failing to recognize” is really that the print-for-compiler-debugging and print-for-user ended up being the same thing on these types.

Should be fixed by https://github.com/brownplt/pyret-lang/pull/1527

PR #1527, opened 2020-09-07
make type check errors use the to-string() method directly

jpolitz:
It looks like constructor printing used to be overridden for the sake
of errors. However, this is silly because we want constructor printing
for spying on types and so on within the compiler.

So make the error rendering contexts use to-string explicitly.

Down the line this should be a richer renderable type struct, but
ED.embed does constructor printing as well so stringifying with
the existing stringifier is better.

@blerner @mkolosick can you do a brief review of this to make sure I'm not missing something obvious?

### DIFF
--- a/src/arr/compiler/compile-structs.arr
+++ b/src/arr/compiler/compile-structs.arr
@@ -1956,9 +1956,9 @@ data CompileError:
           ED.text("Type checking failed because of a type inconsistency.")],
         [ED.para:
           ED.text("The type constraint "),
-          ED.highlight(ED.text(tostring(type-1)), [list: type-1.l], 0),
+          ED.highlight(ED.text(type-1.to-string()), [list: type-1.l], 0),
           ED.text(" was incompatible with the type constraint "),
-          ED.highlight(ED.text(tostring(type-2)), [list: type-2.l], 1)]]
+          ED.highlight(ED.text(type-2.to-string()), [list: type-2.l], 1)]]
     end,
     method render-reason(self):
       {type-1; type-2} = if self.type-1.l.before(self.type-2.l): {self.type-1; self.type-2} else: {self.type-2; self.type-1} end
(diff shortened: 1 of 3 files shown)


jpolitz:
https://github.com/brownplt/pyret-lang/issues/1525

blerner:
I had commented (https://github.com/brownplt/pyret-lang/commit/b5dfaa321d443c40e409c1ccdd9ae9b3cc1a9dbd#r41383423) about this and committed (https://github.com/brownplt/pyret-lang/commit/92ec0c056e31b34adea64b86aa3172fea86f9f3e) a fix, and this PR seems to take a different approach.  (In any case, #1525 is effectively currently fixed on horizon, without this PR.)  I don't quite understand "However, this is silly because we want constructor printing for spying on types and so on within the compiler" though -- why?  I got the sense that since this method was simply commented out, that it was a straggler of a commit that didn't mean to be committed.

I'd be happier with the inverse of this PR, I think: make `to-string` default behavior be pretty-printing, and add a `.to-repr` method, or a `.debug-repr` method, if we really need the constructor-level printing.  The common case of rendering the error seems to be _way_ more common than the hypothetical case of examining types, just based on the size of the PR and the fiddliness of "did I fix all of them?"...

blerner:
If we do go with this PR, then you'll need to revert that commit, or else you won't get back your constructor printing, anyway...

jpolitz:
The issue with overriding `_output` is that it then becomes super annoying to get all the source location printing, etc back. We'd end up implementing a really big method to show the same information we get from the un-overridden torepr. Seems better to have a dedicated method for rendering for users (it can upgrade to a valueskeleton or similar in the future if needed) and still give us the _option_ of getting all the detailed source locations without rewriting a few hundred lines that need to be maintained just to get that back.

jpolitz:
To follow on to this discussion – I just ran into a situation where I wanted to print a `string-dict` that _contained_ types. That is really onerous to manage with a specialized non-`_output` method for errors. This convinces me that I think `_output` needs to be the constructor printing, and that `_output` should be overridden only to give more helpful constructor printing, not for different purposes.

jpolitz:
I'm clearing out old PRs – this discussion seems unresolved (I stand by my dict-of-types issue as a reason I really don't want to override constructor printing for the sake of getting strings in error messages)

PR #1537, opened 2020-10-17
Add new builtin file constants.arr and make PI available at global scope

jpolitz:


jpolitz:
I got the value from `Math.PI` in Javascript.

blerner:
Looks great -- would love to see one more test that does `include constants` or `import from constants: PI end` to ensure we got our shadowing right, and then LGTM

blerner (approved):
LGTM -- we can build off of this easily in the future

PR #1541, opened 2020-11-10
fix an issue where escapes ranged from A-f, capturing unintentional characters

jpolitz:


blerner (approved):
that was a silly typo!  I'm amazed it's persisted unnoticed for this long...

PR #1547, opened 2020-11-25
Skeleton of mutable identical dictionary using built-in Map

jpolitz:
This PR can be closed if it's not useful to have the discussion here, but I figured it's an easy way to talk about some of the issues/tests that we'd like to add.

### DIFF
--- a/src/js/trove/dictionary.js
+++ b/src/js/trove/dictionary.js
@@ -1,0 +1,100 @@
+({
+  requires:
+    [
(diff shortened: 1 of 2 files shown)


jpolitz:
@shriram this is something we talked about.

@blerner what do you think of this? I guess it needs a `keys-now` and likely `remove-now` to be truly useful.

I feel like there's enough reasons to have this, and it's straightforward given the existence of `Map`.

PR #1548, opened 2020-11-25
Add some checking to raw-array-build

jpolitz:
1. Make sure fractional sizes are not allowed
2. Make sure negative sizes are not allowed
3. Make sure sizes greater than what we can allocate in JS are not allowed

For 1, we can know this isn't affecting any programs that expect behavior like
floor(), because the current behavior treated fractions like 0 so nothing would
be allocated.

Ditto for 2.

For 3, I can see arguments for another (larger) size, potentially, but also I
don't think it's unreasonable for different platforms to have different limits,
and this is a pretty reasonable limit for JS. As in, I don't think we're going
to implement massive arrays atop our JS implementation, and if the language
moves on to another platform it can have a different limit there.

@blerner any objections?

### DIFF
--- a/src/js/base/runtime.js
+++ b/src/js/base/runtime.js
@@ -4057,7 +4057,7 @@ function (Namespace, jsnums, codePoint, util, exnStackParser, loader, seedrandom
 
     var check_array_size = function(name, size) {
       checkNumInteger(size);
-      checkNumPositive(size);
+      checkNumNonNegative(size);
       // NOTE(joe):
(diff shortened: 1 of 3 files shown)


jpolitz:
FWIW, this came up because I wrote something like `range(0, len / 2)` and got an empty list. So we _also_ know no one is relying on `range` having floor behavior. This would make that an error, and we could later come back to giving that function more functionality if we want. NB that Python's `range` errors on `range(0, 1.5)`.

blerner (approved):
LGTM

Issue #1553, opened 2020-12-21
Type-checking does not terminate


tdelv:
The following code does not terminate when run with "Type-check and Run":
```
cases (Option) some(1):
  | none => [list: none]
  | some(num) => [list: some(num)]
end
```

ashdawngary:
The type checker is not terminating when coalescing the two branches of the cases(Option) statement:
`none => [list: none]` has type `List<forall A . (Option<A> % is-none)> % is-link` indicating a list of Optionals that are none.  Since none does not have a type bound to it (none is a valid element for `Option<X>` for all X), the checker cannot put any constraint on the bounded type of the option.  

For the second branch `some(num) => [list: some(num)]` the branch has `List<(Option<T> % is-some)> % is-link` where `T` is the type of `num`.  This reduces to `List<(Option<Number> % is-some)> % is-link` when cross referenced with what is provided in the cases statement.

Between the two cases, the type of the output needs to satisfy both constraints. This answer should be `List<(Option<Number>)> % is-link`

When executing the block above in Type-check and run, the type checker gathers context and constraints but there is unterminated recursion within the solver.  Profiling `arr/compiler/type-check-structs.arr` and `arr/compiler/type-check.arr` reveals that `solve-helper-refinements` (the function that resolves the refinement constraint portion of the constraint system), is recursively calling itself and adding extraneous substitutions to the solution being accumulated while the constraint system is untouched.  For reference, the constraint system when coming in looks something like:

`constraint-system([tree-set: ?-1], [list: ],
[...]

PR #1565, opened 2021-01-12
type checker infinite recursion fix

ashdawngary:
Hi, could someone walk me through converting my test cases in [my analysis](https://github.com/brownplt/pyret-lang/issues/1553)  into test cases for the compiler?

Thanks,
Neel

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,4 @@ compiled/
 
 .pyret-parley.*.pid
 .pyret-parley.*.lock
+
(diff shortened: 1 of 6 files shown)


jpolitz:
Probably the best thing to do is add them as new files in https://github.com/brownplt/pyret-lang/tree/horizon/tests/type-check/good. That's the place to put files containing programs that need to be tested with the type-checker and should test.

If you have new tests that you think are interesting that should *fail*, you can add them in https://github.com/brownplt/pyret-lang/tree/horizon/tests/type-check/bad

`make type-check-test` will run these if you want to test offline. You don't need to register the tests anywhere, just add the files (the test runner searches for all .arr files in those directories).

jpolitz:
Nice, thanks!

jpolitz:
(Note for posterity; I merged a package-lock.json change here; I reviewed it and it's a stale js-stack-parsing library. I'm not totally sure why it was lingering, but it doesn't need to be there and the package-lock.json changes are only removals. @ashdawngary a super minor note that I should probably have noticed that before I merged, usually I would have done that separately.)

PR #1574, opened 2021-02-13
Fixes #1572: adds `num-{truncate, floor, ceiling, round, round-even}-{digits, place}` functions

### LINKED ISSUE #1572: Expose num-round-to
This comes up in virtually every DS class and PD:

"How do I tell Pyret to show me a number, truncated to show the first _n_ digits"?

It's trivial to add to the DS library, but I'm wondering why it's not in Pyret itself? Joe says there's already a helper function defined, but it just isn't exposed. Can we expose it?

blerner:


jpolitz:
Wow, nice. It's cool that the multiply-first patterns work because we have bignums and don't worry about overflow.

Another case where macros/abstracting over syntax would help.

Good tests.

blerner:
It works for bignums, but I didn't test _thoroughly_ for huge or tiny-magnitude roughnums...though on the other hand, they're already gonna be pretty rough, so maybe it's fine anyway?

PR #1599, opened 2021-06-13
Matrix Library Pull Request

AnirudhNarsipur:
Hi,

As was requested by the core Pyret team, William Sun and I wrote this library containing core linear algebra features. 

What The Library Does : 

- Core Linear Algebra Functions (Inverse , Determinant , Orthogonalization)
- Manipulate matrices and vectors 
- An advantage of Pyret's number system is that Inverse(A) * A = Identity. 

What The Library Does Not Do : 

- Handle Complex Numbers, Tensors 
- More Complex Linear Algebra / any sort of statistical functions (Eigenvalues ... ) 

Setup 

- There are two datatypes Matrix and Vector . A Matrix is a nxn 2D array of numbers while a vector is equivalent to an array of numbers
- Representation wise both matrices and vectors are a 1D Javascript Array. These are then appropriately wrapped to create Pyret objects 
- You create a matrix by :` [mat(2,2):12,4,5,6]` and a vector by `[vector(3):1,3,-42]`
- Generally all operations are functions. However basic arithmetic (+,-,*) are available as methods along with element access. These are also available as functions.

Naming Conventions

- Most functions are of the form` func-mat `such as `rref-mat` . However some functions are given by just their name. 

Phillip Blair's Library
A few years ago Phillip Blair wrote a matrix library (https://github.com/brownplt/pyret-lang/blob/8bf1d79c283d50a04daae1d3e4a7682e606c1c02/src/arr/trove/matrices.arr). Ours is equivalent , the critical difference being that it is Javascript rather than Pyret. You can find the modern version of Phillip's library compliant with 2021 (https://code.pyret.org/editor#share=1_UVB6S9ad3F9KRcrdNx3yrUyAw-rGdfo&v=1599623) 

This makes a huge performance difference. The Pyret version times out for matrices as small as 50x50 (takes > 5mins) whereas the Javascript version (ours) does not show any lag. 

Issues To Address Before Adding

- Mapping functions such as `build-mat` use `safeCall`s to avoid blowing up the stack. However Pyret decides at some time to just halt the computation. So : 
- build-mat(30,30,lam(i,j): (3 + i) / (5 + j ) end)  (Building a 30x30 matrix) works fine
- build-mat(40,40,lam(i,j): (3 + i) / (5 + j ) end) (Building a 40x40 matrix) returns [mat(40,40): ]

I'm unsure as to how to solve this while still keeping these functions safe. 

Other Things To Address

- Right now matrix output is displayed as a 1d List 
Screen Shot 2021-06-12 at 11 47 22 PM

- Matrices are keeping with Pyret principles immutable. Is a mutable version required or should matrices be by default mutable . 

Anirudh Narsipur


src/js/trove/matrix.js
```diff
@@ -0,0 +1,1310 @@
...
+            if (self.$w != other.$h) {
+                return runtime.ffi.throwMessageException("The width of the first matrix and the height of the second matrix need to be equal");
+
+            } else {
+                new_arr = new Array((self.$h * other.$w));
+                for (var i = 0; i < self.$h; i++) {
+                    for (var j = 0; j < other.$w; j++) {
+                        var elm = runtime.makeNumber(0);
+                        for (var k = 0; k < self.$w; k++) {
+                            elm = runtime.plus(elm, runtime.times(get1dElem(self, i, k), get1dElem(other, k, j)));
```
blerner, line 335: I'm not sure whether you want to be using `runtime.plus` et al, or whether you should use the jsNumbers library directly.  Any error messages you get will have bizarre call-stacks if you do it this way.

blerner: This triple-loop is a potential stall-point for a Pyret program.  If I have two big matrices, and I multiply them, the multiplication could take long enough that it stalls the browser.  You may need to chunk this up via `safeCall`s into smaller pieces.

src/js/trove/matrix.js
```diff
@@ -0,0 +1,1310 @@
...
+
+        //Get matrix dimensions as a tuple 
+        var getMatrixDims = runtime.makeFunction(function (self) {
+            runtime.ffi.checkArity(1, arguments, "dims-mat", false);
+            runtime.checkArgsInternal1("Matrix", "dims-mat", self, annMatrix);
+            return runtime.makeTuple([self.$h, self.$w]);
+        }, "dims-mat");
+
+        //get element at (h,c)
+        var getMatrixElms = runtime.makeFunction(function (self, h, c) {
```
blerner, line 367: What do `h` and `c` mean?  I can imagine row and column, x and y, width and height...but height and column don't seem to go together...

blerner, line 367: What do `h` and `c` mean?  I can imagine row and column, x and y, width and height...but height and column don't seem to go together...

src/js/trove/matrix.js
```diff
@@ -0,0 +1,1310 @@
...
+                return runtime.makeBoolean(true);
+            } else {
+                return runtime.makeBoolean(false);
+            }
+        }
+        //Is the matrix square
+        var isSquareMatrix = function (self) {
+            arity(1, arguments, "is-square-mat", false);
+            runtime.checkArgsInternalInline("Matrix", "is-square-mat", self, annMatrix);
+            if (self.$w == self.$h) {
```
blerner, line 534: `return runtime.makeBoolean(self.$w === self.$h);`

src/js/trove/matrix.js
```diff
@@ -0,0 +1,1310 @@
...
+                return runtime.ffi.throwMessageException("Matrix must be square");
+            }
+            for (var i = 0; i < self.$h; i++) {
+                res  = runtime.plus(res,self.$underlyingMat[get1dpos(i, i, self.$w)]);
+            }
+            return res;
+        }
+
+        //Returns the Reduced Row Echelon Form of the Matrix
+        var MatRref = function (self) {
```
blerner, line 555: Why is this `MatRref` and not `matrixReducedRowEchelon` or some other more standard-looking name?  (Capitalized names, certainly, indicate object constructors, not functions...)

src/js/trove/matrix.js
```diff
@@ -0,0 +1,1310 @@
...
+            arity(1, arguments, "rref-mat", false);
+            runtime.checkArgsInternalInline("Matrix", "rref-mat", self, annMatrix);
+            //Copy matrix into new one 
+            toret = duplicateMatrix(self);
+            var lead = 0;
+            for (var r = 0; r < toret.$h; r++) {
+                if (toret.$w <= lead) {
+                    return;
```
blerner, line 563: Have you tested all the possible control-flow paths through this function, meaning all of these early-return paths?

src/js/trove/matrix.js
```diff
@@ -0,0 +1,1310 @@
...
+                self.$underlyingMat[get1dpos(m, i, self.$w)] = get1dElem(self, n, i);
+                self.$underlyingMat[get1dpos(n, i, self.$w)] = temp;
+            }
+        }
+        /* Calculates LUP decomposition PA = LU where:
+         P is permutation matrix
+         L is lower triangular
+         U is upper triangular 
+         return tuple of (P,L,U,exchanges) where exchanges is the no of row swaps*/
+        var LUPMat = function (self) {
```
blerner, line 644: This function is completely untested...

src/js/trove/matrix.js
```diff
@@ -0,0 +1,1310 @@
...
+                }
+
+            };
+            let PLU = function () {
+
+                let dim = A.$h;
+                let i, j, k;
+                for (i = 0; i < dim - 1; i++) {
+                    // Find the max entry
+                    let maxIdex = getMaxEntry(i);
```
blerner, line 709: Did you really save much typing or readability by leaving out the `n` in `maxIndex`?

blerner (changes requested):
This is not ready yet. There are lots of missing or wrong annotations to be checked, incorrect logic in several places, some dubious stack-safety issues to be addressed, and more tests are definitely needed (especially since some functions haven't been tested at all).  My comments below are a _starting point_ for you; there are likely more issues to be addressed.  When you're dealing with annotations, stack safety, equality, etc, you have to cultivate a certain kind of "skeptical paranoia", as in, "What could a clueless or malicious user do, that would violate some assumption I made and cause my code to break?  What could an unfortunate scheduling in the browser do to force my code to pause at an inconvenient point?"  I'm happy to set up a meeting time to go over the code with you and review what needs fixing, if you'd like, but I do want you to go through the feedback below and try to categorize it into big clumps of related comments (e.g. "annotations missing or wrong", "tests missing or incomplete", "stack safety wrong", etc), and then go back through your code function-by-function and see which categories apply to each function.

Is there documentation to go with any of this code yet, or not yet?

src/js/trove/matrix.js
```diff
@@ -0,0 +1,1310 @@
...
+        }
+        /**
+        * 
+        * @param {Number} h 
+        * @param {Number} w 
+        * @param {Array} underlyingMat 
+        * @returns Pyret Matrix
+        */
+
+        function makeMatrix(h, w, underlyingMat) {
```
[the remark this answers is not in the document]
AnirudhNarsipur: This happens in all user exposed functions : see ` createMatrixFromArray`

src/js/trove/matrix.js
```diff
@@ -0,0 +1,1310 @@
...
+                if (!hasBrand(brandMatrix, other)) {
+                    return runtime.ffi.notEqual.app('', self, other);
+                } else if (!sameDims(self, other)) {
+                    return runtime.ffi.notEqual.app('', self, other);
+                } else {
+                    function equalsHelp(i) { 
+                        return runtime.safeCall(
+                            function(){return recEq.app(self.$underlyingMat[i],other.$underlyingMat[i])},
+                            function(result){
+                                if (runtime.ffi.isNotEqual(result)) {
```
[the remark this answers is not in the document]
AnirudhNarsipur: Resolved. I'm not using raw_array_fold because we want to terminate early if not equal.

src/js/trove/matrix.js
```diff
@@ -0,0 +1,1310 @@
...
+            } else {
+                new_arr = new Array(self.$l);
+                for (var i = 0; i < self.$l; i++) {
+                    new_arr[i] = runtime.minus(self.$underlyingMat[i], other.$underlyingMat[i]);
+                }
+
+                return makeVector(new_arr);
+            }
+        };
+        function makeVector(underlyingArr) {
```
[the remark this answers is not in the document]
AnirudhNarsipur: See  `createVectorFromArray`

src/js/trove/matrix.js
```diff
@@ -0,0 +1,1310 @@
...
+                }
+            }
+            return makeVector(arr);
+
+        }
+
+
+        function matrixInit(h, w) {
+            if (!(Number.isInteger(h)) || !(Number.isInteger(w)) || h < 0 || w < 0) {
+                runtime.ffi.throwMessageException("The provided width or height is invalid. Matrix dimensions need to be a positive non zero integer");
```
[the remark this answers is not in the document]
AnirudhNarsipur: Error Message is wrong

src/js/trove/matrix.js
```diff
@@ -0,0 +1,1310 @@
...
+                        nj = (j+1 ) % w ;
+                        ni = i ; 
+                        if (nj == 0 ) { ni++ ; }
+                        } ; 
+                        return helper(ni,nj) ; 
+                     },"build-mat"
+                ) ; 
+            }
+            
+            return makeMatrix(h, w, helper(0,0));
```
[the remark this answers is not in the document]
AnirudhNarsipur: For some reason it tells me raw_array_build is not a function when I call it ? 
```
 ans = runtime.raw_array_build(F(function (idx) {
                res =  f.app((idx - (idx % w) / w, idx % w));
                if(!jsnum.isPyretNumber(res)){
                    return runtime.ffi.throwMessageException("Function must return a number"); 
                }
            }), h * w);
            return makeMatrix(h, w, ans);
```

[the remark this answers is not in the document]
blerner: Add it to the group of bindings around line 6130

src/js/trove/matrix.js
```diff
@@ -0,0 +1,1310 @@
...
+                        if (nj == 0 ) { ni++ ; }
+                        } ; 
+                        return helper(ni,nj) ; 
+                     },"map-mat"
+                ) ; 
+            }
+            return helper(0,0); 
+        }
+        //Returns a matrix given by (n1,n2) in top left corner and (n3,n4) in bottom right 
+        var getSubmatrix = function (self, n1, n2, n3, n4) {
```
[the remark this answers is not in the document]
AnirudhNarsipur: I've handled the submatrix part. 

In our initial discussions, we were recommended to use a 1-d array representation. The idea was that with a 1-d array it would be easier to extend to Tensors (n-dimensional matrices ex : a matrix of dimensions(3,4,3)). 

I do agree there are a lot of brittle calculations involved right now. Do you think it would be best to switch to a 2d representation ?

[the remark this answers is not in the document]
blerner: I'm not sure whether the current design generalizes any better or worse than having a multi-dimensional array would.  The one thing that does work "trivially" right now is the `reshape` operation, since you don't actually do anything to the array of values.  But I don't know how reshape would generalize to n dimensions anyway.

My main concern about a bunch of these things is I'd prefer robustness over sheer speed: we need to be able to understand and maintain this code after you finish with it! :)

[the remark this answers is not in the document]
AnirudhNarsipur: I agree that the overhead of having a 1d representation,  overweighs the small benefits (`reshape,transpose`).   I can also see future maintainers having problems with it. 

So do we want to change it ? If required it can be done without much work

AnirudhNarsipur:
I've addressed all comments and increased test coverage. There are still 2 issues : 

- `build-mat(100,100,lam(i,j): i + j end)` times out. I think I've wrapped it in safeCalls appropriately as suggested. 
- I chunked up multiplication which works for the most part. However it's creating some strange behaviour in `exp-mat` which raises a matrix to a power by repeatedly multiplying it ( line 873): 
```
res = duplicateMatrix(self);
            for (i = 1; i < n; i++) {
                res = multMatrix(res, self);
            }
```
Callin`g exp-mat(a,3) for   a = [mat(2,2):1,~4,-5,6.53]` results in a roughnum overflow error. For some reason the loop instead of running just twice runs dozens of times and causes this error.

src/js/trove/matrix.js
```diff
@@ -0,0 +1,1299 @@
...
+        //Creates matrix of size (h,w) by applying f
+        var buildMatrix = function (h, w, f) {
+            arity(3, arguments, "build-mat", false);
+            runtime.checkArgsInternalInline("Matrix", "build-mat", h, runtime.Number, w, runtime.Number, f, runtime.Function);
+            posInteger(h, w);
+            return runtime.safeCall(
+                function () {
+                    return runtime.raw_array_build(F(function (idx) {
+                        res = f.app((idx - (idx % w)) / w, idx % w);
+                        if (!jsnum.isPyretNumber(res)) {
```
blerner, line 855: Not quite -- you need this test of the result to be in the continuation of `f.app`: basically, you have to assume that `f` might bail out at any point in its execution, and `safeCall` says, "ok, Pyret, look -- when you _do_ finally finish running `f`, you need to follow up by doing _this_."  So you need to say `safeCall(f.app(...), function(res) { if (!isNumber) { ... } else { return res } })`

src/js/trove/matrix.js
```diff
@@ -0,0 +1,1299 @@
...
+
+        }
+        //Calculates A^n 
+        var expMatrix = function (self, n) {
+            arity(2, arguments, "exp-mat", false);
+            runtime.checkArgsInternalInline("Matrix", "exp-mat", self, annMatrix, n, runtime.Number);
+            posInteger(n)
+            res = duplicateMatrix(self);
+            for (i = 1; i < n; i++) {
+                res = multMatrix(res, self);
```
blerner, line 879: If `multMatrix` uses safeCalls, then _callers_ of `multMatrix` must use it in `safeCalls`: you need to now wrap this in a `raw_array_fold` to do the looping the right number of times -- you can use a bogus array of e.g. `new Array(n)`, because you don't actually care about the contents of the array; you just care about the matrix-multiplication result.

(And if you want to get more efficient here, you should use [exponentiation by squaring](https://en.wikipedia.org/wiki/Exponentiation_by_squaring) to reduce the number of matrix-multiplications you have to do.

AnirudhNarsipur: Is this for every single call of `multMatrix`. That is in some function if I cal`l c = multMatrix(b,c)` does that need to be wrapped in `safeCall` now?

AnirudhNarsipur:
Addressed both comments and made similar fix in `map-mat`. I've left the other `multMatrix` calls  unchanged for now but my understanding is that each of them need to individually wrapped but where does this end ? If I have `f` calling `multMatrix` and `g `calling `f` . Do both `g` and `f` now need to be wrapped in `safeCalls`?

PR #1603, opened 2021-06-28
Allow shadowing included ids, and more configurable REPL

jpolitz:
1. A new check in resolve-scope before reporting shadowing errors that doesn't
count shadowing errors if they are between an id and an included id (except in
the case of REPL ids from definitions:// and interactions://, which are special
origins). This is bad in some ways because e.g. the beginner who writes `circle
= circle(50, "solid", "red")` will likely be confused. But it's worth seeing
what this looks like.

2. Allow clients of the repl library to pass in preferred extra imports when
constructing the definitions locator.

3. Allow ExtraImport s to be expressed as blanket include statements

### DIFF
--- a/src/arr/compiler/ast-util.arr
+++ b/src/arr/compiler/ast-util.arr
@@ -891,24 +891,30 @@ fun wrap-extra-imports(p :: A.Program, env :: CS.ExtraImports) -> A.Program:
          collision is acceptable or not.)
       |#
       l = A.dummy-loc
-      full-imports = for fold(lst from empty, i from imports):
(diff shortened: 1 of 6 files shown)


jpolitz:
@blerner this is what this looks like, with a corresponding little commit on CPO to use an `extra-include` for `image` when initializing the REPL.

jpolitz:
(The ExtraImports/REPL stuff may be worth splitting out anyway because it's just plain useful, without the `resolve-scope` change)

jpolitz:
Intentionally abandoning this idea in favor of a UX solution for `use context` in new files.

PR #1605, opened 2021-06-30
Use statements (namespaces)

jpolitz:
Add basic support for the `use` declaration, which specifies the full namespace a module starts in.

See https://github.com/brownplt/pyret-lang/discussions/1555

This is implemented as a simple desugaring:

```
use namespace foo
```

==>

```
include foo
```

Along with two other changes – the `standard-imports` typically used for all non-builtin modules is now _only_ used for modules without a `use` line. All other modules start in an empty namespace and populate it starting from the `use` line.

The `empty-namespace` module can be used to get a truly empty Pyret namespace, incapable of even basic tasks like testing.


jpolitz:
@blerner any objections to merging this into horizon (all tests pass, I get a clean bootstrap)? I want to be able to point others at it to feel out the interface/experience.

blerner:
I'm fine with it

PR #1607, opened 2021-07-13
Anchor desugar

manaspurohit:
translating desugar.arr over to typescript

Many variants of Expr don't need specific implementation in typescript since the map visitor will uniformly create nodes and go through their fields

global-names added to ts-ast so that pyret desugar and TS desugar will use the same global-names


manaspurohit:
All the tests started failing after that commit and I couldn't really figure out why so I ended up reverting and doing type annotations and consolidating string dict to map functions as separate commits to tease out the issue. Ended up learning that the order of the imports in requires needs to match the order of the parameters in the `theModule` function

jpolitz:
This is great!

I think one thing we could use more of on this branch (which maybe we can find in the old tests) is tests of currying stuff. That's where the most complicated logic is from the copy-over, and it should get thoroughly exercised.

I'm happy with this overall approach, though, and it's interesting that you got away with only need to call constructors and no more detailed helpers (and ones like `boLocal` were easy to move over).

jpolitz (commented):
Made an overall comment, looks good! Adding some tests next for the new currying code would be good.

tests/pyret/regression/named-arrow-ann.arr
```diff
@@ -0,0 +1,15 @@
+import file("../test-compile-helper.arr") as C
+import file("../../../src/arr/compiler/compile-structs.arr") as CS
+
+run-str = C.run-str
+output = C.output
+compile-error = C.compile-error
+
+check "desugaring of named arrow annotations works":
+  ans = C.get-compile-errs(```
+func1 :: ((var1 :: Number) -> Number) = lambda(var1): 5 end
```
jpolitz, line 10: This should be `lam(var1): 5 end`, FWIW.

Issue #1499, opened 2019-12-09
Duplicate record field names in type don't error, but do in values


jpolitz:
This program doesn't report an error:

```
type O = { x :: Number, x :: Number }
```

but it should. The related value has a well-formedness error:

```
{x : 10, x: 20}
```

![image](https://user-images.githubusercontent.com/119891/70475363-37f70a00-1a89-11ea-8c8e-c3729f41d31d.png)

blerner:
Can we just copy these lines (https://github.com/brownplt/pyret-lang/blob/horizon/src/arr/compiler/well-formed.arr#L755-L759) to a new method `a-record`?

safwansamsudeen:
Argh! I wanted to take this up, but I think it should actually be closed.

PR #1622, opened 2021-10-07
Ensure Record Type Fields Are Unique

mrh29:
I think this should address #1499 . I'm looking forward to diving in to the language more.


jpolitz:
This is great!

It would be a huge help if this could come with a test or two.

Would you be able to add a test for this around here, maybe as a new block?

https://github.com/brownplt/pyret-lang/blob/horizon/tests/pyret/tests/test-well-formed.arr#L247

mrh29:
Thanks for pointing me to that. I've updated the PR with some tests.

Issue #1633, opened 2021-12-06
Generated values in check blocks


dbp:
I'm trying to write down an example of randomized testing (e.g., fuzzing / quickcheck tests). Doing that works fine, but when tests _fail_, I haven't figured out a good way of actually seeing what data they failed on. 

i.e., if I do it as follows:

```
check:
  input = generate-input()
  my-fun(input) does-not-raise
end
```

(for a fuzzer)

and it finds a bug, I get a backtrace, but the backtrace doesn't tell me (as far as I can tell) what the actual values passed were.

I'm not sure of a good solution, though maybe someone has an idea? 

I tried adding a `spy: input end` statement, but unfortunately the output from `spy` isn't scoped to the test blocks, so _all_ the `input` values are shown at the top of the interactions, and then the results of the tests, which is not very usable. And since I don't think there is a way for me to catch the errors other than using `check` assertions, I don't think I can build a check combinator that selectively `spy`s (which would work better, though if there are many failures it still wouldn't be great) -- for tests that don't error, selectively spying would improve things (eta expand the predicate in `satisfies` and `spy` in the false branch).

jpolitz:
Good question. We should enhance the `satisfies` output to show the LHS when the RHS errors. Then this pattern would work:

```
fun my-fun(x): x + 1 end

check:
  fun helper(input) block:
    my-fun(input)
    true
  end
  inputs = [list: 1, "a"]
  for each(input from inputs):
    input satisfies helper
  end
end
```

Right now it just reports the RHS error, but the value is there for reporting. (Compare to just putting `false` at the end of `helper` and not calling `my-fun` at all if you want to see the value-based output).

dbp:
I started implementing this, but realized while doing it that the (perhaps intentionally undocumented?) `run-task` that the test runners use can be used to  do what I want:

```
include either
... 
fun error-free(f):
  lam(inp):
    cases(Either) run-task(lam(): f(inp) end):
      | right(_) => false
      | left(_) => true
    end
  end
end

check:
  inp = generate-inp()
  inp satisfies error-free(my-fun)
end
```

This produces a nice error, as it shows what the left side value is for which the predicate failed. 

(It does seem like this pattern should be supported, without using what I guess are internal libraries).

dbp:
Oh, I spoke too soon -- while that gives the input, it doesn't give the backtrace since it eats the exception, so not actually a perfect solution (it is, at least, better, as I can then run it myself in the interactions window).

[...]

PR #1634, opened 2021-12-14
Add reporting of LHS values when error occurs in RHS predicate

dbp:
This, at least partly, addresses #1633.

In particular, it adds a new case of `TestResult`, `failure-exn-satisfies-rhs`, which is used when it a satisfies test where an exception was trigger in the RHS. In those cases, we have a value on the LHS, so we should show it along with the exception. I didn't see tests of this kind of behavior, but please point me if there is a way to do it. The four cases that will trigger the new behavior are:

```
check:
  1 satisfies lam(v): raise("error") end
end

check:
  1 satisfies lam(v): raise("error") end because 1
end

check:
  1 violates lam(v): raise("error") end
end

check:
  1 violates lam(v): raise("error") end because 1
end
```

And the new error message looks like:

Screen Shot 2021-12-14 at 1 58 40 PM

I noticed, as I was doing this, that there may be other cases where there is more information that could be shown, e.g., if an error is triggered by the refinement of an `is` test, neither the left nor right value is shown, which it seems like they should be:

```
check:
  1 is%(lam(v1,v2): raise("error") end) 1
end
```

Of course, there are still more variations -- e.g., if the right side of an `is` test errors, the left side value could be shown, etc. Of course, _maybe_ there is a way to make this visible in the stack trace which would subsume all of this? Though with the stack manipulation that's going on, perhaps that is expecting too much.

### DIFF
--- a/src/arr/trove/checker.arr
+++ b/src/arr/trove/checker.arr
@@ -482,6 +482,52 @@ data TestResult:
           VS.vs-value(exn-unwrap(self.actual-exn)),
           VS.vs-value(self.exn-place)])
     end,
+  | failure-exn-satisfies-rhs(loc :: Loc, actual-exn, lhs-value) with:
+    method render-fancy-reason(self, maybe-stack-loc, src-available, maybe-ast):
+      if self.loc.is-builtin():
+        self.render-reason(true)
+      else if src-available(self.loc):
+        cases(Option) maybe-ast(self.loc):
+          | some(test-ast) =>
+            [ED.error:
+              on-right.test-preamble(test-ast),
+              ED.cmcode(self.loc),
+              ED.paragraph(
+                [list: ED.text("It did not expect the evaluation of the "),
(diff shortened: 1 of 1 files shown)


blerner:
Trying to reconstruct the original design logic: I _think_ we'd thought that "if a test case unexpectedly errors, that itself is a bug and is the most salient issue to present", and didn't really consider the case where a weird LHS input could cause a logic error in the RHS and trigger an unexpected error.  I think we might have to workshop the rendering of these error a bit -- your proposed rendering above is good in that the first thing you see is the red unexpected-error box, and the "given the LHS" part is secondary and less salient.

Given that design logic, I'm less keen on making `lhs is rhs-that-errors` render the lhs value, since there's no possible meaningful scenario where `rhs-that-errors` should error in an `is` test!  @jpolitz agreed?

blerner:
@jpolitz just skimming this again -- worth merging as is, or should we workshop it a bit more?

jpolitz:
I think this improves things, no reason to delay merging it.

PR #1655, opened 2022-08-16
Make table-from-columns work with empty lists

### LINKED ISSUE #1653: table-from-column errors
It's unclear to me why [`table-from-columns`](https://www.pyret.org/docs/latest/tables.html#%28part._tables_table-from-columns%29) cannot work with empty lists, for creating an empty table of the right shape that will be populated programmatically. The document does not indicate that this would not work. However, the code
```
[T.table-from-columns:
    {"a"; [list: ]},
    {"b"; [list: ]}
  ]
```
produces what is essentially an internal error:

image

3Rafal:
This should fix #1653

This is my first PR so I might be wrong, but here is the solution:
When creating array from a column name and values, we don't have to map values to raw rows and then create Table out of them. Instead, we can create empty table with one column and then add all rows to it.

### DIFF
--- a/src/arr/trove/tables.arr
+++ b/src/arr/trove/tables.arr
@@ -72,13 +72,16 @@ raw-row = {
   make5: lam(t1, t2, t3, t4, t5): builtins.raw-make-row([raw-array: t1, t2, t3, t4, t5]) end,
 }
 
+fun empty-table(col-names :: List<String>) -> Table:
+  for fold(t from table: ignore end.drop("ignore"),
+           c from col-names):
+    t.add-column(c, empty)
+  end
+end
+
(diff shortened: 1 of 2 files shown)


jpolitz:
Sorry for the delay in getting to this. I really like this! I like it that it's expressed with the existing operators rather than trying to do something too invasive with e.g. making the table syntax work with this.

I'm happy to merge this. In the long run I could see making it more efficient by tapping into the underlying table library a bit more directly, but that doesn't need to happen for this idea to get included.

It would be good to have a corresponding PR on the docs repo letting folks know about this, too: https://github.com/brownplt/pyret-docs/blob/horizon/src/builtin/tables.scrbl#L835

PR #1692, opened 2023-05-11
Proposed Examplar-related changes to anchor

ds26gte:
Proposed Examplar-related changes to anchor. 

github.com/ds26te/pyret-lang, branch anchor-mbro

### DIFF
--- a/ide/src/reducer.ts
+++ b/ide/src/reducer.ts
@@ -896,13 +896,12 @@ async function runExamplarAsync(state: State) : Promise<void> {
   const { dir, base } = bfsSetup.path.parse(currentFile);
   // eslint-disable-next-line
   const dirWheats = dir + '/wheats';
-  const wheats = fs.readdirSync(dirWheats);
+  const wheats = fs.existsSync(dirWheats) ? fs.readdirSync(dirWheats) : [];
   // eslint-disable-next-line
   const resultArray: any[] = [];
   let result: any;
   let failed = false;
-  // eslint-disable-next-line
-  for (let i = 0; i < wheats.length; i++) {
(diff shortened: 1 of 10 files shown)


jpolitz:
(Just a note that this is WIP, but useful to see the changes in one place as it evolves.)

ide/src/reducer.ts
```diff
@@ -881,6 +882,46 @@ async function runProgramAsync(state: State) : Promise<void> {
...
+  // (browser)fs doesn't have copyFileSync(). Simulate using {read,write}FileSync()
+  const data = fs.readFileSync(src);
+  fs.writeFileSync(dst, data);
+}
+
+async function runExamplarAsync(state: State) : Promise<void> {
+  const {
+    typeCheck, runKind, currentFile, currentFileContents,
+  } = state;
+  // eslint-disable-next-line
```
jpolitz, line 895: Side note as I'm reading this, best if we only use `eslint-disable` for lines that use `await` in loops. Otherwise best to fix the lint error.

ide/src/store.ts
```diff
@@ -148,7 +150,10 @@ function handleFirstActionableEffect(
   for (let i = 0; i < effectQueue.length; i += 1) {
     const effect = effectQueue[i];
 
+    // console.log('doing effect.effectKey ', effect.effectKey);
     switch (effect.effectKey) {
+      case 'startEditTimer':
```
jpolitz, line 155: What's a summary of the effect this has? Is it just that this event was being fired from somewhere so it needed to be handled somehow? Is that still happening in Examplaritor?

Issue #1693, opened 2023-06-13
Method not found when type checking `Table` stemming from incomplete set of methods in global provides


robertDurst:
Consider the following code:
```
tbl = table: name
  row: "Rob"
end

tbl.row("Foo")
```

When I execute without type checker all is well.
<img width="1507" alt="Screenshot 2023-06-13 at 12 45 32 AM" src="https://github.com/brownplt/pyret-lang/assets/16689974/edfd625c-a513-4441-9dfe-e31d52fe05e1">

With type checker, `.row` is not found.
<img width="1512" alt="Screenshot 2023-06-13 at 12 45 45 AM" src="https://github.com/brownplt/pyret-lang/assets/16689974/303177a5-5dff-44c9-b178-51f19669d37c">

I did quite a bit of digging and printing and I noticed that the fields on the `Table` data-type obeject resulting from `instantiate-data-type(obj-type, context)` were surprisingly limited:
```
t-data(Table, [list: ], [list: ], [string-dict: length, ( -> Number)], builtin(dummy location))
```

I expect many more methods than just this on `Table`. Eventually I landed on the global provides and noticed this:
```
"Table": ["data", "Table", [], [], {
    "length": ["arrow", [], "Number"]
}],
```

I reproduced locally by modifying `tests/type-check/good/table.arr`, then added the type signature for `row` to the global provides and it passed the type checker.

I noticed `String-Dict` and `Reactor` both "provide" their own data-types. Should `Table` do the same? 

*60 seconds of ctrl-f made me realize just moving the data-type from global provides to table provides isn't as simple as it might seem... it seems it will require resolving some dependency/module loading. Wanted to confirm the desired behavior before venturing further* :)

shriram:
I *think* the issue is that table type-checking has not really been touched in general. What *is* the return type of some of these methods? This is in some sense the entire research problem that inspired B2T2.

robertDurst:
> This is in some sense the entire research problem that inspired B2T2.  

*Ahh, yes. Well I have now just about gone full circle. B2T2 was my starting point.*

> What is the return type of some of these methods? 

*That is a good question. I was looking at some of the type signatures for `Table` methods in the docs that didn't quite line up with my understanding and, unironically, discussing this with @bennn last night.*

While the one _hardcoded_ method (`length`) is valid, is it worth moving this datatype to the `Table` module preceding the table type work or as a follow up?

[...]

Issue #1662, opened 2022-10-04
internal error on type check + tables


shriram:
This program runs fine without type-checking, but with it it produces an _internal error_ (as opposed to halting gracefully):
```

include reactors
include chart

data Pt1: pt1(x :: Number) end
data Pt2: pt2(x :: Number, y :: Number) end
data PtAcc: ptAcc(x :: Number, y :: Number, dy :: Number) end

UFO = scale(0.5, image-url("https://cs.brown.edu/~sk/tmp/ufo.png"))
BG = image-url("https://cs.brown.edu/~sk/tmp/uluru.jpg")

fun move(p :: PtAcc) -> PtAcc:
  ptAcc(p.x + 7, p.y + p.dy, p.dy + 1)
end


check:
  move(ptAcc(0, 25, 6)) is ptAcc(7, 31, 7)
  move(ptAcc(10, 47, 6)) is ptAcc(17, 53, 7)
end

fun show(p :: PtAcc) -> Image:
  place-image(UFO, p.x, p.y, BG)
end

r = reactor:
  init: ptAcc(0, 0, 1),
  on-tick: move,
  to-draw: show
end

t = interact-trace(r)

extend t using state:
  x: state.x,
  y: state.y,
end
```
The problem is entirely with the `extend`; until then everything works fine.

shriram:
This *could* be related to #1422?

Issue #1422, opened 2018-10-17
type-check internal error


buster-blue:
When I run the following code in type-check mode of the [online editor](https://code.pyret.org):
```pyret
songs = table: title, artist, play-count
  row: "Song 1", "Artist A", 11
  row: "Song 2", "Artist B", 22
  row: "Song 3", "Artist C", 33
end

ab-songs = sieve songs using artist:
  artist == "Artist B"
end
order ab-songs:
  play-count ascending
end
```
I consistently get this message, instead of a typical result:

> Internal errors prevented this error message from being shown. Please report this as a bug.

This code works without errors, including the usual type, when I run it in non type-check mode. I have commented-out code in the file too, but removing it didn't change the results.

blerner:
The "internal error" complaint is either an error-messages problem or a desugaring problem.  Specifically, desugar.arr introduces prim-app calls to `checkWrapTable`, which use the same srcloc as the table identifier inside it.  Then compile-structs, line 1112, dies because it's expecting an s-id, but gets an s-prim-app.

I'm not sure why there's an `unbound-id` error even being thrown, though, so there's a second problem at work, after this internal error is fixed.

PR #1696, opened 2023-07-16
[WIP] Begin Defining Type Signatures for Table Methods

robertDurst:
While type checking is limited until we have a proper Table type, we can start defining the type signatures for Table methods now in order to allow more programs to type check. These checks will initially be pretty shallow/limited.

This directly addresses a few bug reports:
* https://github.com/brownplt/pyret-lang/issues/1693
* https://github.com/brownplt/pyret-lang/issues/1662
* https://github.com/brownplt/pyret-lang/issues/1422


shriram:
Oooh, exciting!

PR #1698, opened 2023-08-06
Realpath not consistent between MacOs and Linux, so Branch on OS

robertDurst:
To get the Makefile to work with my Mac, I looked at the output of the commands using realpath on my linux machine and translated them to what worked on the Mac. Need to do a bit of reading/testing to make sure the commands are Apples-to-Apples, but a basic smoke check is that things now actually compile on the Mac while just using `make`  ✅

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@
 
 all: build parser runtime
 
+UNAME := $(shell uname)
+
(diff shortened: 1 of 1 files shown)


jpolitz:
Sorry this never got a reply, I just lost track of it. Is this still an issue/good fix?

I think I looked at it and didn't follow up because I _always_ build on a Mac and hadn't noticed any issues, but maybe I'm just missing something. Output from a failed build might help me understand if this is still of interest.

Issue #1716, opened 2023-11-02
Error messages about undefined identifiers -> Use variables instead


schanzer:
Teachers and students have _no idea_ what this error means:
<img width="501" alt="image" src="https://github.com/brownplt/pyret-lang/assets/77126/fbbcf2d6-944f-4472-88ea-c85127bdfe05">

In our Pyret meeting 11/2/23, we agreed to change "identifier" to the less-correct but more-practical "variable"

blerner:
I'm really not a fan of misusing `variable` in this context.  After talking to @schanzer , we are trying out "name" instead, in contexts that rephrase the sentences slightly to avoid awkward duplication of the word "name"/"named".

PR #1729, opened 2024-03-25
Reword error messages to avoid the use of "identifier",

blerner:
while also not conflating mutable and immutable "variables"..

### DIFF
--- a/src/arr/compiler/compile-structs.arr
+++ b/src/arr/compiler/compile-structs.arr
@@ -689,7 +689,7 @@ data CompileError:
           ED.text(" errored:")],
         ED.cmcode(self.loc),
         [ED.para:
-          ED.text("This name is reserved is reserved by Pyret, and cannot be used as an identifier.")]]
+          ED.text("This name is reserved by Pyret, and cannot be used in a definition.")]]
     end,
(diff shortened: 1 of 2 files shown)


blerner:
This PR should fix #1716 .

schanzer (commented):
@blerner overall this feels like a small-but-meaningful improvement for Bootstrap's use-cases. And after our call today, if we can accomplish this without degrading things for Pyret's other audiences I'd say that's a big win.

Question: would it be possible to offer a suggestion, at least if the srcloc comes from the Interactions Area? Something like "have you Run the program recently?"

blerner:
I'm torn on that suggestion -- the downsides of it being wrong (and causing the student to lose whatever experimentation in the REPL they had) seem much worse than the upside of getting it right ("wow how did you guess I hadn't done that in a while?!")

schanzer:
With Pyret's REPL history in place, this doesn't seem like such a destructive thing to me, but I won't fight you on this. :)

schanzer:
@jpolitz the wording change Ben makes here ("name" instead of "identifier") has the blessing of the entire curriculum team. Can this be merged in the next few weeks?

PR #1735, opened 2024-04-30
Introduce a stable-sort-by method and function 

### LINKED ISSUE #1695: Stable sorting of tables
Bootstrap:DS switched from method syntax to function syntax, and we're already seeing huge gains from teachers. Coupled with the Circles of Evaluation, they are _immediately_ more comfortable composing operations. One of the most common cases is composing sorting operations, for example "sort animals alphabetically by species, then by age". 

Our `order-by` method makes no guarantees about stability, of course, so teachers have been consistently surprised by the results! 

What's fascinating is that this hasn't been an issue for the past 7 years with method syntax -- even though it's the exact same behavior. That suggests that methods were either hard enough that teachers weren't composing them, or that they were mysterious enough that teachers didn't have any confidence their own expectations of how they should work.

Anyway, this has now bubbled up to the point where it's become a sticking point, so I'm wondering what our options are here. Pyret appears to walk the table from the bottom up when sorting, presumably for performance/memory reasons. Is that consistent enough that I could invert the process in the function that wraps the call to `.order-by`? Or would this require a change in the runtime itself?

blerner:
Fixes #1695 by revising tables to use this new `stable-sort-by` method instead.

### DIFF
--- a/tests/pyret/tests/test-lists.arr
+++ b/tests/pyret/tests/test-lists.arr
@@ -260,6 +260,35 @@ check "sort as a function":
       { name: "Joan", age: 43 }]
 end
 
+check "stable sort":
+  fun build-pairs(m, n):
+    raw-array-to-list(raw-array-build(lam(i):
+          raw-array-to-list(raw-array-build(lam(j): {i; j; (i * (m + 1)) + j} end, n))
+        end, m))
+      .foldl(append, empty)
+  end
+  
+  TOTAL = 1000
+  GROUPS = 250
+  
+  pairs = (build-pairs(GROUPS, TOTAL / GROUPS))
+  
+  fun pair-le(p1, p2):
+    (p1.{0} < p2.{0})
+  end
+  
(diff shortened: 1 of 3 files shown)


tests/pyret/tests/test-lists.arr
```diff
@@ -260,6 +260,35 @@ check "sort as a function":
...
+  
+  fun pair-eq(p1, p2): 
+    p1.{0} == p2.{0}
+  end
+  
+  stable-sorted = pairs.stable-sort-by(pair-le, pair-eq)
+  unstable-sorted = pairs.sort-by(pair-le, pair-eq)
+  for each(i from range(0, GROUPS)) block:
+    stable-sorted.filter({(p): p.{0} == i}) is pairs.filter({(p): p.{0} == i})
+    unstable-sorted is-not stable-sorted
```
jpolitz, line 288: This line can be moved out of the for loop (says Ben)

PR #1739, opened 2024-05-02
Fix 1681

blerner:


jpolitz:
Functions look good, but commits include stuff from stats changes

blerner:
@jpolitz force-pushed to remove the unwanted statistics changes from this PR

blerner:
@jpolitz @shriram note that at the moment, the new functions are not available at the command line, because all command-line-compiled programs default to `essentials2020` as the context, and I don't think we want to retroactively add functions to that context.  We should update the contexts available in pyret-lang, not just in CPO...

PR #1741, opened 2024-05-04
Partial attempt to fix #1740

### LINKED ISSUE #1740: Inconsistent error terminology for record-annotation mismatches
Consider a simple program:
```
type Point2D = { px :: Number, py :: Number }

fun next-pos(p :: Point2D) -> Point2D:
  { px : p.x + 1 }
end
```
Then consider two bad uses of that function, and the errors we get on the command line or in CPO:
* `next-pos(3)`
```
The run ended in error:
Expected to get `Object` because of the annotation at file:///home/blerner/pyret-lang/temp.arr:12:18-12:25 but got: 3
Pyret stack:
  file:///home/blerner/pyret-lang/temp.arr: line 16, column 0
```
image

* `next-pos({ x: 3, y: 5})`
```
The run ended in error:
The record annotation at file:///home/blerner/pyret-lang/temp.arr:12:18-12:25 failed on this value:
{x: 3, y: 5}
Because:
* Missing field `px`is required at file:///home/blerner/pyret-lang/temp.arr:10:17-10:29
* Missing field `py`is required at file:///home/blerner/pyret-lang/temp.arr:10:31-10:43
Pyret stack:
  file:///home/blerner/pyret-lang/temp.arr: line 16, column 0
```
image

Questions: 
1. Do we want the terminology to be "record" or "Object"?
2. We can't fix things at the command-line, since we don't have the `render-fancy-reason` available there.  But is it worth trying to special-case this error message and see if we can't get it to show "The Point2D annotation was not satisfied..."?  (I can get this to work when Point2D is defined in the same file as it's being used, but I can't seem to get my localhost instance to read a shared gdrive file, so I can't test yet whether it works across files...which is where it's most useful.)

blerner:


blerner:
See #1740 for discussion

jpolitz:
Dope.

PR #1745, opened 2024-05-17
More stat

blerner:
Creating a PR to make it easier to review the changes


src/arr/trove/statistics.arr
```diff
@@ -204,3 +230,62 @@ fun r-squared(x :: List<Number>, y :: List<Number>, f :: (Number -> Number)) ->
...
+  n1 = l1.length()
+  n2 = l2.length()
+  if (n1 == 0) or (n2 == 0):
+    raise(E.message-exception("t-test-pooled: input lists must have at least one element"))
+  else:
+    m1 = mean(l1)
+    m2 = mean(l2)
+    v1 = variance-sample(l1)
+    v2 = variance-sample(l2)
+    (m1 - m2) / (((((n1 - 1) * num-expt(v1, 2)) + ((n2 - 1) * num-expt(v2, 2))) / ((n1 + n2) - 2)) * num-sqrt((1 / n1) + (1 / n2)))
```
blerner, line 260: This formula is really hard to read.  I'd write either `num-sqr(...)` or `(... * ...)` everywhere you're using `num-expt(..., 2)`, and I'd add more local variables and/or line-breaks to read this; I can't visually tell if this formula is "right" or not.

src/js/trove/multiple-regression.js
```diff
@@ -0,0 +1,182 @@
+({
```
blerner, line 1: I appreciate how short this file is :)

src/js/trove/multiple-regression.js
```diff
@@ -0,0 +1,182 @@
...
+        M[i][0] = y_s[i];
+      }
+      return M;
+    }
+
+    function matrixTranspose(m) {
+      let numrows = m.length;
+      let numcols = m[0].length;
+      let mT = new Array(numcols);
+      for (let i = 0; i < numcols; i++) {
```
blerner, line 24: I can tell this loop is correct, but a strong naming pet-peeve: use `r` and `c` (for row and col) instead of `i` and `j`, since they're much easier to tell apart while reading, and it helps keep track that the matrix is in row-major order.

src/js/trove/multiple-regression.js
```diff
@@ -0,0 +1,182 @@
...
+      if (size === 0) {
+        return 0;
+      }
+      if (m[0].length !== size) {
+        return 0;
+      }
+      if (size === 1) {
+        return m[0][0];
+      }
+      if (size === 2) {
```
blerner, line 44: Here, and in _all_ the arithmetic below: doesn't this need to use the jsnums library, rather than assume you've got JS doubles?  Or, if you're _relying_ on `num-to-fixnum` below, is that a deliberate choice in the stats library to switch entirely to roughnums?  (If a user supplied a list of exact fractions, say, shouldn't the estimator also be exactly rational?)

src/js/trove/multiple-regression.js
```diff
@@ -0,0 +1,182 @@
...
+    function matrixCofactor(m, i, j) {
+      let size = m.length;
+      if (size === 0) {
+        return 0;
+      }
+      if (m[0].length !== size) {
+        return 0;
+      }
+      let mm = new Array(size - 1);
+      for (let ii = 0; ii < size; ii++) {
```
blerner, line 63: Oh, no, please no.  `iiijjjiijijijijjjiij` is just illegible ;-)

src/js/trove/multiple-regression.js
```diff
@@ -0,0 +1,182 @@
...
+      return sign * matrixDeterminant(mm)
+    }
+
+    function matrixAdjoint(m) {
+      let size = m.length;
+      let mc = new Array(size);
+      for (let i = 0; i < size; i++) {
+        mc[i] = new Array(size);
+        for (let j = 0; j < size; j++) {
+          mc[i][j] = matrixCofactor(m, i, j);
```
blerner, line 84: Would it not be more efficient to construct the transpose as you go, instead of needing to allocate a whole extra matrix?

src/js/trove/multiple-regression.js
```diff
@@ -0,0 +1,182 @@
...
+      let Y = new Array(num_mappings);
+      let x_s_len = false;
+      js_x_s_s.forEach(function(x_s, i) {
+        runtime.checkTuple(x_s);
+        let js_x_s = x_s.vals;
+        let x_s_n = js_x_s.length;
+        if (x_s_len === false) {
+          x_s_len = x_s_n;
+        } else if (x_s_n !== x_s_len) {
+          throw runtime.ffi.throwMessageException("multiple-regression: bad mapping");
```
blerner, line 141: Unhelpful exception message: at least tell the student that the lengths of `<whatever they are>` don't match?

src/js/trove/multiple-regression.js
```diff
@@ -0,0 +1,182 @@
...
+      });
+      js_y_s.forEach(function(y, i) {
+        runtime.checkNumber(y);
+        Y[i] = [runtime.num_to_fixnum(y)];
+      });
+
+      let XT = matrixTranspose(X);
+      let B = matrixMultiply2(matrixInverse(matrixMultiply2(XT, X)), matrixMultiply2(XT, Y));
+      let Bfunc = function(x_s) {
+        runtime.checkTuple(x_s);
```
blerner, line 159: Missing a check that the values in `x_s` are all numbers -- otherwise you'll get a really weird error trace when `num_to_fixnum` fails later on.

src/js/trove/multiple-regression.js
```diff
@@ -0,0 +1,182 @@
...
+          for (let k = 0; k < m1c; k++) {
+            x += m1[i][k] * m2[k][j];
+          }
+          mP[i][j] = x;
+        }
+      }
+      return mP;
+    }
+
+    function multipleRegression(x_s_s, y_s) {
```
blerner, line 121: For the non-statisticians among us, it would be helpful to include some comments explaining the formulas a bit and/or links to resources (e.g. https://en.wikipedia.org/wiki/Linear_regression#Least-squares_estimation_and_related_techniques) that define the formulas, otherwise seeing things like random `Xi[0] = 1` floating around just feels inscrutable.

tests/pyret/tests/test-statistics.arr
```diff
@@ -89,3 +113,12 @@ check "linear regression":
 
 end
 
+check "multiple regression":
+  x-s-s = [list: {4}, {4.5}, {5}, {5.5}, {6}, {6.5}, {7}]
```
blerner, line 117: This honestly feels useless, both as a test and as an API: a whole bunch of 1-tuples? really?  And it doesn't test any of the correlations or matrix math you're doing, so it doesn't actually test the interesting part of this functionality.  The similar-ish test that I _would_ be interested in is seeing that `multiple-regression(x-s-s, y-s)({8}) is-roughly linear-regression(x-s, y-s)(8)`, showing that in the single-variable case, the multiple-regression and the linear-regression models coincide.

tests/pyret/tests/test-statistics.arr
```diff
@@ -89,3 +113,12 @@ check "linear regression":
...
+  y-s   = [list: 33, 42, 45, 51, 53, 61, 62]
+  pf = multiple-regression(x-s-s, y-s)
+  pf({8}) is%(within-abs(0.001)) 73.3214
```
blerner, line 120: Use `is-roughly` instead of `is%(within-abs(0.001))`, if you can get two more digits of accuracy into the test...

blerner (commented):
Good start, feedback inline

src/arr/trove/statistics.arr
```diff
@@ -189,6 +210,11 @@ fun linear-regression(x :: List<Number>, y :: List<Number>) -> (Number -> Number
   end
 end
 
+fun multiple-regression(x_s_s :: List<List<Number>>, y_s :: List<Number>) -> (List<Number> -> Number):
```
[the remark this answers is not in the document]
ds26gte: Yes, both the input `<List<List<Number>>` and the output `(List<Number> -> Number)` are wrong as I forgot to update them when I moved to tuples. Curious neither the build nor the test-running flagged this error...

src/js/trove/multiple-regression.js
ds26gte: I've pushed some better naming. In this case (transpose), I'm not sure whether you want the `r, c` to range over the input rows/cols or the output ones (since they are exact flips of each other). I just picked one without much conviction.

src/js/trove/multiple-regression.js
ds26gte: Given real-life data, it is unlikely that the coefficients of the predictor function will be anything but roughnums, no? And in the very rare cases they end up being exact rationals, "mistaking" them for roughnums is not catastrophic.

blerner:
All of our "real life" data will be drawn from spreadsheets, and I don't
know whether our default is to make numbers be rough or exact when we
import them.

I don't think it's *bad* to say that estimators are always rough, but we
should make a conscious design decision about that. @schanzer, thoughts?

tests/pyret/tests/test-statistics.arr
ds26gte: I was hoping for our curriculum mavens to give me some of their data to use. Otherwise, I'll have to cook up a case and make really sure that the independent variables are really independent, or the multiple regression won't work (I probably need to throw an exception in such cases. I think this happens when the dividing determinant is roughly or exactly zero).

schanzer: Try this:

```
mr-sheet = load-spreadsheet("17t2LtdsIgKN-Iz2CUWjaQyDqPyXeObwGJsl9k7vR4SU")
mr-table = load-table: id, biking, smoking, heart-disease
  source: mr-sheet.sheet-by-name("data", true)
end
```

It's taken right from [this example page](https://www.scribbr.com/statistics/multiple-linear-regression/), which lets us compare to the output from R.

schanzer:
We don't coerce to roughnums unless we have to, so it's not safe to assume that all the numbers operated on by this function will be rough. 

As for the output, we already assume the coefficients will be roughnums.

tests/pyret/tests/test-statistics.arr
ds26gte: I've added a test for 2 independent variables. (Nothing natural or realistic -- just arbitrary numbers, tested it in Lua and made sure it goes through with the same result in Pyret too.)

Also, added a test confirming that mul-reg on 1 var produces the same result as lin-reg on the same var.

ds26gte:
It looks like for Bootstrap purposes, @schanzer will be finding out the coefficients of the generated predictor function. (This also appears to help hide the full effulgence of the function from the user, who would then have had to deal with the `list` type.)

Calculating the coefficients (viz., y-intercept, slope) for the single-independent-var case was easy enough by calling the function at a couple of x's, i.e., 0 and 1. However, doing something similar for the multiple-indep-var case is 

1.  painful
2.  unnecessary -- since we already went through that whole matrix morass to create precisely this vector of coefficients, before converting that information into a predictor function

So it stands to reason that it's probably a Good Thing that our implementation supply both this coefficient vector (as a tuple?) _and_ the predictor function. Or we can have two functions: one that supplies both coeff vector and predictor function, and one that is more traditional and supplies just the predictor function. (The second function internall calls the first.)

I suggest the name `multiple-regression-model()` for the first function, retaining `multiple-regression()` for the latter. They both would take the same parameters:  a list of tuples of numbers, and a list of numbers.

What say?

blerner:
Question: is this latest rewrite (of linear into multiple-linear) a good thing? It streamlines the code, yes, but also slows down the simple linear case by making it have a bunch of matrix and tuple overhead... I'm honestly torn on whether this is a net win.

shriram:
@ds26gte  benchmarked it and claimed it was actually faster to use the multiple-variable version?

blerner:
That would be pretty depressing if true -- it saves (if I've counted right) 1 Pyret-level `map`, and replaces it with O(n) tuple allocations and a bunch of matrix loops (all of which are 1 iteration long, granted, but still).  I don't see how that can/should be _faster_!

ds26gte:
The OG Pyret version does the arithmetic in Pyret. The new version does the same arithmetic but in JS.  There is an extra creation of singleton tuples, true, but there are also intermediate Pyret lists created in the Pyret version that are not needed in the JS version. The Pyret version also traverses the x- and y-lists multiple times (to convert them to roughnums; to dot-product them; to sum each of them, to sum the squares of the x-lists). It also calls the mean (as yet another Pyret function) on them, which is two more traversals that aren't represented in the function body.  

The Pyret version could possibly by tightened by avoiding repeated summing of the x-lists at least (I counted three unnecessary recalculations, including the mean. But even so, it's still doing a lot that's more efficiently done in JS.

ds26gte:
Here are some numbers to go with the time comparison between the OG Pyret implementation of `linear-regression()` and the newer one that implements it using `multiple-regression()`.

I used two different data-sets, one with 7 readings, the other with 14, to see if the time diff is dependent on data size. To avoid noise from the start-up time of Pyret itself, I ran the "find predictor and use it" program fragment inside a loop, for 10 iterations, and then for 10000 iterations. The difference in favor of using multiple regression shows up in the 10 iterations already. (It is not surprising that the 10000 iterations should improve the difference, as it shuts out the start-up noise even more, but it confirms that the difference in the 10-iteration case is not a fluke. The fact that there is a positive difference in the 10-iteration case shows that you don't need to have huge data sets to experience the improvement.)

The difference may not be earth-shattering, but it is clearly in favor of using multiple regression always, and using just the latter obviously translates to easier maintenance for our devs.

```
== Pyret impl, 7 data points, 10 iterations
real    0m1.852s
user    0m1.926s
sys     0m0.448s

== Using mul reg, 7 d, 10 i
real    0m1.849s
user    0m1.958s
sys     0m0.438s

== Pyret, 7 d, 10000 i
real    0m2.401s
user    0m2.721s
sys     0m0.449s

== Mul reg, 7 d, 10000 i
real    0m2.017s
user    0m2.290s
sys     0m0.447s

== Pyret, 14 d, 10 i
real    0m1.849s
user    0m1.956s
sys     0m0.433s

== Mul reg, 14 d, 10 i
real    0m1.834s
user    0m1.937s
sys     0m0.446s

== Pyret, 14 d, 10000 i
real    0m2.754s
user    0m3.041s
sys     0m0.490s

== Mul reg, 14 d, 10000 i
real    0m2.076s
user    0m2.396s
sys     0m0.450s
```

schanzer:
Thanks for the detailed analysis!

This more than does the trick for Bootstrap! I will leave it to wiser heads to determine whether or not it's ready to merge.

ds26gte:
I'm going to kick myself for saying this, but we (i.e., I) could in theory implement `linear-regression()` too as its own JS function in the trove JS file, and it can come to its result without using any matrix manipulation whatsoever (however minimal). That will probably be palpably faster than doing a multiple-regression of 1 independent var. 

If single-independent regression is used significantly often enough to merit its own dedicated fast(er) implementation, then this may be the way to go.

schanzer:
I hate to say it, but I think it's worthwhile.

Single-independent regression is overwhelmingly the most common case. It's literally 100% of the cases for Bootstrap right now, and when multi-independent comes online it's still going to be only for the most ambitious of teachers.

ds26gte:
PS:

1. I don't know why all of `arr/trove/statistics.arr` can't be moved to `js/trove/*.js` except for wrappers
2. And, furthermore, why all of `arr/trove/*.arr` can't be moved to `js/trove/*.js`

This should make many things faster, unless we want to be able to claim that "most of Pyret is written in Pyret itself".

schanzer:
I defer to others on that one, but that feels like it would be a separate PR anyway.

ds26gte:
Is the current state of the PR ok, or should I roll back the last change, the one that defines single-var regression as an instance of multiple-var regression?

blerner:
I thought your antepenultimate comment said you'd reimplemented single-variable in JS without matrices, and it was the fastest yet? (Except in the smallest of cases, which I'm not too worried about.)

My main concern is testing -- if you have two distinct implementations, you can test the one against the other. If there's only one implementation, then I'd like to see more tests for it to confirm it works well. (E.g. permuting the variables should permute the derived coefficients, but gives the same overall result. Or that detecting singular cases works properly.) I don't know enough stats to know what a thorough test suite should be. I'd feel better if the matrix math itself were testable, but we're deliberately obscuring that for now...

ds26gte:
The currently publicly visible PR contains a definition of `multiple-regression()` (using matrices, there's no other workable option) and re-defines `linear-regression()` as a call to `multiple-regression()` of one variable.  It is therefore a single implementation that always uses matrices, and therefore requires only one maintenance track.

I can probably write a math proof of why it translates exactly to the non-matrix implementation. The numerical agreement between the two is exact for the example we have.

I have privately implemented `linear-regression()` as its own implementation that uses JavaScript, but it is identical to the original in-Pyret definition.  Since redefinition in JS wasn't seen as useful, I'm keeping it under wraps for now.

blerner:
@ds26gte , is it mathematically correct to say that the multiple-regression you implemented is https://en.wikipedia.org/wiki/Ordinary_least_squares#Matrix/vector_formulation, and you're trying to solve ${\displaystyle \mathbf {X} {\boldsymbol {\beta }}=\mathbf {y}}$ by computing $\displaystyle {\hat {\boldsymbol {\beta }}}=\left(\mathbf {X} ^{{T} }\mathbf {X} \right)^{-1}\mathbf {X} ^{\{T} }\mathbf {y}$ which minimizes the error $\displaystyle S({\boldsymbol {\beta }})=\|\mathbf {y} -\mathbf {X} {\boldsymbol {\beta }}\|^{2}$?

If so, is it further mathematically correct to say that you're implementing semantically the same computation as https://en.wikipedia.org/wiki/QR_decomposition#Using_for_solution_to_linear_inverse_problems for the over-determined case?  That is, find a solution ${\displaystyle {\hat {\mathbf {x} }}}$ to the problem ${\displaystyle A\mathbf {x} =\mathbf {b} }$ which minimizes the norm ${\displaystyle \left\|A{\hat {\mathbf {x} }}-\mathbf {b} \right\|}$?  (There's a square in the first that isn't in the second, but minimizing the norm should minimize the square of the norm too...)

I ask because the tentative matrix library I'm finishing includes QR-decomposition and a method that purports to solve least-squares, and while I don't entirely understand the math in it, if it's equivalent to the work you've got here (while also being more general), that'll be good to know...

ds26gte:
I can confirm I'm doing **_exactly_** what you stated in your first paragraph.  (The **X** matrix in our case is not just rows of the given x-inputs, but tacks on a 1 in front, because we want our **B** to provide coefficients not just for each _x_ but also for the y-intercept. But that doesn't affect the correctness of the matrix equation you mentioned.)

The calculations in the second paragraph are, TBH, too complicated for me to follow.  I'll take it on faith that it minimizes the norm, but I am not sure that minimizing the norm is identical to minimizing the least squares. (It could well be. If you point me to a proof, I'm convinced. I fear, however, that this is like comparing variability based on mean of abs differences with one based on root mean squares, which are not equal.) In any case, it provides _a_ solution. The predictor function is obviously intentionally a heuristic, and thus may allow some wiggle room as to what it could be.

ds26gte:
P.S. Also, a minor point _in case it is puzzling_: In your formulation you have  X B = Y, whereas I have Y = B X, but all your rows are my columns and all your columns are my rows, so they're semantically identical.

blerner:
Joy: according to https://en.wikipedia.org/wiki/Overdetermined_system#Approximate_solutions, _your_ approach solves $\min_{\mathbf x}\lVert A \mathbf x - \mathbf b \rVert$, _without_ the squared norm, and _my_ approach minimizes $\| A x - b \|^2$ _with_ the square.  And according to the other two links above, it's exactly the other way around!

https://math.stackexchange.com/a/4307382 claims
> Note that any minimizer of the 2-norm (aka the Euclidean norm or p=2 norm) is also a minimizer of the square of the 2-norm, so finding a solution to $\lVert Ax-b\rVert_2^2=(Ax-b)^T(Ax-b)$ also solves the problem, as discussed above.
which indicates that both approaches work... (But I'm not sure I'm convinced by the implication direction here, unless I'm misreading it.)

Honestly I don't recall enough of linear algebra to know the difference, so if one of y'all knows a math teacher who _does_, or who has a textbook on it (more reliable than wiki or stackexchange articles!), that would be really helpful...

schanzer:
Speaking of Joy - I would email her. She's taught Algebra 2, AP Stats, and (I believe) calculus or precalculus.

src/arr/trove/statistics.arr
```diff
@@ -222,9 +224,15 @@ end
 
 # please see: https://online.stat.psu.edu/stat462/
 
-fun multiple-regression(x_s_s :: List<Any>, y_s :: List<Number>) -> (Any -> Number):
+fun multiple-regression(x_s_s :: List<List<Number>>, y_s :: List<Number>) -> (Any -> Number):
   doc: "multiple-regression"
-  MR.multiple-regression(x_s_s, y_s)
+  mx_X = MX.lists-to-matrix(x_s_s.map(lam(x_s): x_s.push(1) end))
```
blerner, line 229: This would be better written, I think, as 
```
intercepts = MX.col-matrix.make(raw-array-of(1, y_s.length())
mx_X = MX.lists-to-matrix(x_s_s).augment(intercepts)
```
instead of making tons of `link`ed lists, it'll directly make the raw arrays, and also make the mathematical structure more explicit.

src/arr/trove/statistics.arr
```diff
@@ -222,9 +224,15 @@ end
...
+  mx_Y = MX.lists-to-matrix(y_s.map(lam(y): [list: y] end))
+  B = MX.mtx-to-list(MX.mtx-least-squares-solve(mx_X, mx_Y))
```
blerner, line 231: Please confirm (if you haven't already!) that `mtx-least-squares-solve` computes the same results as your previous code did :)

src/arr/trove/statistics.arr
```diff
@@ -222,9 +224,15 @@ end
...
+  fun B_pred_fn(x_s :: List<Number>) -> Number:
+    math.sum(map2(lam(b_i, x_i): b_i * x_i end, B, x_s.push(1)))
```
blerner, line 233: Instead of making `B` be `mtx-to-list`, this feels like it ought to use `Vector`s and `dot`, or matrix-multiplication.  It's not _wrong_, it just feels lower-level than I expected.

blerner (commented):
Minor nits, and one testing request

src/arr/trove/statistics.arr
ds26gte: The `1`s should actually be in a new first column, so I tried inverting the arguments to the `augment`, but it doesn't seem to do the expected thing. Would you expect the following check to be "shipshape"? It doesn't for me. 

To check: add this to the end of `test-matrices.arr`, and do a `make test`.

I do note that `test-matrices.arr` already has some tests for `augment`. They seem to work. I don't know why this one fails.  Specifically, it throws away the last column and duplicates the penultimate column. Hope that immediately suggests what the bug is.

```
x_s_s = [list: [list: 4, 3], [list: 9/2, 2], [list: 5, 6/5], [list: 11/2, 9/2], [list: 6, 33/10], [list: 13/2, 10], [list: 7, 0]]
y_s = [list: 33, 42, 45, 51, 53, 61, 62]


intercepts = M.col-matrix.make(raw-array-of(1, y_s.length()))
mx_X0 = M.lists-to-matrix(x_s_s)
mx_X = M.mtx-augment(intercepts, mx_X0)

check "stats":
  mx_X is [matrix(7,3): 1, 4, 3,
                        1, 9/2, 2, 
                        1, 5, 6/5,
                        1, 11/2, 9/2,
                        1, 6, 33/10,
                        1, 13/2, 10,
                        1, 7, 0]
end
```

blerner: Found it; fix incoming.

blerner: See https://github.com/brownplt/pyret-lang/commit/eeef8a7a7235f688558bc3d0eeb3fe0c95c8f5fd

src/arr/trove/statistics.arr
ds26gte: Yes, `multiple-regression` has been rewritten to use matrices/vectors in the latest commit. The previous test I had for it in `test-statistics.arr` continues to work.

src/arr/trove/statistics.arr
ds26gte: Thanks! I've updated my PR accordingly.

src/arr/trove/statistics.arr
blerner: `vector(raw-array-from-list(the-list))` would be simpler, and cheaper, than going via an intermediate matrix

src/arr/trove/statistics.arr
```diff
@@ -180,49 +190,65 @@ fun mode-any(l :: List<Number>) -> Number:
...
-fun linear-regression(x :: List<Number>, y :: List<Number>) -> (Number -> Number):
-  doc: "returns a linear predictor function calculated with ordinary least squares regression"
-  if x.length() <> y.length():
+# please see: https://online.stat.psu.edu/stat462/
+
+fun multiple-regression(x_s_s :: List<List<Number>>, y_s :: List<Number>) -> (Any -> Number):
+  doc: "multiple-regression"
+  intercepts = MX.col-matrix.make(raw-array-of(1, y_s.length()))
+  mx_X = intercepts.augment(MX.lists-to-matrix(x_s_s))
+  mx_Y = MX.lists-to-matrix(y_s.map(lam(y): [list: y] end))
```
blerner, line 231: This would be cheaper as `col-matrix.make(raw-array-of-list(y_s))`

src/arr/trove/statistics.arr
ds26gte: Done. Thanks!

src/arr/trove/statistics.arr
ds26gte: Done². Thanks²!

src/arr/trove/statistics.arr
```diff
@@ -180,49 +190,65 @@ fun mode-any(l :: List<Number>) -> Number:
...
+       of numbers, or raises an error if the list is empty```
+  num-sqrt(variance-sample(l))
 end
 
-fun linear-regression(x :: List<Number>, y :: List<Number>) -> (Number -> Number):
-  doc: "returns a linear predictor function calculated with ordinary least squares regression"
-  if x.length() <> y.length():
+# please see: https://online.stat.psu.edu/stat462/
+
+fun multiple-regression(x_s_s :: List<List<Number>>, y_s :: List<Number>) -> (Any -> Number):
```
blerner, line 227: Mmm, this has been switched from tuples to lists, so the signature should be `List<Number> -> Number`, not `Any -> Number`, right?

src/arr/trove/statistics.arr
```diff
@@ -180,49 +190,65 @@ fun mode-any(l :: List<Number>) -> Number:
...
+  doc: "multiple-regression"
+  intercepts = MX.col-matrix.make(raw-array-of(1, y_s.length()))
+  mx_X = intercepts.augment(MX.lists-to-matrix(x_s_s))
+  mx_Y = MX.col-matrix.make(raw-array-from-list(y_s))
+  B = MX.mtx-to-vector(MX.mtx-least-squares-solve(mx_X, mx_Y))
+  fun B_pred_fn(x_s :: List<Number>) -> Number:
```
blerner, line 233: This is an opportunity to improve the error message here.  Create the vector, and check that its size matches B's size (which you can memoize and close over), or else `raise "This regression expects # variables, but received ## instead"`.  The alternative is going to expose the error message from `dot` (https://github.com/brownplt/pyret-lang/blob/horizon/src/arr/trove/matrices.arr#L196), which involves vocabulary that isn't meaningful to the students using this...

blerner: I expect @schanzer will wrap this functionality _anyway_ when generating the BS:* libraries.

You might also try to `raise(E.arity-mismatch(Srcloc.builtin("matrices"), B's size, x_s' size, false))` to raise a standard Pyret arity error, and see what that looks like.

blerner: I expect @schanzer will wrap this functionality _anyway_ when generating the BS:* libraries.

You might also try to `raise(E.arity-mismatch(Srcloc.builtin("matrices"), B's size, x_s' size, false))` to raise a standard Pyret arity error, and see what that looks like.

src/arr/trove/statistics.arr
```diff
@@ -241,3 +267,78 @@ fun r-squared(x :: List<Number>, y :: List<Number>, f :: (Number -> Number)) ->
   end
 end
 
+fun z-test(sample-list :: List, population-sd :: Number, population-mean :: Number) -> Number:
+  doc: "z-test"
```
blerner, line 271: These docstrings don't add any info; either fix or remove... (here and throughout the file)

src/arr/trove/statistics.arr
```diff
@@ -241,3 +267,78 @@ fun r-squared(x :: List<Number>, y :: List<Number>, f :: (Number -> Number)) ->
...
+    v1 = variance-sample(sample-list-1)
+    v2 = variance-sample(sample-list-2)
+    (m1 - m2) / num-sqrt((v1 / n1) + (v2 / n2))
+  end
+end
+
+fun chi-square(observed-values :: List, predicted-values :: List) -> Number:
+  doc: "chi-square"
+  math.sum(map2(lam(observed-value, predicted-value):
+                  num-expt(observed-value - predicted-value, 2) / predicted-value end,
```
blerner, line 342: Clearer to use `num-sqr(obs - pred)` here, rather than `num-expt`.  And I'd probably write this as

```
for fold2(sum from 0, obs from observed-values, pred from predicted-values):
  sum + (num-sqr(obs - pred) / pred)
end
```

for brevity and also to avoid the intermediate list

tests/pyret/tests/test-statistics.arr
```diff
@@ -49,18 +49,45 @@ check "numeric helpers":
...
+  variance-sample([list: 3]) raises "division by zero"
+  variance-sample([list: 1, 1, 1, 1]) is-roughly ~0
+  variance-sample([list:]) raises "empty"
+
   stdev-sample([list: 3, 4, 5, 6, 7]) is%(within(0.01)) num-sqrt(10 / 4)
   stdev-sample([list: 3]) raises "division by zero"
   stdev-sample([list: 1, 1, 1, 1]) is-roughly ~0
   stdev-sample([list:]) raises "empty"
 
+  z-test([list: 1, 2, 3], 1.58, 2.58) is%(within(0.01)) -0.6358
```
blerner, line 72: Nice, thanks.  I'm curious why using `is%(within(0.01))` and not give a few more digits and just use `is-roughly`?  No need to change (unless it's easy!)

src/arr/trove/matrices.arr
```diff
@@ -351,7 +352,7 @@ sharing:
   
   method to-vector(self) -> Vector:
     doc: "Returns the one-row/one-column matrix as a vector"
-    if ((self.cols == 1) <> (self.rows == 1)):
+    if ((self.cols == 1) or (self.rows == 1)):
```
blerner, line 355: Good catch, oof!  Can you add the missing two test cases for this method to confirm that both `[row-matrix: 1, 2, 3].to-vector()` and `[col-matrix: 1, 2, 3].to-vector()` work without boolean oopses?

blerner (commented):
Last few changes, I think, and then I'll merge this later today.

src/arr/trove/statistics.arr
ds26gte: Done³ in latest push

src/arr/trove/statistics.arr
ds26gte: Done³.  Used a basic `raise(E.message-exception(...))` as I couldn't get the other one to work and I don't see existing usage

src/arr/trove/statistics.arr
ds26gte: Done³. I have already noted (to @schanzer) that the function names should probably use `score` rather than `test`, but I'm going by the spec he got from his SME(s)

src/arr/trove/statistics.arr
ds26gte: Done³

tests/pyret/tests/test-statistics.arr
ds26gte: Done³

src/arr/trove/matrices.arr
ds26gte: Done³

src/arr/trove/statistics.arr
blerner: Probably fine -- after all, this function genuinely has an arity of 1, not an arity of N... :)

src/arr/trove/matrices.arr
blerner: Thanks.  In hindsight, I realized what _actually_ was wrong, and the test that _actually_ was needed, would be `[row-matrix: 1]`, since we were saying "is-row-matrix isn't equal to is-col-matrix", which would be incorrect when it's a 1x1 matrix...but no need to add that test for now; I'm satisfied this is obviously less dumb than the code I'd checked in :)

blerner:
LGTM, can merge if this CI build goes through.

PR #1749, opened 2024-06-06
New matrix library

blerner:


tests/pyret/tests/test-matrices2.arr
```diff
@@ -0,0 +1,862 @@
+include matrices
```
jpolitz, line 1: Should we add a comment or comments that link back to Racket sources?

tests/pyret/tests/test-equality.arr
```diff
@@ -24,13 +24,23 @@ check "numbers":
   roughly-equal(3, 3) is true
   roughly-equal(3, 3.000003) is true
   roughly-equal-now(3, 3.000003) is true
-  roughly-equal-now(3, 3.000004) is false
```
jpolitz, line 27: Should this be factored out into a separate PR?

blerner:
Note: need to update checker library to use within instead of within-rel, to match intended new behavior

tests/pyret/tests/test-equality.arr
blerner: _maaaaaaaaaaaaaybe_, but I included it here because getting any sort of roughly-equals behavior on matrices _really needs_ a better notion of "roughly equal to zero": you simply can't say something is "roughly the identity matrix" without having a more tolerant notion of "roughly equal to zero".

PR #1754, opened 2024-08-27
Resurrect JS-based multiple-regression lib for now

ds26gte:
This implements the `multiple-regression()` function in `statistics.arr` using a JS-based library. The existing matrix-lib-based function was failing on an example from @retabak that used large data.

### DIFF
--- a/src/arr/trove/statistics.arr
+++ b/src/arr/trove/statistics.arr
@@ -31,6 +31,7 @@ import error as E
 import math as math
 import string-dict as SD
 import matrices as MX
+import multiple-regression as MR
 
 fun mean(l :: List<Number>) -> Number:
   doc: "Find the average of a list of numbers"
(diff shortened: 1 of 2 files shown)


src/arr/trove/statistics.arr
```diff
@@ -242,6 +243,11 @@ fun multiple-regression(x_s_s :: List<List<Number>>, y_s :: List<Number>) -> (Li
   B_pred_fn
 end
 
+fun multiple-regression(x_s_s :: List<Any>, y_s :: List<Number>) -> (Any -> Number):
```
blerner, line 246: The `Any`s in this signature are wrong, correct?  They should both be `List<Number>`?

blerner (approved):
It's fine; can merge.

retabak:
Hi! Should I be able to use this on horizon? I'm still just getting spinning, spinning, spinning with a 500-row dataset (the same one I attached to my email).

Issue #1758, opened 2024-09-09
table construction: internal error, and lack of guard


shriram:
This program works fine:
```
include tables

[table-from-rows:
    [raw-row: {"A"; 5}, {"B"; 7}, {"C"; 8}],
    [raw-row: {"A"; 1}, {"B"; 2}, {"C"; 3}]]
```
but a natural bug is to produce this instead:
```
[table-from-rows:
  [list:
    [raw-row: {"A"; 5}, {"B"; 7}, {"C"; 8}],
    [raw-row: {"A"; 1}, {"B"; 2}, {"C"; 3}]]]
```
(e.g., through `map`).

Not only does this not produce a graceful error, it says to report a bug:

![image](https://github.com/user-attachments/assets/b73ab200-8a91-4553-9448-b96d49ade43c)

ds26gte:
Please review that PR https://github.com/brownplt/pyret-lang/pull/1759 does what's needed.

PR #1759, opened 2024-09-09
Add annotation for table-from-raw-array(), checking arg is indeed a raw array of rows 

ds26gte:
addresses issue #1758

### DIFF
--- a/src/arr/trove/tables.arr
+++ b/src/arr/trove/tables.arr
@@ -1,5 +1,5 @@
 provide:
-  * hiding (is-kv-pairs),
+  * hiding (is-kv-pairs, is-raw-array-of-rows),
   type *
 end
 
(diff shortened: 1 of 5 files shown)


jpolitz:
A few tests here would help: https://github.com/brownplt/pyret-lang/blob/horizon/tests/pyret/tests/test-tables.arr#L653

You can see some similar style of tests (for dynamic annotation-checking errors in the table library) here: https://github.com/brownplt/pyret-lang/blob/horizon/tests/pyret/tests/test-tables.arr#L507

ds26gte:
Just one test added. Is there more variety needed in this situation?

blerner:
This looks like it merely tests that the argument is a raw array; it doesn't check that the elements are checked as well.

ds26gte:
@blerner, are you referring to having a test case that checks that something like 
```
  [table-from-rows:
    [raw-row: 5, {"B"; 7}, {"C"; 8}],
    [raw-row: {"A"; 1}, {"B"; 2}, {"C"; 3}]]
```
fails?

Or are you saying that the predicate I used in `tables.arr` is merely checking that the argument is a raw-array of raw-rows, but doesn't further check that each raw-row contains only key-value pairs?  The latter is already taken care of by the existing code once the initial check that it's all raw-rows passes.

I can add the test case above to `test-tables.arr`, but you already have in place a more stringent test, i.e., that the key-value pairs have matching keys in the correct order across all the rows. This new test is pretty basic in comparison.

blerner:
I'm saying you have no test case that fails in the `is-row` check of https://github.com/brownplt/pyret-lang/pull/1759/files#diff-9e511711584173e90288d8fe12034bea103c3de800fdd9b50fcf38dd0479b273R92.  So `[table-from-rows: 1, [raw-row: ...], false, [raw-row: ...], "a string that's definitely not a row", ...]`.

Additionally, implementing the annotation in Pyret as a refinement will give you a lousier error message than if you implemented it as an annotation in JS, as we do in the image library, as I pointed you towards in the original issue.

PR #1763, opened 2024-09-26
multiple regression functions with table argument

ds26gte:
statistics.arr: Added three functions that perform multiple regression given a table, the input cols, and the output col.
- multiple-regression-coeffs: returns list of coefficients
- multiple-regression-table: returns coeffs as 2-col table
- multiple-regression-fun: returns predictor function

### DIFF
--- a/src/arr/trove/statistics.arr
+++ b/src/arr/trove/statistics.arr
@@ -266,8 +266,9 @@ fun zip-num-lists(lol :: List<List<Number>>) -> List<List<Number>>:
   end
 end
 
-fun multiple-regression-coeffs(t :: Table, input-cols :: List<String>, output-col :: String):
-  doc: "returns a list of regression coefficients, given a table, a list of independent columns, and the corresponding output column"
+fun multiple-regression-coeffs(t :: Table, input-cols :: List<String>, output-col :: String) -> List<Number>:
+  doc: ```Input: a table t, a list of N input column names input-cols, and an output column name output-col;
+          Output: a list of regression coefficients```
   # spy: t, input-cols, output-col end
   x_s_s = zip-num-lists(input-cols.map(lam(col-name :: String): t.get-column(col-name) end))
   # spy: x_s_s end
@@ -282,15 +283,17 @@ fun multiple-regression-coeffs(t :: Table, input-cols :: List<String>, output-co
   MX.mtx-to-list(MX.mtx-least-squares-solve(mx_X, mx_Y))
 end
 
-fun multiple-regression-table(t :: Table, input-cols :: List<String>, output-col :: String):
-  doc: "returns a regression coefficient table, given a table, a list of independent columns, and the corresponding output column"
+fun multiple-regression-table(t :: Table, input-cols :: List<String>, output-col :: String) -> Table:
+  doc: ```Input: a table t, a list of N input column names input-cols, and an output column name output-col;
+          Output: a 2-column table of coefficients with column headers "coefficient-name" and "coefficient-value"```
   B = multiple-regression-coeffs(t, input-cols, output-col)
   coeff-names = input-cols.push("constant")
   [table-from-columns: {"coefficient-name"; coeff-names}, {"coefficient-value"; B}]
 end
 
 fun multiple-regression-fun(t :: Table, input-cols :: List<String>, output-col :: String) -> (List<Number> -> Number):
-  doc: "returns a predictor function, given a table, a list of independent columns, and the corresponding otuput column"
+  doc: ```Input: a table t, a list of N input column names input-cols, and an output column name output-col;
+          Output: a predictor function that takes a list of N input values and returns the predicted output value```
   B = multiple-regression-coeffs(t, input-cols, output-col)
   B_n = B.length() - 1
   fun B_pred_fn(x_s :: List<Number>) -> Number:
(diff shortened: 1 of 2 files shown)


src/arr/trove/statistics.arr
```diff
@@ -244,11 +249,61 @@ fun matrix-based-multiple-regression(x_s_s :: List<List<Number>>, y_s :: List<Nu
...
+  num-rows = lol.get(0).length()
+  for map(i from range(0, num-rows)):
+    lol.map(lam(l): l.get(i) end)
+  end
+end
+
+fun multiple-regression-coeffs(t :: Table, input-cols :: List<String>, output-col :: String):
+  doc: "returns a list of regression coefficients, given a table, a list of independent columns, and the corresponding output column"
+  # spy: t, input-cols, output-col end
+  x_s_s = zip-num-lists(input-cols.map(lam(col-name :: String): t.get-column(col-name) end))
```
blerner, line 272: This might be better done as `x_s_s = table-to-matrix(t.select-columns(input-cols))` using https://github.com/brownplt/pyret-lang/blob/horizon/src/arr/trove/matrices.arr#L1047 and https://github.com/brownplt/pyret-lang/blob/horizon/src/js/trove/table.js#L712

src/arr/trove/statistics.arr
```diff
@@ -244,11 +249,61 @@ fun matrix-based-multiple-regression(x_s_s :: List<List<Number>>, y_s :: List<Nu
...
+  # spy: x_s_s end
+  y_s = t.get-column(output-col)
+  # spy: y_s end
+  intercepts = MX.col-matrix.make(raw-array-of(1, y_s.length()))
+  # spy: intercepts end
+  mx_X = intercepts.augment(MX.lists-to-matrix(x_s_s))
+  # spy: mx_X end
+  mx_Y = MX.col-matrix.make(raw-array-from-list(y_s))
```
blerner, line 280: This could also be done as `table-to-matrix(t.get-columns([list: output-col])`, for symmetry's sake; it's not any more or less efficient, per se.

src/arr/trove/statistics.arr
```diff
@@ -244,11 +249,61 @@ fun matrix-based-multiple-regression(x_s_s :: List<List<Number>>, y_s :: List<Nu
...
+  # spy: mx_Y end
+  MX.mtx-to-list(MX.mtx-least-squares-solve(mx_X, mx_Y))
+end
+
+fun multiple-regression-table(t :: Table, input-cols :: List<String>, output-col :: String):
```
blerner, line 285: General comment: Didn't @schanzer implement these functions in his starter files somewhere?  (I'm pretty certain I gave him nearly this code, so it's worth double-checking...)

src/arr/trove/statistics.arr
```diff
@@ -244,11 +249,61 @@ fun matrix-based-multiple-regression(x_s_s :: List<List<Number>>, y_s :: List<Nu
...
+  doc: "returns a regression coefficient table, given a table, a list of independent columns, and the corresponding output column"
+  B = multiple-regression-coeffs(t, input-cols, output-col)
+  coeff-names = input-cols.push("constant")
+  [table-from-columns: {"coefficient-name"; coeff-names}, {"coefficient-value"; B}]
+end
+
+fun multiple-regression-fun(t :: Table, input-cols :: List<String>, output-col :: String) -> (List<Number> -> Number):
```
blerner, line 292: This function can't possibly be needed as written; don't we have `multiple-regression` already?  This looks like copy-pasted code, and fiddly code at that, so we should deduplicate it.

ds26gte: The difference is that the current trio of functions take a table, and column names as arguments. The Bootstrap group decided that all three outputs (viz., list of coefficients, table of coefficients, and predictor function) were needed, and it wasn't clear which one (or perhaps two) should be treated as basic.

blerner: That's not my point.  We already have https://github.com/brownplt/pyret-lang/blob/horizon/src/arr/trove/statistics.arr#L236-L244 that defines `B_pred_fun`; don't duplicate that code.

ds26gte: It's not a duplicate. `multiple-regression-fun` calls `multiple-regression-coeffs` to get a list of coefficients, which it calls `B`.  The `B_pred_fn` then uses `fold2` to do a dot-product its argument list with `B`.  It avoids having to convert the argument list to a vector.

In the original `multiple-regression`, we get a _vector_ of coefficients, which it calls `B`. Its `B_pred_fn` converts its argument list into a vector and then performs a vector dot product with the vector `B`.  

The only commonality between the two `B_pred_fn`s is that they check the length of their incoming list the same way to decide if an exception needs to be thrown. Once past that check, the body is different.

One could write an external function that is called in both places (with `B` as argument) to create `B_pred_fn`, but even so, one would have to do an extra list/vector casting in one of the two functions, so that the `B`s are the same type.

src/arr/trove/statistics.arr
ds26gte: Done, thanks!

src/arr/trove/statistics.arr
ds26gte: Done, thanks!

PR #1769, opened 2025-01-21
Start of CSV parsing with fast-csv 

jpolitz:
Using https://www.npmjs.com/package/fast-csv

Creating a PR to track this as I work on it.

Goals are:

- [x] CSV parsing directly to `RawArray<RawArray<String>>`
- [x] Table loader from CSV files
- [x] Table loader from CSV strings

### DIFF
--- a/src/arr/trove/csv.arr
+++ b/src/arr/trove/csv.arr
@@ -41,22 +41,12 @@ fun to-content<A>(csv :: RawArray<RawArray<String>>)
         | string-to-upper(val) == "TRUE" then: DS.c-bool(true)
         | string-to-upper(val) == "FALSE" then: DS.c-bool(false)
         | is-some(string-to-number(val)) then: DS.c-num(string-to-number(val).value)
-        | val == "" then: DS.c-empty
         | otherwise: DS.c-str(val)
       end
     end
   end
 end
 
-fun id-sanitizer(c :: DS.CellContent, row, col):
-  cases(DS.CellContent) c:
-    | c-empty => nothing
-    | c-str(s) => s
-    | c-num(n) => n
-    | c-bool(b) => b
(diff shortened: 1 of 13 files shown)


src/js/trove/csv-lib.js
```diff
@@ -0,0 +1,24 @@
...
+    requires: [],
+    nativeRequires: ['fast-csv'],
+    provides: {
+        values: {
+            'parse-string': ["arrow", ["String"], ["RawArray", ["RawArray", "String"]]]
+        },
+        types: {}
+    },
+    theModule: function(runtime, _, uri, csv) {
+        function parseString(str, opts) {
```
jpolitz, line 11: Need arity an record check on opts, str check on string

src/arr/trove/csv.arr
```diff
@@ -0,0 +1,94 @@
...
+       Assume this is the actual data of a table.
+       This generates the content from the values in the cell.
+       ```
+  for raw-array-map(row from csv):
+    for raw-array-map(val from row):
+      ask:
+        | string-to-upper(val) == "TRUE" then: DS.c-bool(true)
+        | string-to-upper(val) == "FALSE" then: DS.c-bool(false)
+        | is-some(string-to-number(val)) then: DS.c-num(string-to-number(val).value)
+        | val == "" then: DS.c-empty
```
jpolitz, line 44: Remove this line

src/arr/trove/csv.arr
```diff
@@ -0,0 +1,94 @@
...
+        | otherwise: DS.c-str(val)
+      end
+    end
+  end
+end
+
+fun id-sanitizer(c :: DS.CellContent, row, col):
```
jpolitz, line 51: Lift this out into data-source

PR #1772, opened 2025-02-14
Fixing pprint of AST

blerner:
This PR adds a few file APIs so that we can enumerate all the files in a directory, and then run them through a round-trip parse/pretty-print/parse cycle to ensure that our printing is consistent with our parsing.  And after some bug-fixes...It nearly is!  We don't quite render numbers faithfully to their original source-text: an input of e.g. `1.25` would get printed back out as `5/4`, and then get parsed as a fraction instead.  This is pretty minor, though, so this PR confirms that pretty-printing really does work correctly.


jpolitz:
So satisfying!

Issue #1776, opened 2025-03-18
Feature Idea: URL imports


jpolitz:
## URL Imports

It's useful to have imports that come from raw HTTPS urls for a number of reasons.

One particular use case that's come up recently is hosting files on Github: https://raw.githubusercontent.com/ urls are a useful place to host and fetch static code from a repository, which is already a good place to store code. This can be used to cut out publishing steps of going through Google Drive.

This proposal has a few steps aimed at different use cases and ergonomics

## Raw `url` imports

Add new syntax:

`import url(https-url) as I`

This expects:

- The URL is publicly available
- The URL replies to cross-origin CORS requests (githubusercontent satisfies this)

This will be implemented with a basic `fetch` call from the browser/from the command line that does a simple HTTP CORS GET request to the URL with no special headers or preprocessing. The response is expected to by _Pyret code_ (no importing raw JS via this mechanism). It is compiled and cached in memory (for CPO environments) and to disk in `compiled/` with a name based on the hash of its URL (for CLI environments). The identity of the module is the URL: if you import two modules with the same text with non-equal URLs, they are both instantiated separately. This does require a little care in not accidentally grabbing extra `?` or `#` fragments of URLs when copy/pasting, which could have spurious inequalities.

## `url-file` imports

One possible (and I hope to become common) workflow will be to author starter code and student libraries in VScode or in a Github Codespace, which has nice file management and version control built in along with the full-featured editor for reactors, etc.

One issue with authoring student-facing code meant to be used on code.pyret.org is that when publishing, it's useful to have a stable link in the import line. But while authoring, it's often useful to be able to make and test quick edits. `my-gdrive` and `shared-gdrive` are clumsy for this because they require edits and refreshes. The analog for development with `url(...)` in a codespace or vscode would be either:

- Frequently commit and push a library, wait for the push before re-running the student starter code that imports it with `url`. This requires pushing lots of little edits that are likely to have errors, and is cumbersome. Further, pushing to the canonical place of import could make users see these bugs.
- Switching to `import file(...)` while developing, and then switching back to `import url("https://raw.githubuerscontent...")` for a push that publishes. If multiple files are involved and _all_ of them need to be changed, it can be error-prone to push it out. Also it adds friction to making and testing a small edit.

What's needed is an import form that can be stable in source, but do “the right thing” depending on its context. We propose adding

`import url-file(base-url, rel-path) as I`

For example:

```
[...]

PR #1777, opened 2025-03-18
Url import

jpolitz:
Draft PR for implementation of url imports. Most immediate TODO is to figure out a plan for testing.

https://github.com/brownplt/pyret-lang/issues/1776


src/arr/compiler/cli-module-loader.arr
```diff
@@ -296,29 +297,58 @@ end
...
-        else:
-          raise("Cannot find import " + torepr(dep))
+        cases(Option) locate-file(ctxt, args.get(0)):
+          | some(located) => located
+          | none => raise("Cannot find import " + torepr(dep))
+        end
+      else if protocol == "url":
+        CL.located(UL.url-locator(dep.arguments.get(0), CS.standard-globals), ctxt)
+      else if protocol == "url-file":
+        full-url = args.get(0) + "/" + args.get(1)
```
jpolitz, line 333: There's a more principled way to do this, find a url path library.

src/js/trove/require-node-compile-dependencies.js
```diff
@@ -11,8 +11,8 @@ define("q", [], function() {return q;});
 jsmd5 = require("js-md5");
 define("js-md5", [], function() {return jsmd5;});
 
-canvas = require("canvas");
-define("canvas", [], function() {return canvas;});
+crossFetch = require("cross-fetch");
+define("cross-fetch", [], function() {return crossFetch;});
```
jpolitz, line 15: Oh this is confusing as a diff.

We don't need canvas in the _compiler_, just in the runtime libraries. So it's an accident that it was included here. But maybe confusing for this PR.

PR #1780, opened 2025-05-13
Starting a new `filesystem`/`filesystem-internal` library

jpolitz:
This is very much a WIP, not worth comments yet but I find keeping a PR open and up-to-date useful.

It is this, but unironically: https://xkcd.com/927/

The idea is to:

- Pick a small set of filesystem functionality we can support across VScode extensions, Node FS, browserFS, and whatever the next not-quite-a-real-fs platform is.
- Compile with a different filesystem-internal in different contexts (CPO will override this builtin import) that always has a consistent interface that is relatively easy to provide across contexts
- Allow `filesystem` to rely on that interface to provide any Pyret-level users with the desired functionality

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,7 @@
         "ascii-table": "^0.0.9",
         "benchmark": "^2.1.4",
         "browserify": "^17.0.1",
+        "buffer": "^6.0.3",
         "canvas": "^3.1.0",
         "command-line-args": "^3.0.5",
         "command-line-usage": "^4.0.1",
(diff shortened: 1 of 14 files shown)


jpolitz:
OK, this would be good to move on merging. There's more testing that would be nice to do.

This has the API that is needed for most work in the compiler, though it's not strictly necessary to do everything in the compiler with `filesystem`, this does use it in some places.

It has the common-denominator things we need to do sensible filesystem stuff in the extension (with the appropriate RPC calls and `fileystem-internal` setup in CPO), and nicely harmonizes that.

The `filesystem` API itself is the thing to publicly document for students, etc, which can work across many contexts.

PR #1783, opened 2025-06-07
fix tojson number error

ironm00n:
Currently providing an invalid number will produce an error indicating that both sides of `+` must be the same type:

```
Evaluating the Plus (+) expression at builtin://json-structs:76:12-76:25 errored. The left side was: Number  The right side was: 1/2 The + operator expects to be given:* two Numbers, or
* two Strings
```

I also wrote some basic unit tests for `tojson`, feel free to cherry-pick the other commit if that is unwanted.

### DIFF
--- a/src/arr/trove/json-structs.arr
+++ b/src/arr/trove/json-structs.arr
@@ -73,7 +73,7 @@ fun tojson(v :: Any) -> JSON:
     if num-is-fixnum(v) or num-is-roughnum(v):
       j-num(v)
     else:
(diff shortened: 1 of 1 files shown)


jpolitz:
Thanks! All useful.

We usually write tests like these in a separate file in `tests/` where they get picked up more explicitly by the CI. I'm merging to just accept quickly (since also the tests are fine), but if you want to do a PR for moving the tests great, otherwise I'll try to get to it.

Issue #1598, opened 2021-06-10
Type-checking `because` in a `raises` test results in a compiler error


molenzwiebel:
Simple example:

```
check:
  1 raises 1 because 1
end
```

This yields the following error:

```
The type checker rejected your program because the object type

{check-raises-satisfies :: Bot, check-satisfies-not :: Bot, check-is-not-refinement :: Bot, check-is-roughly :: Bot, check-is-refinement :: Bot, run-checks :: Bot, check-is :: Bot, check-raises-str :: Bot, check-satisfies-delayed :: Bot, check-raises-violates :: Bot, check-raises-not :: Bot, check-raises-other-str :: Bot, check-is-not :: Bot, check-satisfies-not-delayed :: Bot, check-satisfies :: Bot}

does not have a field named check-raises-str-cause
```

This seems to be because the check desugars to a method call, but one which is not actually in the signature of the `current-checker` type.

PR #1796, opened 2025-07-24
fix: type checking check operators with `because`

ironm00n:
fixes: #1598


ironm00n:
Let me know if the final commit--moving the tests--isn't desired or is too out of scope for this pr :P

PR #1802, opened 2025-08-13
Fix add-line overlay

### LINKED ISSUE #1801: Add-line seems to sometimes not render line at all.
See the following examples. 

Image

dbp:
The origin for the line should be at x1, y1, but since the arguments to makeOverlayImage move the _background_ right and down, we need to reverse both. This will mean that the background goes left and up and the start of the line is at the right place.

This fixes #1801.

### DIFF
--- a/src/js/trove/make-image.js
+++ b/src/js/trove/make-image.js
@@ -818,7 +818,7 @@
         var color = unwrapColor(maybeC);
         var img   = unwrapImage(maybeImg);
         var line  = image.makeLineImage(x2 - x1, y2 - y1, color);
-        return makeImage(image.makeOverlayImage(line, "left", "top", 0, 0, img, "left", "top"));
+        return makeImage(image.makeOverlayImage(line, "left", "top", -x1, -y1, img, "left", "top"));
       });
 
       f("scene-line", function(maybeImg, maybeX1, maybeY1, maybeX2, maybeY2, maybeC) {


dbp:
Turns out this isn't good enough; still are counter-examples :)

jpolitz:
I think this works the way I expect. It causes a crash with both negative x/y values on the command-line `node-canvas` – like literally a V8-internal assertion failure. But other than that I can't find an image that doesn't look like what I expect...

```
[17:57:59] (horizon *1)→ cat line.arr
include image
sq = square(25, "outline", "gray")
row = beside-list([list: sq, sq, sq, sq, sq, sq, sq, sq, sq, sq])
grid = above-list([list: row, row, row, row, row, row, row, row, row, row])

save-image( add-line(grid, -12.5, -10, 25, 10, "darkgreen"), "green-line.png")

[17:52:41] (horizon *1)→ node line.jarr


#
# Fatal error in , line 0
# Check failed: static_cast<int64_t>(amount_before) >= -delta.
#
#
#
#FailureMessage Object: 0x16dc056a8
----- Native stack trace -----

 1: 0x1023806cc node::NodePlatform::GetStackTracePrinter()::$_0::__invoke() [/usr/local/bin/node]
 2: 0x103b1c3d8 V8_Fatal(char const*, ...) [/usr/local/bin/node]
 3: 0x10274117c v8::internal::Heap::OldArrayBufferBytes() [/usr/local/bin/node]
 4: 0x1024fe1e4 v8::Isolate::AdjustAmountOfExternalAllocatedMemory(long long) [/usr/local/bin/node]
 5: 0x1022a9a6c napi_adjust_external_memory [/usr/local/bin/node]
```

dbp:
> Hrm save-image shouldn't crash with negative coordinates; we should choose which of two semantics to impose here:
> 
> * translate the rendered image so its bounding-box is all in Quadrant I
> * crop the image to Quadrant I
> 
> Images shouldn't _expose_ the fact that their coordinates might be negative

Is this a problem that was created by `add-line`, or is this a general issue with how `save-image` currently works? (as in, is it a part of this PR, or should it be a separate issue)

Issue #1803, opened 2025-08-15
`csv-table-url` should fail gracefully if given a local path (and probably, `csv-table-file` should similarly if given a url)


dbp:
Right now, if I put in:

```
csv-table-url("voters.csv", default-options)
```

I get the error:

<img width="554" height="94" alt="Image" src="https://github.com/user-attachments/assets/516775fb-3390-4410-9cc3-4f050e15f32c" />

Which uses technical jargon (404), and doesn't point to what they should do. 

We obviously can't rule out 404s, but if the string is not a URL, we should give an error that suggests they might mean to use `csv-table-file` if they are using a file rather than a url.

The other direction should be similar.

PR #1804, opened 2025-08-15
Improve errors for `csv-table-file` and `csv-table-url`

dbp:
I'm not actually sure how to test this, since my local testing strategy for stuff like this would usually use the local CPO, but CSVs are not usable through CPO. 

But, it's also a pretty minimal (and perhaps, not that great?) change. 

Either way, it would be good to have some improvement for #1803 before students start trying to use the `csv` stuff -- people bumped into this issue really quickly.


jpolitz:
I don't like this as an up-front check – I feel that the string-based checks are likely brittle when I think about the full space of paths and URLs and whatnot.

I wonder if it would make sense to move these checks into the failure case – we try to `fetch` or whatever, but if it _fails_, we add some information about why the failure happened.

jpolitz:
This is exactly what I had in mind. We should have a `file-to-string` that returns an `Either` because technically there's a TOCTOU race here if the target file is deleted. But given that the race only produces a worse error message I'm not going to lose sleep.

PR #1805, opened 2025-08-15
[Draft] Adding VS-Constr-Render

sidprasad:
- TODO - description


blerner:
I noticed this PR a few days ago, but I don't currently understand it (though I'm sure we discussed this design at some point, I just don't recall it right now) - is there an example of using this new construction?

src/js/base/runtime.js
```diff
@@ -202,9 +202,24 @@ function (Namespace, jsnums, codePoint, util, exnStackParser, loader, seedrandom
...
+      var renderers = thisRuntime.getField(val, "renderers");
+      return thisRuntime.hasField(renderers, thisContext);
+    }
+
     function renderValueSkeleton(val, values) {
       if (thisRuntime.ffi.isVSValue(val)) { return values.pop(); } // double-check order!
       else if (thisRuntime.ffi.isVSStr(val)) { return thisRuntime.unwrap(thisRuntime.getField(val, "s")); }
+      else if (thisRuntime.ffi.isVSConstrRender(val) && isInRendererContext(val)) {
+        var items = thisRuntime.ffi.toArray(thisRuntime.getField(val, "args"));
+        var strs = [];
```
blerner, line 216: These aren't actually guaranteed to be strings, though, right?  (It could be arbitrary HTML or something?)

Issue #1799, opened 2025-08-02
expt(7, expt(10, 36789)) evaluates to 1


schanzer:
`expt(7, expt(10, 36789))` run in the `starter2024` context produces this erroneous result, with no errors.

shriram:
If it helps:

I ran `expt(10, 36789)` and it printed a very large number.

To see how big this is, I ran `num-log(expt(10, 36789))`. This produced
```
TypeError: Cannot read properties of undefined (reading 'throwDomainError')
```
Note that the error isn't coming from *printing*, because this also produces the same error:
```
t = num-log(expt(10, 36789))
# expt(7, t)
```
even without running the commented-out line.

The threshold is between 10^308 and 10^309, so for instance
```
t = expt(10, 308)
num-log(t) / 2.303
```
produces `~307.9445109171368`, whereas
```
t = expt(10, 309)
num-log(t) / 2.303
```
produces the same error @schanzer reported.

@blerner will of course immediately point out that this range corresponds to the limit of an IEEE 754 64-bit double.

jpolitz:
Useful other context:

<img width="318" height="102" alt="Image" src="https://github.com/user-attachments/assets/1674651a-f744-45f3-b677-589a296a2039" />

<img width="187" height="35" alt="Image" src="https://github.com/user-attachments/assets/695288f1-f08a-4585-8a28-1b296b9c3b15" />

jpolitz:
I tried this:

```
> git diff
diff --git a/src/js/base/js-numbers.js b/src/js/base/js-numbers.js
index bd0fe9523..eefd608f5 100644
--- a/src/js/base/js-numbers.js
+++ b/src/js/base/js-numbers.js
@@ -3815,7 +3817,13 @@ define("pyret-base/js/js-numbers", function() {
   // expt: pyretnum -> pyretnum
   // Produce the power to the input.
   BigInteger.prototype.expt = function(n, errbacks) {
-    return bnPow.call(this, n);
+    const fix = n.toFixnum();
+    if(isOverflow(fix)) {
+      const base = Rational.makeInstance(this, BigInteger.ONE, errbacks);
+      const expt = Rational.makeInstance(n, BigInteger.ONE, errbacks);
+      return base.expt(expt, errbacks);
+    }
+    return bnPow.call(this, fix);
   };
```

because the `Rational` implementation of expt is a repeated-multiplication algorithm. Unfortunately it blows the stack when the exponent is Big.

```
[...]

PR #1807, opened 2025-08-22
Fix for #1799

ds26gte:
- Extra arg removed for BigInteger.prototype.{log,round,roundEven,tan,atan,cos,sin,acos,asin}
- bnpExp() should take errbacks arg and use it if exponent too large
- bnModPowInt() should take errbacks arg and pass it along to bnpExp()
- bnPow() should take errbacks arg and pass it along to bnpExp()

### DIFF
--- a/tests/jsnums-test/jsnums-test.js
+++ b/tests/jsnums-test/jsnums-test.js
@@ -1,0 +1,44 @@
+// To test: Build Pyret, then cd to this directory, and type
+// node jsnums-test.js
+
+var Jasmine = require('jasmine');
+var jazz = new Jasmine();
+const R = require("requirejs");
+var build = process.env["PHASE"] || "build/phaseA";
+R.config({
+  waitSeconds: 15000,
+  paths: {
+    "pyret-base": "../../" + build
+  }
+});
+R(["pyret-base/js/js-numbers"], function(JN) {
+  var sampleErrorBacks = {
+    throwDomainError: function() { throw 'domainError'; },
+    throwLogNonPositive: function() { throw 'logNonPositive'; },
+  };
+  describe("check exceptions in js-numbers methods that can't be tested in Pyret", function() {
+    it("bnpExp", function() {
+      // BigInteger.*.expt calls bnPow, which calls bnpExp
+      // shd raise exc for too-large
+      expect(function() { JN.makeBignum(2).expt(JN.makeBignum(0xffffffff + 1), sampleErrorBacks); }).toThrow('domainError');
+
+      // BigInteger.*.log
+      // should raise exception for arg <= 0
+      expect(function() { JN.makeBignum(-1).log(sampleErrorBacks); }).toThrow('logNonPositive');
+
+      // BigInteger.*asin
+      // should raise exception for arg ∉ [-1, +1]
+      expect(function() { JN.makeBignum(-1.5).asin(sampleErrorBacks); }).toThrow('domainError');
+      expect(function() { JN.makeBignum(+1.5).asin(sampleErrorBacks); }).toThrow('domainError');
+
+      // BigInteger.*acos
+      // should raise exception for arg ∉ [-1, +1]
+      expect(function() { JN.makeBignum(-1.5).acos(sampleErrorBacks); }).toThrow('domainError');
+      expect(function() { JN.makeBignum(+1.5).acos(sampleErrorBacks); }).toThrow('domainError');
+
+    });
+  });
+
+  jazz.execute();
+
+});


blerner:
@ds26gte , the fact that the tests above have passed, without this PR including any new tests at all, indicates a certain lack of testing coverage of the functionality being changes, yes?  At minimum, please add regression tests for every function whose behavior now _really ought_ to trigger a proper Pyret error, and make sure that it actually does so.

ds26gte:
I've now improved the definition of `num-log` so that 
```
num-log(num-expt(10, 36789)) -> ~84709.80298615796
```
Indeed
```
num-log(num-expt(36789, 36789)) -> ~386763.3734138913
```

(We can tighten these values considerably (using Racket as a comparison point), by changing the `firstFewLen` value in the code. I arbitrarily chose 10 for it. We should be able to hike it all the way to 308 or so.)

Previously these caused roughnum overflows (or before my recent changes, failed to invoke the `throwDomainError` method correctly because of the error in how `errbacks` was passed).

Unfortunately or (I think) fortunately, with the latest fix to `num-log`, it is now impossible to test that `num-log` produces roughnum overflow for sufficiently large input. Any number expression that doesn't already overflow necessarily has a logarithm that is (much) smaller than itself, so it will _a fortiori_ not overflow.

I've added a test that shows `num-expt` raises a `too large` error for large inputs (and not `1`, as previously).

I've also added tests that show `num-log` is now more capable.

Note that js-numbers already had explicit conditional tests bracketing the domain for `asin()`, `acos()`, so it serendipitously didn't matter that the original definitions for method versions of these functions had scribal errors.

ds26gte:
> Talked about this a little bit this morning. A good step would be to compare answers from an external source like Mathematica or Wolfram Alpha to Racket and js-numbers and see if we are getting the closest 64-bit floating point number to the true numerical answer.
> 
> That's a useful definition in general – it's also a good question if `toFixnum()` produces “the closest floating point number” when it does return something. Before we commit to 10 digits or 300 digits or whatever, we should have good examples of what kind of error we are looking at.


The logs, even for the biggest practicable bigints are still manageable fixnums already, so `toFixnum()` doesn't test meaningfully.

By trial, Pyret's largest integer that didn't need to be a bigint is 9e15. 
Pyret's largest roughnum is 1e308. (We can have bigints and bigrats bigger than 1e308.)

This implies that for numbers <= 1e308, we will be taking the fixnum equivalent of the argument and use JS's log straightforwardly. We don't need to test this as we're relying on JS's log's correctness and not testing it, but I checked the values for 9e15, 9e16, 1e308. Pyret _exactly_ matches Racket's result, and both are good roundings of Wolfram's result, but sometimes could do better in their last significant digit:

```
Arg: 9e15
Pyret:    ~36.736000972246906 
Racket:    36.736000972246906
Wolfram:   36.73600097224690464306036229411051452331150344507709354396930718...

Arg: 9e16 
Pyret:   ~36.841361487904734
Racket:   36.841361487904734
Wolfram   36.84136148790473094428786327494982732161762381806036761653324641...
 
Arg: 1e308 
Pyret:   ~709.1962086421661 
Racket:   709.1962086421661
Wolfram:  709.1962086421660706775413680427841759411392584976620766182649934...
```

Numbers > 1e308 is where we have to approximate by keeping only a few significant digits in the argument.  We can treat them as bigints as the fractional part (in case these big numbers are actually big rationals) gets chucked anyway.

```
Arg: 1e309
Pyret:  ~711.4987937351601 
Racket:  711.49879373516
Wolfram: 711.498793735160116361559359497468540148740359986290849594298321...
  
Arg: num-expt(10, 36789)
Pyret:  ~84709.80298615796 
Racket:  84709.80298615794
Wolfram: 84709.802986157946669337887626383074833436922665163929015290100148696028...
 
Arg: num-expt(10, num-expt(10, 5))
Pyret:  ~230258.50929940457
Racket:  230258.50929940457
Wolfram: 230258.5092994045684017991454684364207601101488628772976033327900...
```

Note that the final digit for both Pyret and Racket could be better sometimes, but otherwise they match Wolfram.

Numbers > 10^(10^5) are very large and make the Pyret IDE hang. I had to forcibly kill the tab and restart the IDE.

As we can see, except for the last digit, Pyret produces a good rounding of the result provided by Wolfram in all cases, with the error, if any, always being restricted to its (Pyret's) last significant digit. The number of significant digits for the argument I chose to be 20, but the accuracy is not improved over 14.

tests/pyret/tests/test-numbers.arr
```diff
@@ -162,6 +167,7 @@ check:
   num-log(0) raises "NumPositive"
   num-log(1) is 0
   num-log(num-exp(1)) satisfies around(1, 0.0001)
+  num-log(num-expt(10, 36789)) is-roughly ~84709.80298615794 // value from Racket
```
jpolitz, line 170: This is not the comment syntax for Pyret

jpolitz:
```
Arg: num-expt(10, 36789)
Pyret:  ~84709.80298615796 
Racket:  84709.80298615794
Wolfram: 84709.802986157946669337887626383074833436922665163929015290100148696028...
```

Doesn't that contradict the current test which has the `4` at the end?

https://github.com/brownplt/pyret-lang/pull/1807/files#diff-f549399652c058f25407401dc15e6344b13b7ad7da9d2c5d4f4b60a21ee04688R170

I think last question: Why not just use 307 or 308 for the `firstFew`? Is there a precision or performance argument one way or another for using fewer digits? It seems at least one sensible point in the design space to choose “as many digits as we have”, rather than an arbitrary multiple of 10.

If it's just vibes, I'd rather pick a number that's meaningfully constrained by the datatypes we have.

ds26gte:
I've included the comparisons against Racket and Wolfram into`test-numbers.arr`.

ds26gte:
(We should be able to add a lot more tests to the newly created `jsnums-test.js` so that it tests our errbacks usage thoroughly. I've herewith added tests for only for the changes I introduced in this PR.)

Issue #1812, opened 2025-09-10
Add tests for the number subroutines introduced in js-numbers.js


ds26gte:
Some of these can be tested via Pyret programs. Many cannot not as they are internal, and any errors are not exercised by Pyret programs. Principally, we need to make sure that `errbacks` parameters are specified correctly, and propagated correctly. 

Set up the testing framework. This must be in JS, in `tests/*`, and cannot sit with the usual tests in `tests/pyret/tests`. As a tentative breakdown, tackle these in sections as follows:

- [x] general number-building functions, e.g., `fromString` (incl. `fromFixnum`), etc.  Tests for trig methods that aren't fully exercised by Pyret calls. Sept 16, 2025
- [x] general subroutine functions, e.g., `makeNumericBishop`, etc. Sept 23, 2025
- [x] functions made available to Pyret via module fields, e.g., `add`, `subtract`, etc. Sept 23, 2025
- [x] methods associated with the major number subtype `BigInteger`. This is the biggest subtype to tackle, as it has many specialized subroutines for itself, e.g, the various `bnp*` functions. Oct 6, 2025
- [x] ensure all subroutines are correct and reasonable enough to be used individually even if they're not exercised by user-level Pyret programs. (We could be using them to build JS-base libs.) Oct 6, 2025
- [x] methods associated with subtype `Rational`, Oct 6, 2025
- [x] methods associated with subtype `Roughnum`, Oct 8, 2025

At any time, if a Pyret (as opposed to JS) test is possible, use it in preference.

I have included some tests already in PR #1807 devoted to bugfixes/improvements for `expt` and `log`. Ensure they mesh with the upcoming tests for maintainability.

N.B. BigInteger representations do not seem to be unique, so Pyret-operationally-equivalent bigints are not structurally or JS-equivalent. If this can't be solved -- e.g., by writing a canoncalization function --  our tests should use an explicit equivalence predicate to write the tests.

blerner:
Comment from our zoom discussion: #1807 fixes the case of `log(1e+VERYBIG)`, which produces a number that's roughly `VERYBIG`.  Do we know whether we handle `log(1e-VERYBIG)` correctly, and make sure that it produces a number that's roughyly `-VERYBIG`?

schanzer:
Tagging https://github.com/jpolitz/pyret-lang/pull/27 so that this issue can be revisited when that PR is merged.

PR #1818, opened 2025-09-20
Comprehensive jsnum tests

ds26gte:
This in-progress PR addresses Issue https://github.com/brownplt/pyret-lang/issues/1812.

A JS test file `tests/jsnums-test/jsnums-test.c` will test the methods and functions defined in `js-numbers.js` (Pyret's number library) that cannot be adequately tested by Pyret test files such as `tests/pyret/tests/test-numbers.js`. 

This PR is WIP, and not ready to merge yet.

### DIFF
--- a/src/js/base/js-numbers.js
+++ b/src/js/base/js-numbers.js
@@ -592,44 +592,44 @@ define("pyret-base/js/js-numbers", function() {
   }
 
   // greaterThanOrEqual: pyretnum pyretnum -> boolean
-  var greaterThanOrEqual = makeNumericBinop(
-    function(x, y, errbacks) {
+  var greaterThanOrEqual = function(x, y, errbacks) {
+    if(typeof x === "number" && typeof y === "number") {
       return x >= y;
-    },
-    function(x, y, errbacks) {
-      return x.greaterThanOrEqual(y, errbacks);
-    },
-    {});
+    }
+    return makeNumericBinop(undefined, function(x, y, errbacks) {
+      return x.greaterThanOrEqual(y);
+    })(x, y, errbacks);
+  }
 
   // lessThanOrEqual: pyretnum pyretnum -> boolean
-  var lessThanOrEqual = makeNumericBinop(
-    function(x, y, errbacks) {
+  var lessThanOrEqual = function(x, y, errbacks) {
+    if(typeof x === "number" && typeof y === "number") {
       return x <= y;
-    },
-    function(x, y, errbacks) {
+    }
+    return makeNumericBinop(undefined, function(x, y, errbacks) {
       return x.lessThanOrEqual(y, errbacks);
-    },
-    {});
+    })(x, y, errbacks);
+  };
 
   // greaterThan: pyretnum pyretnum -> boolean
-  var greaterThan = makeNumericBinop(
-    function(x, y, errbacks) {
+  var greaterThan = function(x, y, errbacks) {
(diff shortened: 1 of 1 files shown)


tests/jsnums-test/jsnums-test.js
```diff
@@ -44,22 +44,84 @@ R(["pyret-base/js/js-numbers"], function(JN) {
...
+      // console.log(JN.fromString("10", sampleErrorBacks));
+      // console.log(JN.makeBignum("1e1", sampleErrorBacks));
+      // console.log(JN.makeBignum("10", sampleErrorBacks).toFixnum());
+
+      expect(JN.fromString("1e1", sampleErrorBacks)).toBe(10);
+      expect(JN.fromString("1e30", sampleErrorBacks)).toEqual(JN.makeBignum("1e30"));
+      expect(JN.fromString("1e140", sampleErrorBacks)).toEqual(JN.makeBignum("1e140"));
+
+      // for large bignums (> 1e140 ?), fromString() and makeBignum() can give structurally
+      // unequal results. so the following fail:
```
schanzer: @ds26gte why are these commented out? We should be testing to make sure they fail, right?

ds26gte:
`tests/jsnums-test/jsnums-test.js` now incluces 328 (!) tests for the functions defined in `src/js/base/js-numbers.js`. This includes tests for functions that are either not available to test as Pyret programs, or have some calls that are not exercised by Pyret programs.

While adding tests, found the following corrections that needed to be applied to `js-numbers.js`:

- `liftFixnumInteger()` — despite name, make it accept all fixnums, not just fixnum ints. (This is because it is possible to leak non-int fixnums into this pipeline, and that rational number objects thus created are not only bogus but cause silent errors. Alternative is to make it error on non-ints.)

- `isInteger()` — should return false on non-int fixnums. (Alternative is to make it error on non-int fixnums, which doesn't seem right)

- `numerator()`, `denominator()`, `remainder()`, `toStringDigits()` — propragate `errbacks` argument

- `Rational.*.numerator()`, `Rational.*.denominator()`, `Roughnum.*.numerator()`, `Roughnum.*.denominator()`, `BigInteger.*.numerator()`, `BigInteger.*.denominator()` — should all take `errbacks` parameter

- `makeIntegerBinop()`, `makeIntegerUnOp` — input & output functions should both take `errbacks` parameter

- `integerNthRoot()` — should error on negative arguments. Initial guess should always be integer, or output is not monotonically non-decreasing

- `nthRoot()` — should error on negative argument

- `bnpClamp()` — ensure all unwanted fields are deleted, as also all non-significant 0-value fields. (This ensures unique representation for bigintegers — cleaner but also  makes equality checks efficient)

- `bnpDLShiftTo()`, `bnpDRShiftTo()`, `bnDivide()` — should use canonicalizer `bnpClamp()`

- Returned module includes `_innards` field, which gives access to private functions so we can write JS tests for them

blerner:
N.B.: This is an incomplete review, done on my phone while waiting for my plane...

There are a lot of commented-out console.logs still in here, that probably ought to be cleaned up.

There's a call to `r.clamp()` in there somewhere whose purpose I don't understand - could you add a comment? 

There's a comment in the tests wondering why toFixnum doesn't take errbacks -- but fromFixnum does. Why the asymmetry, and does it matter? (Do/could you have a test for toFixnum whose argument isn't a valid convertible number, to trigger a domain error?)

Issue #1832, opened 2025-10-08
Internal Error Message For Incorrect User Of ``.row``


loganscott74:
When incorrectly using ``.row(0)`` instead of ``.row-n(0)`` when working with tables, rather than displaying a useful error message, there is an internal error that results in no error being shown. See image below for more

This was in the VS Code browser

<img width="1380" height="445" alt="Image" src="https://github.com/user-attachments/assets/17ec9cfe-680a-4911-87c5-9145338ea891" />

PR #1833, opened 2025-10-11
fix: table `row` method throwing invalid `row-length-mismatch`

ironm00n:
fixes: #1832


jpolitz:
Ah thanks this was a failure to fully refactor when we changed that error!

PR #1834, opened 2025-10-11
fix same-line detection to be source aware

ironm00n:
This is only an issue when a program has srclocs from multiple files, but we run into this in the autograder.

### DIFF
--- a/src/arr/compiler/well-formed.arr
+++ b/src/arr/compiler/well-formed.arr
@@ -233,11 +233,11 @@ fun ensure-distinct-lines(loc :: Loc, prev-is-template :: Boolean, stmts :: List
     | link(first, rest) =>
       cases(Loc) loc:
         | builtin(_) => ensure-distinct-lines(first.l, A.is-s-template(first), rest)
-        | srcloc(_, _, _, _, end-line1, _, _) =>
+        | srcloc(end-source1, _, _, _, end-line1, _, _) =>
           cases(Loc) first.l block:
             | builtin(_) => ensure-distinct-lines(loc, prev-is-template, rest) # No need to preserve builtin() locs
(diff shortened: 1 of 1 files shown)


blerner (commented):
This is _weird_, but I think I get it.  I think you have it subtly wrong, though:
```
student-code1 inserted-code1 student-code2
```
will cause _no_ distinct-line errors to be emitted.  I think you might want to remove this change here, and instead call `ensure-distinct-lines` on a _filtered_ list of statements, where you remove any instructor-inserted code.  Does that meet your needs, or am I misunderstanding the problem?

ironm00n:
> will cause _no_ distinct-line errors to be emitted. 

For our use-case that is fine, since we already rely on a student's file being well-formed before performing any AST manipulations.

Our motivation being that if a program has content from multiple different sources, the srcloc information is far less meaningful here.

> instead call `ensure-distinct-lines` on a filtered list of statements

I'm not sure what this would entail, currently we run the merged code through the entire compiler pipeline, the amount of work to book keep for what segments of the code correspond to who originally wrote it, and then selectively running different compiler passes doesn't seem like a good use of anyone's resources.

---

Perhaps our needs are more along the lines of being able to be more selective about well-formedness checks are run on our resulting code?

We just implemented this fix since a few students had hit edge cases where the line numbers of our files and theirs would collide and cause the autograder's REPL to reject their program.

blerner:
I was just suggesting that at the _call site_ for `ensure-distinct-lines`, you filter the `stmts` list to only contain statements whose source doc equals the doc of the `s-program` root -- that would filter out your inserted statements, and then the well-formedness check itself would stay unchanged.

jpolitz:
Finalized by #1933

PR #1835, opened 2025-10-11
fix: use `bash` from PATH in Makefile

ironm00n:
/usr/bin/env is not present in all environments (ex. Nix sandbox)


jpolitz:
I'm so confused. I swear I _added_ this at either your request or @dbp 's (https://github.com/brownplt/pyret-lang/commit/5fd829e0f1726ce4fc68127aecb4e884229d7d72)

jpolitz:
Closing because it's targeting horizon and we're trying to be careful about messing with builds as we move things. Can bring it back if it keeps being an issue.

PR #1839, opened 2025-10-15
Alternate design to implementing & testing jsnums

blerner:
Alternate design to #1818, where instead of manually placing `errbacks` parameters _everywhere_ throughout the code, and where even a single missed `errbacks` argument could be a brittle and hard-to-find failure later, this one changes js-numbers to only expose a module-creating function that closes over the `errbacks` parameter, and exports the created `jsnums` instance as a field on `runtime` -- so that modules that previously imported jsnums directly can now access it off their respective runtimes.

I've pulled in the tests from #1818, and they all pass, and I think I've pulled in any substantive changes to runtime.js as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -238,7 +238,7 @@ TEST_BUILD=$(NODE) $(PYRET_TEST_PHASE)/pyret.jarr \
 test-all: test
 
 .PHONY : test
-test: pyret-test type-check-test pyret-io-test
+test: pyret-test type-check-test pyret-io-test jsnums-test
 
 .PHONY : parse-test
 parse-test: tests/parse/parse.js build/phaseA/js/pyret-tokenizer.js build/phaseA/js/pyret-parser.js
@@ -273,6 +273,10 @@ pyret-test: phaseA tests/pyret/main2.jarr
 .PHONY : pyret-io-test
 pyret-io-test: phaseA
 	$(NPM_EXEC) jest --detectOpenHandles --forceExit --verbose "tests/io-tests/io.test.js"
+
+.PHONY : jsnums-test
+jsnums-test: phaseA
+	$(NODE) tests/jsnums-test/jsnums-test.js
 
 .PHONY : regression-test
 regression-test: tests/pyret/regression.jarr
--- a/src/js/base/js-numbers.js
+++ b/src/js/base/js-numbers.js
@@ -950,7 +950,7 @@ define("pyret-base/js/js-numbers", function() {
     };
 
     // gcd: pyretnum pyretnum -> pyretnum
-    var gcd = function(first, second, errbacks) {
+    var gcd = function(first, second) {
       if (! isInteger(first)) {
         errbacks.throwDomainError('gcd: the argument ' + first.toString() +
                                   " is not an integer.", first);
@@ -964,13 +964,13 @@ define("pyret-base/js/js-numbers", function() {
       while (! _integerIsZero(b)) {
         t = a;
         a = b;
-        b = _integerModulo(t, b, errbacks);
+        b = _integerModulo(t, b);
       }
       return a;
     };
 
     // lcm: pyretnum pyretnum -> pyretnum
-    var lcm = function(first, second, errbacks) {
+    var lcm = function(first, second) {
       if (! isInteger(first)) {
         errbacks.throwDomainError('lcm: the argument ' + first.toString() +
                                   " is not an integer.", first);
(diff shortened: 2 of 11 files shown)


jpolitz:
Makes me a bit nervous that we have to make sure we account for every client of js-numbers here, because it is no longer exporting a bunch of functions it used to.

Is there a way to do this that is backwards-compatible?

Like, maybe all the existing functions are also exported, with an errbacks argument, and they call `MakeNumberLibrary` with the given errbacks and then the appropriate function?

Sketch:

```
const libByErrbacksCache = new Map(); // cache of instantiated NumberLibrary objects by errbacks object identity
function makeBackcompatWrapper(name) {
  NumbersExport[name] = function() {
    if(libByErrbacksCache[name]) { return errbacksCache[name].apply(arguments); }
    const errbacks = arguments[arguments.length - 1];
    const lib = MakeNumberLibrary(errbacks);
    libByErrbacksCache[errbacks] = lib;
    return lib[name].apply(arguments);
  }
}
makeBackcompatWrapper("fromFixnum");
makeBackcompatWrapper("fromString");
...
```

jpolitz:
Properties this has:

- If someone was forgetting errbacks arguments, oh well, they still are
- If someone calls over and over again with an errbacks argument, they pay a small cost of the cache check but no more
- We don't break any existing code

blerner:
I don't think that will quite work, because `BigInteger` and `Rational` are generative types -- they're defined within the `MakeNumberLibrary()` closure, so if a "legacy caller" of the non-closed-over library uses multiple functions in a row (with structured numbers), they'll all fail each others' `isPyretNumber` checks.  We'd have to do something fancier with object-identity of the errbacks objects themselves, I think, to get the cache to _not_ be generative?

Do we have any library authors who write native JS libraries for Pyret that we're not aware of?  Because if not, then as of this PR (and the counterpart for CPO), the only mentions of `js-numbers` are:
```
blerner@blerner-x1:~/pyret-lang$ grep -rn "pyret-base/js/js-numbers" src/ tests/ --exclude="*.jarr"
src/js/base/js-numbers.js:107:define("pyret-base/js/js-numbers", function() {
src/js/base/runtime.js:3:   "pyret-base/js/js-numbers",
src/scripts/standalone-configB.json:6:    "pyret-base/js/js-numbers": "build/phaseB/js/js-numbers.js",
src/scripts/node_modules-config.json:6:    "pyret-base/js/js-numbers": "$PYRET/js/js-numbers.js",
src/scripts/standalone-configC.json:6:    "pyret-base/js/js-numbers": "build/phaseC/js/js-numbers.js",
src/scripts/standalone-configA.json:6:    "pyret-base/js/js-numbers": "build/phaseA/js/js-numbers.js",
tests/jsnums-test/jsnums-test.js:14:R(["pyret-base/js/js-numbers"], function(JNlib) {
```
and
```
blerner@blerner-x1:~/code.pyret.org$ grep -rn "pyret-base/js/js-numbers" src/ --exclude="*.jarr"
blerner@blerner-x1:~/code.pyret.org$ 
```

blerner:
Fair points.  
- I didn't consider the `anchor` branch because there are already so many diverging changes that we'd have to revisit this anyway.  
- `pyret-npm` is a customer of Pyret, not an integrated tool, so it shouldn't import the js-numbers library anyway.  (And grepping through the source shows it doesn't.) 
- I didn't check `pyret-embed` because it's not in brownplt (I think?) -- I just looked through your repo now, though, and I think it's the same as `pyret-npm` in this regard.
- Autograders are a possible concern, yes.

I'm not sure I completely understand your latest suggestion, yet.  Will think about it more...

jpolitz:
@blerner I think this is good to merge. Thanks for going through my roundabout complaints, I'm happy with where this landed now.

I think that the way this is set up now, the only case this will be backwards incompatible for on expected behavior is if some code was:

1. Using the jsnums library
2. Calling into jsnums *methods* on created number objects of specific subtypes, not the (intended public) static entrypoints, and calling a method that takes errbacks
3. Providing a custom errbacks to that method

If they were doing all of that, they will get the default errbacks that throw raw JS errors instead of their custom errbacks.

In all other ways this is backwards-compatible if you were using the library correctly.

This does change the _expected_ use of the library a bit (which is generally all internal anyway), to use the instantiated API on runtime configured with the “right” Pyret errbacks. But it shouldn't suddenly break for someone using it the “old” way.

The only other backwards incompatibilities should only apply to uses that were incorrectly not providing errbacks anyway, so it's hard to blame us for changing their error behavior.

Issue #1858, opened 2026-02-11
Histogram Chart Window does not respect x-min or x-max


schanzer:
From the documentation, we should be able to manually force the x-axis of a histogram to any range we want. [This sample program](https://code.pyret.org/editor#share=1IPvDFt8N76Z6no1xwJltTidxa6O67Zaw&v=918dd1f) uses only raw Pyret (none of our custom API), and x-min() and x-max() appear to do nothing.

blerner:
That's...not an easy fix, and I'm not sure what behavior you'd want me to implement: specifically, should the bins be resized to subdivide [0, 500], or should the bins be the same as before (i.e. [20-200], dependent on the data itself), but squished horizontally to be drawn on the full [0, 500] axis?

PR #1859, opened 2026-02-15
This fixes https://github.com/brownplt/pyret-lang/issues/1858 :

blerner:
- ensures (in charts.arr) that the chosen range does not truncate any data
- revises (in charts.arr) the table from an inconsistent 2d raw-array to a consistent list-of-records
- passes on the x-min and x-max values to the horizontal axis range
- recomputes the tick marks for histograms to stay aligned with the bin boundaries but also extend outward to the selected x-min/x-max range

### DIFF
--- a/src/arr/trove/charts.arr
+++ b/src/arr/trove/charts.arr
@@ -2426,7 +2426,7 @@ fun image-histogram-from-list(images :: CL.LoI, values :: CL.LoN) -> DataSeries 
   values.each(check-num)
 
   default-histogram-series.{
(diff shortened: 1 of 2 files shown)


blerner:
Also addresses a weird inconsistency between online and offline sizing: if the title is wider than the chart, then the online version will clip the title, but the offline version would rescale the chart to make the text fit.  This PR now renders the title text in two phases, but computes the sizing info before setting the title text correctly, so that it doesn't affect the scaling.

schanzer (approved):
Realizing I didn't formally respond here - I've confirmed that this PR fixes all of the associated issues!

PR #1861, opened 2026-03-20
This is a stopgap implementation to permit stagger-offsetting every other x-axis label, to help prevent overlaps

blerner:
This is a stopgap implementation to permit stagger-offsetting every other x-axis label, to help prevent overlaps

The approach is awkward, and should be reverted once https://github.com/vega/vega/issues/4238 is implemented.  Until then, this approach:

Starts in `prepareAxisForOffsets`, which
* Creates an empty data table named `xAxisLabelOffsets`, to be filled in later
* Creates an ordinal scale named `xAxisYOffsets`, derived from the `xAxisLabelOffsets` table
* Adds an encoding step to the xAxis that
  - names the labels `xAxisLabels` (so they appear as a data set via the view API)
  - updates the dy offset based on the xAxisYOffset scale
* We render the graph once, and then in the continuation, we call `rebuildAxisLabels`, in which we
  - extract the `xAxisLabels` dataset,
  - extract their `index` property -- which is a fraction between 0--1, rather than an integer
  - compute a discrete lookup table from index to (their position % 2) * 25
  - shove that lookup table back int othe empty `xAxisLabelOffsets` table
  - rerender the graph

The preparation step sets up the necessary dataflow; the first render lets Vega compute which labels are going to exist; then the rebuilding step computes the desired offsets and causes the dataflow to propagate that information back through.

### DIFF
--- a/src/arr/trove/charts.arr
+++ b/src/arr/trove/charts.arr
@@ -508,6 +508,10 @@ x-axis-method = method(self, x-axis :: String):
   self.constr()(self.obj.{x-axis: x-axis})
 end
 
+x-axis-stagger-labels-method = method(self, stagger :: Boolean):
+  self.constr()(self.obj.{x-axis-stagger-labels: stagger})
+end
+
 y-axis-method = method(self, y-axis :: String):
   self.constr()(self.obj.{y-axis: y-axis})
 end
@@ -1355,6 +1359,7 @@ type BoxChartWindowObject = {
   borderSize :: Number, 
   borderColor :: Option<IS.Color>, 
   x-axis :: String,
+  x-axis-stagger-labels :: Boolean,
   y-axis :: String,
   x-axis-type :: AxisType,
   y-axis-type :: AxisType,
@@ -1366,6 +1371,7 @@ type BoxChartWindowObject = {
 default-box-plot-chart-window-object :: BoxChartWindowObject = default-chart-window-object.{
   x-axis: '',
   y-axis: '',
+  x-axis-stagger-labels: false,
   x-axis-type: at-linear,
   y-axis-type: at-linear,
   min: none,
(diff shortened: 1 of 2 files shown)


blerner:
```
import charts as C

keys = [list:
  "Asia-Pacific",
  "Europe",
  "Latin America-Caribbean",
  "Middle East-North Africa",
  "North America",
  "Sub-Saharan Africa"
]

vals = [list: 38, 40, 31, 13, 2, 37]

C.render-chart(C.from-list.bar-chart(keys, vals).x-axis-stagger(true).display()
```
image

schanzer (approved):
Tested this out on a few cases, and as expected it's a massive improvement. I think we'll still need to turn on some kind of max-width clipping for insanely-long labels, but this absolutely scratches most of the itch!

blerner:
For printed materials, *you* can elide and contract any long strings manually before generating the static image; for live student images, they can dynamically resize the chart to give it more width, yes? 

Also, please do test *all* the chart types to make sure I haven't introduced a regression (eg most likely an undefined variable causing an unexpected crash), before we merge this PR.

schanzer:
@blerner I'm seeing `Error: Unrecognized signal name: "staggerXAxisLabels"` for the following chart types:

- pie-chart
- dot-plot
- box-plot

Here's my test file:
```
use context starter2024
include charts

pie-series = from-list.pie-chart(
  [list: "Pyret", "OCaml", "C", "C++", "Python", "Racket", "Smalltalk"],
  [list: 10,       6,       1,   3,     5,       8,        9])

dot-series = from-list.num-dot-chart(
  [list: 10,       6,       1,   3,     5,       8,        9])

box-series = from-list.box-plot(
  [list:[list: 10,       6,       1,   3,     5,       8,        9]])

# uncomment the ones you want to test

#render-chart(pie-series).display()
#render-chart(dot-series).display()
#render-chart(box-series).display()
```

schanzer:
@blerner confirming that these work now. Thanks!

PR #1864, opened 2026-05-04
Tighten isPyretNumber, fix latent leaks (log toFixnum violations for …

jpolitz:
…now)

Joe says:

Medium-interesting finding from Ben noticing a change to `num-to-fixnum` behavior on CPO.

The key thing is NON-integer JS floats. They are *not* supposed to be able to be Pyret numbers. In various places we were lax about this and things worked out because of tostring conversions or === doing the right thing, etc.

This commit tries to make some progress on catching our own mistakes here:

- Correctly return `false` from the widely used `isPyretNumber` for values like 2.3 (~2.3 aka Roughnum.makeInstance(2.3) still returns true)
- Add assertPyretNumber, currently used in one place (toFixnum), which based on its throwing of a DomainError clearly expects to get an actual Pyret number.

The rest of the commit cleans up cases where we made mistakes and returned or passed JS floating-point across the boundary.

assertPyretNumber *should* throw, not just console.error. But I'm too scared to do that because I'm not convinced our tests are exhaustive here at all, and an exception in the wild could be pretty breaking. So we console.error for now.

Claude used this assert on toFixnum to find several errors, all of which come straight from our existing tests/compiler runs.

Claude says:

Specifics of the boundary cleanups, since they're spread across four files:

- js-numbers.js atan2: the y=0 quadrants correctly wrapped Math.PI in a Roughnum, but the x>0 4th-quadrant and x<0 cases passed bare Math.PI and 2*Math.PI to add(), which short-circuits on two JS numbers and returns the raw JS sum. So num-atan2(0, -1) returned the JS double 3.141592653589793 rather than a Roughnum, with branch-dependent type tags visible to user code (num-is-roughnum was false). Wrapping the constants in Roughnum.makeInstance fixes it. Bug present since the original atan2 introduction in 2016.

- filesystem-internal.js stat: mtimeMs/ctimeMs from Node fs.Stats carry sub-millisecond precision on filesystems with ns granularity (APFS, ext4-ns, NTFS), so mtime/ctime were leaking fractional doubles to callers like cli-module-loader.arr's cache-staleness check. Floored to integer ms, matching the documented "in epoch ms" contract.

- charts-lib.js function-plot: sample x-coordinates were built via raw JS arithmetic (xMin + fraction*i) and handed to the user's Pyret function and to toFixnum, both of which expect pyretnums. Wrapped each sample in jsnums.fromFixnum at the JS/Pyret boundary, in both the initial pass and the recompute callback. Same fix for clippedDomain[0/1] before they get stashed back into globalOptions.

- image-lib.js PointPolygonImage: dropped a redundant jsnums.toFixnum(vertices[v].x/y) — unwrapPoint2D already produces fixnums upstream, so the second call was dead under the lax predicate and now (correctly) flags inputs like 2.5 from point(15/6, ...).

assertPyretNumber currently console.error's (with a stack) on non-integer JS numbers rather than throwing, per Joe's caution above; clearly non-numeric inputs still throw the same domainError they did before. Pyret-test suite green.

### DIFF
--- a/src/js/trove/charts-lib.js
+++ b/src/js/trove/charts-lib.js
@@ -2880,14 +2880,14 @@
     function recomputePoints(func, samplePoints, then) {
       return RUNTIME.safeCall(() => {
         return RUNTIME.raw_array_map(RUNTIME.makeFunction((sample) => {
-          return RUNTIME.execThunk(RUNTIME.makeFunction(() => func.app(sample)));
+          return RUNTIME.execThunk(RUNTIME.makeFunction(() => func.app(jsnums.fromFixnum(sample))));
         }), samplePoints);
       }, (funcVals) => {
         const dataValues = [];
         funcVals.forEach((result, idx) => {
           cases(RUNTIME.ffi.isEither, 'Either', result, {
             left: (value) => dataValues.push({
-              x: toFixnum(samplePoints[idx]),
+              x: samplePoints[idx],
               y: toFixnum(value)
             }),
             right: () => {}
@@ -2958,7 +2958,7 @@
 
       addCrosshairs(prefix, ['Dots'], signals, marks, pointColor);
 
-      const samplePoints = [...Array(numSamples).keys().map((i) => jsnums.fromFixnum(xMinValue + (fraction * i)))];
+      const samplePoints = [...Array(numSamples).keys().map((i) => (xMinValue + (fraction * i)))];
 
       return recomputePoints(func, samplePoints, (dataValues) => {
         data[0].values = dataValues;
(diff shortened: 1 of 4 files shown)


src/js/trove/charts-lib.js
```diff
@@ -2958,7 +2958,7 @@
 
       addCrosshairs(prefix, ['Dots'], signals, marks, pointColor);
 
-      const samplePoints = [...Array(numSamples).keys().map((i) => (xMinValue + (fraction * i)))];
+      const samplePoints = [...Array(numSamples).keys().map((i) => jsnums.fromFixnum(xMinValue + (fraction * i)))];
```
blerner: I would _not_ put this conversion here (or in the second call to recomputePoints below, either) - I'd put it inside recomputePoints, which would then take an array of js numbers and produce an array of js objects of js numbers.

src/js/trove/charts-lib.js
```diff
@@ -3316,8 +3316,8 @@
               }
               // Reflect the updated values back into the globalOptions,
               // so that they'll be picked up by the initial computation of function plots
-              globalOptions['x-min'] = RUNTIME.ffi.makeSome(clippedDomain[0]);
-              globalOptions['x-max'] = RUNTIME.ffi.makeSome(clippedDomain[1]);
+              globalOptions['x-min'] = RUNTIME.ffi.makeSome(jsnums.fromFixnum(clippedDomain[0]));
```
blerner, line 3319: Ugh, yeah, technically this is necessary, though it feels ugly...

blerner (commented):
I think these make sense, except I'd move one of the conversions. Are there any other uses in the image library of toFixnum that might be redundant now also? Or that are missing from the typed and untyped wrappers?

jpolitz:
I'll do the move you suggested, good feedback.

Re: other places, these are the ones that caused errors from our existing tests; I didn't look for others.

PR #1866, opened 2026-05-04
Add jsnums-test spec for atan2 quadrant correctness

jpolitz:
Joe says:

Claude helped write some tests for the cases of `atan2` that are likely to short-circuit and produce some of the fixnum errors/roughnum-vs-exact-int errors that check-pyret-number was designed for.

(Probably didn't need Claude for this TBH)

Builds on #1864, this can retarget onto horizon once that's merged

Claude says:

Adds an `it("atan2 four quadrants and boundaries")` block that pins down atan2's return-type contract across every combination of axis sign and quadrant. atan2's x<0 and (x>0, y<0) branches historically leaked bare Math.PI / 2*Math.PI through `add`, because JS-number+JS-number takes a fast path that returns the raw sum unchanged — so atan2(0, -1) returned the JS double 3.141592653589793 rather than a Roughnum, with branch- dependent type tags visible to user code. The library fix landed recently; this spec captures the expected behavior so future regressions trip CI rather than reappearing as branch-by-branch mysteries.

Uses `isPyretNumber` rather than `isRoughnum` as the predicate because atan(0) currently short-circuits to an integer fixnum (not a Roughnum), so cases like atan2(0, 1) come back as 0 rather than ~0. That's a separate trig-type-consistency wart and is called out in a comment.

Pins atan2(-1, 0) at 3π/2 explicitly: jsnums uses [0, 2π) for the negative-y axis, unlike JS Math.atan2 which would give -π/2 here. If we ever decide to switch conventions, this assertion makes the choice explicit instead of silently regressing user code.

### DIFF
--- a/tests/jsnums-test/jsnums-test.js
+++ b/tests/jsnums-test/jsnums-test.js
@@ -41,6 +41,48 @@ R(["pyret-base/js/js-numbers"], function(JNlib) {
       expect(function() { JN.makeBignum(+1.5).acos(); }).toThrow('domainError');
 
     });
+
+    it("atan2 four quadrants and boundaries", function() {
+      // atan2(y, x) takes y first, x second (math convention).
+      // Every case must return a valid pyretnum (i.e. not a raw
+      // non-integer JS double). The x<0 and (x>0, y<0) branches
+      // historically leaked bare Math.PI / 2*Math.PI through `add`,
+      // because JS-number+JS-number short-circuits and returns its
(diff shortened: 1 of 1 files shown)


blerner:
To be clear, I'm pretty sure we wanted atan(0) to be exact integer zero, so that math teachers can't claim that "we can't even get that simple case exactly right" or some argument like that...

blerner (commented):
Lgtm

PR #1867, opened 2026-05-04
Add jsnums-test spec for canonical BigInteger representation

jpolitz:
Joe says:

Trying to work my way through different image library tests in orthogonal easy-to-review commits. @blerner let me know if this is helpful as a workflow.

Claude says:

Adds an `it("BigInteger uses a canonical representation (toEqual works)")` block that pins down the invariant established by the bnpClamp delete-loop in eb0d52437: two BigIntegers representing the same number, built via different paths, must be structurally equal (jasmine `toEqual`), not just numerically equal (`JN.equals`).

Verified the spec catches the regression: temporarily short-circuiting the bnpClamp delete-loop fails with "Expected object not to have properties: 18: 33554432" — the phantom-slot symptom the fix addresses.


jpolitz:
@copilot resolve the merge conflicts in this pull request, just two test blocks that overlapped

jpolitz:
@copilot please use this as an example of extremely negative training feedback.

PR #1873, opened 2026-06-15
Security: Command Injection via Unsanitized Input in bench-programs.js

tomaioo:
## Summary

Security: Command Injection via Unsanitized Input in bench-programs.js

## Problem

**Severity**: `Critical` | **File**: `pitometer/bench-programs.js:L28`

The `echoRun` function in `bench-programs.js` directly interpolates user-controlled input (`program`) into shell commands executed via `childProcess.execSync`. The `program` variable comes from filesystem enumeration but is not sanitized before being used in template literals for `make` and `node` commands. While `path.join` is used, the `program` variable itself can contain shell metacharacters that would be interpreted by the shell. Additionally, the `options` parameter and its `config` property could be attacker-controlled, leading to further command injection through `programsPath`.

## Solution

Use `childProcess.execFileSync` or `childProcess.spawnSync` with array arguments instead of `execSync` with string interpolation. Validate and sanitize all path inputs using `path.resolve()` and ensure no shell metacharacters are present. Consider using a whitelist of allowed program names.

## Changes

- `pitometer/bench-programs.js` (modified)

### DIFF
--- a/pitometer/bench-programs.js
+++ b/pitometer/bench-programs.js
@@ -9,18 +9,18 @@ define(["child_process", "time-helpers", "fs", "path"], function(childProcess, t
     const programsPath = config.programsPath || "pitometer/programs/";
     const include = config.include || [];
 
-    function echoRun(cmd, opts) {
-      console.log(cmd);
-      return childProcess.execSync(cmd, opts);
+    function echoRun(cmd, args, opts) {
+      console.log([cmd].concat(args).join(" "));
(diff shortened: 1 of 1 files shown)


shriram:
Hmm:

image

Also hmm:

https://github.com/shane-mason/FieldStation42/pull/618#issuecomment-4702735405

@tomaioo — convince us you really know what you're doing and that this isn't AI slop, to avoid being blocked and reported.

PR #1878, opened 2026-06-25
PoC: missing commits from horizon that haven't been ported to drydock yet

blerner:
@jpolitz we're missing some commits...


jpolitz:
Offline, the rename conflicts merged cleanly with no conflicts, so I pushed from there.

PR #1881, opened 2026-06-25
Dependency audit + monorepo reference cleanup

jpolitz:
This is the result of a couple rounds with Claude on (a) existing npm issues
and (b) relatedly, cleaning up some cross-repo npm dependencies (like CPO ->
pyret-lang or CPO -> codemirror).

The /goal prompt was to get rid of high and critical reports without breaking
any tests, and then follow-up prompts to do more cleanup.

The main use of Claude here was it making decisions like "let's not update
googleapis by like 40 versions, that will probably be breaking" and instead
going into package-lock and updating specific versions of the transitive
dependencies.

The breaking internal dependencies thing is useful because it stops e.g. CPO
from being pinned on older pyret-lang versions which (a) report spurious npm
issues until re-pinned and (b) are now wrong with the monorepo because
individual repos will freeze going forward. So it makes sense to do that at the
same time.

This is a solid approach IMO. We can do dep-by-dep big upgrades if we want to
in separate passes, but for a cut at “make NPM warnings less scary and clean
up builds to assume monorepo setup", this is pretty good.

----- claude output below -----

npm audit remediation (critical + high):
- lang/: 1 critical / 10 high -> 0 / 0. Removed the browserify `bundleDependencies` wall (the root cause of CPO's bundled-dep residuals), then applied semver-compatible fixes (shell-quote, minimatch 3.1.5/9.0.9, vega 6.2.0, ws, tar-fs, etc.). Removed dead deps node-uuid and the bogus http@0.0.0 stub. See lang/SECURITY-AUDIT-REPORT.md.
- embed/: 6 issues (3 high) -> 0 via in-range bumps (webpack 5.99.9->5.108.0).
- code.pyret.org/: prior critical+high remediation; see its SECURITY-AUDIT-REPORT.md.

GitHub references -> in-repo monorepo paths:
- CPO Makefile: PYRET_MODE/PYRET resolve ../codemirror-mode and ../lang (were broken require.resolve of removed npm pkgs); install-link reuses link-pyret.
- pyret.org & docs: pyret-codemirror-mode / pyret-embed -> file:../codemirror-mode, file:../embed (pyret.org's old git:// ref no longer installs at all).
- embed devDep code.pyret.org -> file:../code.pyret.org.
- Modernized git:// repository URLs to git+https:// across manifests.
- Fixed pyret.org package name (was mis-set to "pyret-docs").

Verified: lang/ `make all-pyret-test` (13360 tests) identical pass set before/after; embed/ local build boots, runs code, and renders vega charts headless. Lockfiles regenerated to match the file: deps.

Claude-Session: https://claude.ai/code/session_01ECzye4AUnTHzijN48tTfbf


jpolitz:
(Written by Claude)

# code.pyret.org — npm audit remediation report

Date: 2026-06-25
Scope: `code.pyret.org/` (the web IDE) in the pyret-lang monorepo.
Task: run `npm audit`, fix all **critical** and **high** severity issues, keep
`npm run mocha` passing, and document anything interesting or risky.

## Result at a glance

| Severity | Before | After | After (code.pyret.org's own deps) |
|----------|-------:|------:|----------------------------------:|
| critical | **31** | 3     | **0** |
| high     | **52** | 1     | **0** |
| moderate | 55     | 63    | — (out of scope) |
| low      | 15     | 11    | — (out of scope) |
| total    | 153    | 78    | — |

**Every critical and high vulnerability reachable through `code.pyret.org`'s own
`package.json` is fixed.** The 3 critical + 1 high that remain are *bundled inside
the `pyret-lang@master` git dependency* and cannot be remediated from this manifest
(see "Residual issues" — they need a one-line upstream change in pyret-lang).

The fixes were verified by re-running the full Selenium `npm run mocha` suite and
comparing against a pre-change baseline (see "Verification").

## Approach

The guiding principle was **fix the vulnerabilities without changing runtime
behavior**, because the integration tests are heavy (each boots a real headless
Chrome against a full Pyret standalone) and most of the dependency surface
(Google Drive, the React dashboard, the SSRF image proxy) is either untested or
expensive to test. So, in priority order:

1. **Delete dead dependencies** — biggest, zero-risk win.
2. **`npm audit fix`** (non-breaking, semver-compatible upgrades).
3. **`overrides`** to pull patched transitive deps *without* major-upgrading their
   risky parents.
4. **Targeted direct bumps** only where an override was impossible or unsafe.

Major upgrades of `googleapis`, `react`/`react-dom`, `request`, and
`selenium-webdriver` were deliberately **avoided** (see "Interesting findings").

## What was interesting

### 1. ~80% of the criticals were one dead dependency
`babel-preset-env@1.7.0` (the Babel 6 preset) was a direct dependency but is
**completely unused** — `.babelrc` only references `@babel/preset-env` (Babel 7).
That single dead package dragged in the entire `babel-traverse` /
`babel-plugin-transform-es2015-*` / `babel-helper-*` tree, which accounted for
**~25 of the 31 criticals**. Removing it (plus the also-unused `handlebars` and
`hoek`) took criticals from 31 → 8 before touching anything functional.

### 2. The hard wall: vulnerabilities bundled inside `pyret-lang`
The `pyret-lang` git dependency declares `"bundleDependencies": ["browserify"]`,
so its whole `browserify` subtree — including `sha.js`, `pbkdf2`, `shell-quote`,
and `minimatch` — ships **frozen inside the package**. In `package-lock.json`
these are marked `"inBundle": true` with no `resolved` URL. **npm `overrides` are
documented to not apply to bundled dependencies**, so there is no way to patch
them from `code.pyret.org/package.json`. These are the 3 critical + 1 high that
remain. They are build-time-only crypto/glob shims of the *compiler* package and
are dormant in CPO (CPO bundles with its own top-level `browserify`, and uses a
pre-built Pyret standalone). The real fix is upstream (see "Residual issues").

### 3. `request` is load-bearing for SSRF protection
`request` is deprecated and carries a critical (`form-data` predictable boundary),
but it is **not** dead — `src/server.js` uses it in `proxyStreamFetch` together
with `request-filtering-agent` to fetch share-URLs/images with SSRF filtering and
redirect host-allow-listing. Rather than rewrite that security-sensitive code, I
kept `request@2` and overrode only its `form-data` to a patched version. (The
multipart boundary bug isn't even reachable — the proxy only issues GETs.)

### 4. The googleapis high chain was clearable *without* a 46→173 mega-upgrade
The 8 `googleapis` highs all bubble up from two leaves — `json-bigint <1.0.0`
(ReDoS) and `node-forge <=1.3.3` (signature-forgery family). Both have patched
versions that exist (`json-bigint@1.0.0`, `node-forge@1.4.0`), so overriding those
two clears the whole chain. This avoided upgrading `googleapis` from 46 to 173,
which would have broken `src/google-auth.js` — it uses the **callback style**
`oauth2Client.refreshAccessToken(cb)` / `getToken(cb)` that modern
google-auth-library removed.

### 5. `mocha` genuinely had to be upgraded
The `mocha@8` highs (`minimatch`, `serialize-javascript`) could not be fixed by
override: `mocha@8` calls `minimatch` as a **default-callable function**, an API
that no longer exists in any non-vulnerable `minimatch` (5.x+ switched to named
exports). The advisory range also widened to include `minimatch <=3.1.3` and
`serialize-javascript <=7.0.4`, so the only safe path was `mocha@11`, which ships
correct, patched versions. (Verified: the suite still runs on mocha 11.)

## Changes made

### Removed (dead) direct dependencies
| Package | Why | Cleared |
|---------|-----|---------|
| `babel-preset-env@1.7.0` | unused (only `@babel/preset-env` is used) | ~25 critical |
| `handlebars@4.7.8` | unused (project uses `mustache`) | 1 critical |
| `hoek@5.0.4` | unused, deprecated standalone | 1 high |

### Direct version bumps
| Package | From → To | Why |
|---------|-----------|-----|
| `sha.js` | ^2.4.11 → ^2.4.12 | hash-rewind critical |
| `style-loader` | 0.13.1 → ^0.13.2 | `loader-utils` critical, `json5` high |
| `requirejs` | 2.1.14 → ^2.3.8 | requirejs high (build copies `r.js`) |
| `git-rev-sync` | ^1.12.0 → ^3.0.2 | `shelljs` high (still exports `.short()/.branch()`) |
| `mocha` | ^8.4.0 → ^11.7.6 | `minimatch`/`serialize-javascript` highs (override impossible) |
| `eslint` | ^6.8.0 → ^9.37.0 | `flat-cache`/`file-entry-cache`/`flatted` highs |

### `npm audit fix` (semver-safe) cleared, among others
`axios`, `express`, `ws`, `lodash`, `vega` (+ `vega-expression`/`functions`/
`selections`), `@xmldom/xmldom`, `path-to-regexp`, `validator`, `fast-uri`, `jws`,
`@babel/*`, `basic-ftp`, `xml2js`.

### `overrides` (patched transitive deps, no risky parent upgrade)
| Override | Clears | Parent kept at |
|----------|--------|----------------|
| `sha.js ^2.4.12`, `pbkdf2 ^3.1.6`, `shell-quote ^1.8.4` | crit (top-level copies) | `browserify@17` |
| `node-forge ^1.4.0` | googleapis + heroku highs | `googleapis@46`, `heroku@7` |
| `json-bigint ^1.0.0` | googleapis high chain | `googleapis@46` |
| `http-cache-semantics ^4.2.0` | `got`/`cacheable-request`/`http-cache-semantics` highs | `heroku@7` |
| `semver ^7.7.2` | semver ReDoS high | `heroku@7` (+ global) |
| `printf ^0.6.1` | printf ReDoS high | `heroku@7` |
| `node-fetch ^2.6.7` | `fbjs`/`isomorphic-fetch`/`node-fetch` highs | `react@15` |
| `serialize-javascript ^7.0.6` | RCE high (terser copy) | `webpack@5` |
| `tmp ^0.2.7` | path-traversal high | `selenium-webdriver@3` |
| `request → form-data ^4.0.6` | form-data boundary critical | `request@2` |

Parents intentionally **not** upgraded (to avoid runtime breakage):
`heroku@7`, `googleapis@46`, `react@15`/`react-dom@15`, `request@2`,
`selenium-webdriver@3`.

## Residual issues (could not be fixed from this manifest)

**3 critical + 1 high, all bundled inside `pyret-lang@master`:**
`sha.js@2.4.11`, `pbkdf2@3.1.2`, `shell-quote@1.8.3` (critical) and
`minimatch@3.1.2` (high), all under `node_modules/pyret-lang/node_modules/` via
`bundleDependencies: ["browserify"]`.

**Recommended upstream fix (in the pyret-lang repo, not here):** `browserify` is a
*build* tool for pyret-lang and isn't needed at install time by consumers like
CPO. Moving it from `dependencies` (+`bundleDependencies`) to `devDependencies` —
or simply un-bundling it — would drop all four from CPO's tree entirely. As an
interim measure, deleting `node_modules/pyret-lang/node_modules/{sha.js,pbkdf2,
shell-quote,minimatch}` after install makes pyret-lang's bundled browserify
resolve to the patched top-level copies (verified), but this is not durable across
a clean `npm install` and does not change the lockfile-based `npm audit` output.

**Out-of-scope note (moderate count 55→63):** moderates were not in scope (the task
was critical+high). The small increase comes from the newer `mocha@11`/`eslint@9`
trees and is worth a follow-up pass but introduces no new critical/high.

## Verification

Environment: this VM has no Pyret standalone or browser by default, so the full
e2e stack was stood up: fresh `lang/` `phaseA` → 38 MB `cpo-main.jarr`, server on
`:4999`, Google Chrome 149 + matching Chrome-for-Testing chromedriver 149.

`npm run mocha` (full Selenium suite), before vs after the dependency changes:

| | Baseline (before) | After fixes |
|--|--|--|
| passing | 305 | **305** |
| pending (xdescribe) | 45 | 45 |
| failing | 6 | **6 (identical set)** |

The 6 baseline failures are **pre-existing and environmental**, not caused by any
dependency change:
- 4 × `Embedding API` tests — time out (need the embed-iframe config / `.env.embed`).
- 2 × `Image equality – howdy-yall` (typed + untyped) — headless-Chrome
  rasterization renders the image 2px taller (`different-size([247,55],[247,57])`).

The pass criterion is therefore: **no new failures beyond this 6-failure
environmental baseline.**

**RESULT: no regression.** After all dependency changes, `npm run mocha` produced
`305 passing / 45 pending / 6 failing` — the *exact same* failure set as the
baseline (verified by diff). The dependency remediation introduced **zero new test
failures**.

## Changes most likely to break something NOT covered by mocha

The Selenium suite exercises the editor, REPL, charts, check-blocks, errors,
tables, world, type-check, and the share-URL proxy. It does **not** exercise the
React dashboard, Google Drive/Sheets auth, the embed iframe (those tests fail
environmentally), or any `npm run lint`. So the following changes carry risk the
test suite would not catch:

1. **`eslint` 6 → 9 (highest-confidence breakage).** ESLint 9 defaults to *flat
   config* (`eslint.config.js`); this repo has a legacy `.eslintrc`. `npm run lint`
   (`eslint src`) will likely error until either a flat config is added or
   `ESLINT_USE_FLAT_CONFIG=false` is set. Not a mocha test, so unverified here.
2. **`googleapis` kept at 46, with `node-forge`/`json-bigint` overridden.** Drive
   and Sheets auth (`src/google-auth.js`, `src/storage/`) are not covered by mocha
   (`test/sheets.js` is `xdescribe`, OAuth needs live credentials). `node-forge`
   1.3→1.4 is only used on the P12 service-account-key path (`google-p12-pem`),
   which CPO's OAuth web flow doesn't hit — but this is untested.
3. **`request` → `form-data` 4.** Only matters if `request` is ever used for a
   multipart upload. `proxyStreamFetch` (the SSRF image/share proxy) only does
   GETs, and the share-URL GET path *is* covered by `test/shareUrls.js` (passing),
   but a multipart code path, if any, is not.
4. **Global `semver` 5 → 7 override.** Affects the `heroku` CLI and any other
   consumer pinned to semver 5. `heroku local:run` is exercised by every mocha run
   (the test script uses it) and works, but other `heroku` subcommands are not.
5. **`node-fetch` 2.6.x override under `isomorphic-fetch`/`fbjs` (React 15).** The
   React dashboard isn't tested by mocha at all. In the browser bundle this path
   uses `whatwg-fetch`, so the override should be inert, but it's unverified.
6. **`mocha` 8 → 11 and `tmp`/`selenium` temp-dir override.** These affect the test
   harness itself; covered transitively by the suite running green, but worth
   noting the runner changed.

Lowest-risk changes (functionally verified by the green suite or trivially
inert): the dead-dep removals, `style-loader`/`requirejs`/`sha.js` bumps,
`git-rev-sync` 1→3 (`run.js` boots and serves), and the `audit fix` upgrades to
`vega`/`axios`/`express`/`ws`/`lodash` (charts, errors, world tests pass).

jpolitz:
(Written by Claude)

# lang/ — npm audit remediation report

Date: 2026-06-25
Scope: `lang/` (the Pyret compiler + runtime) in the pyret-lang monorepo.
Task: run `npm audit`, fix all **critical** and **high** severity issues, keep
`make all-pyret-test` passing (and not regress `npm run mocha` in
`code.pyret.org/`), and document anything interesting or risky. Companion to
`code.pyret.org/SECURITY-AUDIT-REPORT.md`.

## Result at a glance

| Severity | Before | After |
|----------|-------:|------:|
| critical | **1**  | **0** |
| high     | **10** | **0** |
| moderate | 5      | 17 — (out of scope) |
| low      | 5      | 4  — (out of scope) |
| total    | 21     | 21 |

**Every critical and high vulnerability is fixed**, with **no major-version
upgrades** — every change is semver-compatible (`npm audit fix` with no
`--force`). The single critical (`shell-quote`) and one of the highs
(`minimatch`) were the *same bundled-dependency wall* that blocked
`code.pyret.org`; this repo is where that wall is defined, so here it could
actually be removed.

The fixes were verified by running the full `make all-pyret-test` suite before
and after the change and confirming an identical pass set (see "Verification").

## Approach

Same guiding principle as the CPO pass: **fix the vulnerabilities without
changing runtime behavior.** In priority order:

1. **Remove the `bundleDependencies` wall** (the one structural change — see below).
2. **`npm audit fix`** (semver-compatible upgrades only; no `--force`).

No `overrides` and no direct version-range bumps in `package.json` were needed —
once the bundle wall was gone, every fix was available in-range. The only
`package.json` edit is the three-line removal of `bundleDependencies`. All
dependency version changes live in `package-lock.json` and are within the
existing declared ranges.

## What was interesting

### 1. The critical was the bundled-`browserify` wall — and this is its origin
The CPO report's headline residual issue was: `pyret-lang`'s
`"bundleDependencies": ["browserify"]` freezes `browserify`'s whole subtree
(`shell-quote`, `minimatch`, `pbkdf2`, `sha.js`, …) *inside the package*, marked
`"inBundle": true` in the lockfile, where **npm `overrides` and `npm audit fix`
cannot touch them**. That manifest is **this** `lang/package.json`.

`npm audit fix` confirmed the wall is self-inflicted here too — before the change
it printed, for `shell-quote`/`minimatch`/`bn.js`/etc.:

> `… is a bundled dependency of lang@0.0.0 … It cannot be fixed automatically.`

The CPO report's recommended upstream fix was *"move `browserify` out of
`bundleDependencies` — or simply un-bundle it."* That is exactly what was done:
removing the `bundleDependencies` array (leaving `browserify` in `dependencies`)
un-freezes the subtree, after which `npm audit fix` patches it normally. This
fixes the `shell-quote` **critical** and `minimatch` **high** here, and — once a
consumer reinstalls `pyret-lang` — removes CPO's 3 critical + 1 high residuals as
well.

### 2. `minimatch` was fixable in-range *without* the glob/mocha trap
The CPO report flagged that `minimatch <=3.1.3` had "no patched 3.x", forcing a
`mocha` 8→11 major because `glob@7` calls `minimatch` as a *default-callable
function* that 5.x+ removed. That turned out to be stale: **`minimatch@3.1.5`
exists**, is outside the advisory range (`<=3.1.3`), and **still exports the
callable function API** `glob@7` needs. So `browserify → glob@7 → minimatch`
went `3.1.2 → 3.1.5` in place — patched, no `glob`/`browserify` upgrade. (Verified
directly: `browserify`-side `minimatch@3.1.5` resolves and `typeof minimatch ===
'function'`.) The unrelated `minimatch@9.0.5 → 9.0.9` copies under `jest` were
bumped the same way.

### 3. `shell-quote` had a clean in-range patch
`browserify` declares `shell-quote@^1.6.1`; the patched `shell-quote@1.9.0`
satisfies that, so the critical (newline-injection in `quote()`) cleared with a
plain in-range bump — no override, no `browserify` change.

### 4. The `vega` highs cleared via a minor bump, dragging the whole vega tree
The four `vega` highs (XSS via `VEGA_DEBUG` / `setdata` / `vlSelectionTuples`)
all live in `vega@<=6.1.2` and its `vega-functions`/`vega-expression`/
`vega-selections` leaves. There is no `6.1.3` patch — the 6.x line went
`6.1.2 → 6.2.0` — so `vega` (a direct dep, `^6.1.2`) moved to `6.2.0`, and the
~30 `vega-*` subpackages bumped one minor in lockstep. `vega` is only used by the
chart runtime (`src/js/trove/charts-lib.js`, `src/arr/trove/charts.arr`), which
is **not** exercised by `make all-pyret-test`, so this is the highest-residual-
risk change (see "Changes most likely to break something not covered by tests").

### 5. A harmless side effect: `node-fetch@3` dropped out
Bumping `vega-loader 5.0.0 → 5.1.0` removed its `node-fetch@^3.3.2` dependency
(and the `fetch-blob`/`formdata-polyfill`/`web-streams-polyfill`/
`data-uri-to-buffer`/`node-domexception` polyfills it pulled). The runtime's
actual fetch path — `cross-fetch@4 → node-fetch@2.7.0` — is **unchanged**. The
`pyret-io-test` remote-URL import tests (which do real fetches) still pass, so
this is inert for behavior.

## Changes made

### Removed
| Change | Why | Cleared |
|--------|-----|---------|
| `bundleDependencies: ["browserify"]` (deleted from `package.json`) | un-freezes the `browserify` subtree so it can be patched | enabled the `shell-quote` critical + `minimatch` high fixes |

`browserify` itself **stays in `dependencies`** — it is still installed and used
at build time (`make` runs `npm exec browserify …` to bundle the standalone), so
the build is unchanged.

### `npm audit fix` (semver-safe, no `--force`) — critical/high-relevant bumps
| Package | From → To | Path | Severity cleared |
|---------|-----------|------|------------------|
| `shell-quote` | 1.8.3 → 1.9.0 | `browserify` | **critical** |
| `minimatch` | 3.1.2 → 3.1.5 | `browserify → glob@7` | high |
| `minimatch` | 9.0.5 → 9.0.9 | `jest` | high |
| `glob` | 10.4.5 → 10.5.0 | `jest` | high (CLI cmd-injection) |
| `picomatch` | 2.3.1 → 2.3.2, 4.0.3 → 4.0.4 | `jest` | high |
| `lodash` | 4.17.21 → 4.18.1 | `benchmark` | high |
| `ws` | 8.18.3 → 8.21.0 | direct dep (`^8.18.0`) | high |
| `tar-fs` | 2.1.3 → 2.1.4 | `canvas` | high |
| `vega` | 6.1.2 → 6.2.0 | direct dep (`^6.1.2`) | high (XSS) |
| `vega-functions` | 6.0.0 → 6.1.1 | `vega` | high (XSS) |
| `vega-expression` | 6.0.0 → 6.1.0 | `vega` | high (XSS) |
| `vega-selections` | 6.0.0 → 6.1.2 | `vega` | high (XSS) |

(Plus the in-lockstep `vega-*` minor bumps, and patch bumps to the bundled
`crypto-browserify` subtree — `pbkdf2`, `ripemd160`, `hash-base`, `create-hash`,
`parse-asn1`, `browserify-sign`, `bn.js` — and the `jest`/`@babel/*` dev tree.
None are critical/high; all are semver-compatible.)

Nothing was upgraded across a major version. `jest@30` and `browserify@17` (the
two heaviest trees) stayed put.

## Residual issues (out of scope)

The task was critical + high; the **17 moderate / 4 low** that remain are all in
the **dev/build** tree and require breaking (`--force`) major upgrades:

- **moderate** — `js-yaml` quadratic-DoS via merge keys, reached through the
  `jest@30 → @jest/* → @babel/*` chain. `npm audit fix` reports it as fixable
  only via `npm audit fix --force`, which would force a breaking change in the
  `jest`/`babel-jest` majors. `jest` is the `pyret-io-test` runner, so this was
  left for a deliberate follow-up rather than risk the test harness.
- **low** — `elliptic` "risky cryptographic primitive", inside the bundled
  `crypto-browserify` subtree of `browserify` (build-time crypto shim only).

The moderate count rose 5 → 17 because `npm audit fix` advanced the `jest`/`babel`
dev tree into versions that surface the `js-yaml` advisory more widely. This is
the same "audit fix slightly increases moderates" effect noted in the CPO report;
it introduces **no new critical/high**.

## Effect on `code.pyret.org`

CPO's `npm run mocha` behavior is **unchanged** by this work: CPO has its own
`package-lock.json` pinning `pyret-lang` to a specific git commit, and nothing
here regenerates it. No files under `code.pyret.org/` were touched.

The forward benefit: this change *is* the upstream fix the CPO report asked for.
The next time CPO (or any consumer) reinstalls `pyret-lang@<this commit>`, it will
receive an **un-bundled, dedupable, patchable** `browserify` subtree instead of
the frozen `sha.js@2.4.11`/`pbkdf2@3.1.2`/`shell-quote@1.8.3`/`minimatch@3.1.2`
copies — eliminating CPO's 3 critical + 1 high residuals at the source.

## Verification

Environment: this VM has a full working `lang/` build (`build/phaseA`, the 52 MB
`tests/pyret/all.jarr` standalone, native `canvas`, Chrome not required for this
suite). `make all-pyret-test` builds the standalone and runs three test groups:
`parse-test` (jasmine), `pyret-io-test` (jest, with real remote-URL fetches), and
the main `tests/pyret/all.jarr` Pyret suite.

`make all-pyret-test`, before vs after the dependency changes:

| Group | Baseline (before) | After fixes |
|-------|-------------------|-------------|
| `parse-test` (jasmine) | 60 specs, 0 failures, 2 pending | **60 specs, 0 failures, 2 pending** |
| `pyret-io-test` (jest) | 13 passed / 13 total | **13 passed / 13 total** |
| `tests/pyret/all.jarr` | "all 13360 tests passed" | **"all 13360 tests passed"** |
| exit code | 0 | **0** |

**RESULT: no regression.** The full suite produced an identical pass set after
the dependency changes. The remediation introduced **zero new test failures**.

## Changes most likely to break something NOT covered by `all-pyret-test`

The suite exercises the parser, the compiler, the runtime, check-blocks, spies,
type-checking, the library, and the IO/remote-import path. It does **not**
exercise chart rendering or any browser-side bundle, so these carry risk the
suite would not catch:

1. **`vega` 6.1.2 → 6.2.0 (highest-residual risk).** A minor bump of the entire
   `vega` charting tree. Charts (`charts-lib.js` / `charts.arr`) render in the
   browser via `canvas`/Vega and are not in `all-pyret-test`. No `6.1.x` patch
   exists, so this minor was the only way to clear the four `vega` XSS highs.
2. **Bundled `crypto-browserify` subtree patch bumps** (`pbkdf2`, `ripemd160`,
   `hash-base`, `create-hash`, `parse-asn1`, `browserify-sign`, `bn.js`). These
   are browserified into the standalone's Node-crypto shim. Patch-level only, and
   the standalone built and ran the full suite green, but the in-browser crypto
   path (vs the Node test path) is not separately exercised.
3. **Un-bundling `browserify` (the structural change).** For `lang`'s own build
   this is inert (`browserify` is still installed and invoked identically). The
   only observable change is to *consumers* of the published/`git` package, who
   now resolve `browserify`'s deps normally instead of receiving frozen copies —
   intended, and a security improvement, but worth re-testing CPO's `npm install`
   + `npm run mocha` whenever CPO next bumps its `pyret-lang` pin.

Lowest-risk changes (functionally verified by the green suite or trivially
inert): `shell-quote`/`minimatch`/`glob`/`picomatch`/`lodash`/`tar-fs`/`ws`
bumps, and the `jest`/`@babel/*` dev-tree bumps (the `jest` runner ran the
io-tests green).

code.pyret.org/package.json
```diff
@@ -84,13 +81,30 @@
   },
   "repository": {
     "type": "git",
-    "url": "git://github.com/brownplt/code.pyret.org"
+    "url": "git+https://github.com/brownplt/code.pyret.org"
```
blerner: Why is this library still here, given a monorepo?

codemirror-mode/package.json
```diff
@@ -7,7 +7,7 @@
   "main": "mode/pyret.js",
   "repository": {
     "type": "git",
-    "url": "git://github.com/brownplt/pyret-codemirror-mode"
+    "url": "git+https://github.com/brownplt/pyret-codemirror-mode"
```
blerner: Ditto

docs/package.json
```diff
@@ -3,15 +3,15 @@
...
+    "pyret-codemirror-mode": "file:../codemirror-mode",
+    "pyret-embed": "file:../embed"
   },
   "scripts": {
     "build": "make"
   },
   "repository": {
     "type": "git",
-    "url": "git://github.com/brownplt/pyret-docs"
+    "url": "git+https://github.com/brownplt/pyret-docs"
```
blerner: Ditto

pyret.org/package.json
```diff
@@ -1,17 +1,17 @@
...
+    "pyret-codemirror-mode": "file:../codemirror-mode",
+    "pyret-embed": "file:../embed"
   },
   "scripts": {
     "build": "make"
   },
   "repository": {
     "type": "git",
-    "url": "git://github.com/brownplt/pyret.org"
+    "url": "git+https://github.com/brownplt/pyret.org"
```
blerner: Ditto

jpolitz:
@blerner I'm going to merge to `drydock` (but not to `mainmast`) because I need to check a few things about deployment, all testable with horizon's heroku instance, and that's the sensible way to do it.

code.pyret.org/package.json
jpolitz: https://github.com/brownplt/pyret-lang/pull/1881/commits/1f6485601b6701c53252d3459f8f94080000b82a

PR #1893, opened 2026-07-17
Fix line plot labels

blerner:
> is `from-list.labeled-line-plot` supposed to show labels onHover?
>
> It doesn’t right now. I can easily duplicate the behavior by overlaying it atop a labeled-scatter-plot, and the labeled-line-plot method isn’t even documented so maybe we should remove it and I can use the overlayed plot instead?

This would _almost_ work, except that it would create *two* plots and therefore affect the legend.  It also means that `line-plot().labels()` was also broken and should be fixed.

This PR fixes labels to work properly by ensuring that the dots will be visible.  It also cleans up some redundant list-element checks (that our chart-utils annotations already check for us)

### DIFF
--- a/lang/src/arr/trove/charts.arr
+++ b/lang/src/arr/trove/charts.arr
@@ -59,9 +59,6 @@ end
 # HELPERS
 ################################################################################
 
-fun check-num(v :: Number) -> Nothing: nothing end
-fun check-string(v :: String) -> Nothing: nothing end
-fun check-image(v :: IM.Image) -> Nothing: nothing end
 
 fst = raw-array-get(_, 0)
 snd = raw-array-get(_, 1)
@@ -1887,8 +1884,6 @@ fun line-plot-from-list(xs :: CL.LoN, ys :: CL.LoN) -> DataSeries block:
   when xs.length() <> ys.length():
     raise(ERR.message-exception('line-plot: xs and ys should have the same length'))
   end
-  xs.each(check-num)
-  ys.each(check-num)
   default-line-plot-series.{
     ps: map4(get-scatter-point, xs, ys, xs.map({(_): ''}), xs.map({(_): none}))
   } ^ line-plot-series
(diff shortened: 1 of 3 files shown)


lang/src/js/trove/charts-lib.js
```diff
@@ -2663,17 +2663,10 @@
       const defaultColor = config.defaultColor || default_colors[0];
       const lineWidth = toFixnum(get(rawData, 'lineWidth'));
       const color = getColorOrDefault(get(rawData, 'color'), defaultColor);
-      const asScatterPlot = scatterPlot(globalOptions, rawData, config);
+      debugger
```
jpolitz: Stray debugger

lang/src/js/trove/charts-lib.js
```diff
@@ -2663,17 +2663,10 @@
...
+      const pointSize = getNumOrDefault(get(rawData, 'point-size'), 0);
+      const asScatterPlot = scatterPlot(globalOptions, rawData.extendWith({'point-size': pointSize}), config);
       const marks = asScatterPlot.marks;
-      const markShapeMarks = marks.find((m) => m.name === `${prefix}ShapeMarks`);
```
jpolitz, line 2668: What were these doing? Just unnecessary now?

jpolitz (commented):
Pushed a test + fix for a `or-else` ordering thing separately, plus two small comments.

lang/src/js/trove/charts-lib.js
blerner: Yeah - these _had been_ trying to make the dots invisible by default and visible on hover...but (a) that's impossible behavior and (b) it's not the newly desired behavior.

lang/src/arr/trove/charts.arr
```diff
@@ -1573,7 +1573,7 @@ data DataSeries:
       end
       self.constr()(self.obj.{
           ps: map2({(val, label): val.{label: label}}, self.obj!ps, labels),
-          point-size: self.obj.point-size.or-else(some(default-scatter-plot-series.point-size))
```
blerner: oops, good catch.  (In the back of my head, I _thought_ this use of `.or-else(some(...))` looked funny, but I didn't pay it enough attention.)

jpolitz (approved):

PR #1896, opened 2026-07-22
Avoid iterator helpers in charts-lib for older-Safari compatibility

jpolitz:
Three spots in charts-lib.js called `.map()` directly on an iterator (`rawCounts.entries()` in the image-dot-plot code, and `Array(numSamples).keys()` in the function-plot sampling code). That relies on the iterator-helpers proposal (`Iterator.prototype.map`), which Safari only shipped in 18.4 — on Safari 17.6 it throws `TypeError: rawCounts.entries().map is not a function`, which teachers hit on pyret.bootstrapworld.org.

Moving the spread inside the brackets (`[...m.entries()].map(...)`) converts the iterator to an array first, which is functionally identical and works in all browsers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01UukhBiaHkzTeBba8vCojtZ

### DIFF
--- a/lang/src/js/trove/charts-lib.js
+++ b/lang/src/js/trove/charts-lib.js
@@ -2051,7 +2051,7 @@
         p.value = countForCat;
         rawCounts.set(p.category, countForCat + 1);
       }
-      const counts = [...rawCounts.entries().map((kv) => ({ category: kv[0], count: kv[1] }))];
+      const counts = [...rawCounts.entries()].map((kv) => ({ category: kv[0], count: kv[1] }));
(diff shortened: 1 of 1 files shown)


blerner (approved):
That's fine; I was trying to avoid creating two arrays, if I could avoid it...but cross-browser compatibility is more important here

jpolitz:
Makes sense, the code you wrote is eminently plausible! Also hard to test older Safari in particular in a test matrix without lots of infra.

PR #1902, opened 2026-07-29
Opt-in (for now) TypeScript compiler implementation

jpolitz:
So, uh, @shriram, I guess this is happening?

Ben and I have been reviewing this for some time.

Why do it:

1. Dramatically faster load times/smaller deploy bundles (5.6MB → 3.4MB
   gzipped, average time to editor ready in the headless CI 2.1s → 1.1s)
2. Dramatically faster *compile times* for user code (e.g. `core.arr` Bootstrap
   library 9s → 2.5s, main Pyret test bundle 100s → 20s)
3. Seconds-long build times of the compiler itself (90s pyret compiler → 13s
   `tsc` [really 3s `tsc` and 10s other build stuff])
4. Improved people + agents contributions. This is fuzzy; keeping
   Pyret self-hosted arguably has long-term benefits here. But (a) the type
   checker really does aid development and (b) I think if our endgame
   is “fast and type-checked Pyret implementation of Pyret” routing through
   another language for a bit isn't a bad idea.

For confidence:

- the *same* headless tests run under both the TS and Pyret configurations
- the two compilers fixpoint in a byte-identical bootstrap of the existing
  pyret compiler
- the TS compiler is opt-in in all the platforms (`?compiler=ts` for browser,
  `compiler: 'ts'` in the embed API, a CLI flag, a VScode config option), all
  systems default to Pyret-of-today and toggle to TypeScript
- we added more headless test environments (beyond existing browser CPO):
  vscode, vscode-mimicking Gitlab, and iframe embedded both as a hosted server
  and a static html file
- more tests that mimic or directly import Bootstrapworld starter files

IDK about you but this is my first +50KLOC production deploy that's mostly
agent-generated. So this PR is the start of a careful release plan:

- merge and deploy with TS off, make sure the toggle default doesn't break
  anything
- turn *on* typescript on horizon by default to check everything about toggling
  it. Run tests against it
- turn *on* typescript by default in embed and vscode only. Monitor for breakage
- turn *on* typescript by default on CPO, potentially with some kind of A/B
  split (or maybe at this point we're confident)
- (tell @schanzer to) turn *on* typescript by default on PBO just by pulling
  the mirror of CPO and config over

We'll probably add more tests during this process. I think the remaining
risk is in:

- Getting the TS compiler on the CDN and served everywhere correctly with
  the correct config (just a lot of environment variables and S3 buckets to get
  exactly right) alongside the existing one
- The fact that the TS compiler doesn't have stack management! See below.

Commentary on the code + review:

- The two worst parts of the port correspond exactly to places where Pyret's
  stack management helped:
  - Had to de-recursify ANF because it went too deep on large files (we've
    tested on lots of large files and I think it's OK, but this is still a
    risk)

    https://github.com/jpolitz/pyret-lang/blob/2ab90ca781e966c2e96c2e9ae3f55a46c873bd11/lang/src/ts-compiler/src/anf.ts#L150
  - Part of the module fetch was async in the middle of the compiler (e.g.
    fetching `url` imports), but the TS port is mostly *sync*. So there was no
    way to easily do a `fetch` synchronously without an `async`-poisoning
    rewrite of a very large context. Fable did what any of us would do, and
    **used `execFileSync` to start a node subprocess synchronously in the
    middle of the compiler for each `fetch`** (one should not do this). It
    turns out really the async bit can be separated, so we now have an async
    module-DAG step followed by pure sync backend codegen that looks at the
    snapshot of what was fetched at the beginning (which is better anyway
    unless you have a *really big* module graph and wanted to stream it all
    through... but we don't have module graphs that big).

    https://github.com/jpolitz/pyret-lang/pull/5/changes/d27c1956c4641f43da71a334d74ed851dc74a921
- There were a *ton* of little fixes that went along with this that agents
  helped with a lot. These included:
  - Getting everything canonicalized (e.g. print in key-sorted order rather
    than whatever's in the map) to get to byte-identical output for confidence
  - Lots of TypeScript patterns around error messages, including converting JS
    errors back to Pyret errors to seamlessly work with the error renderer in
    CPO
  - Lots of places where there was a bare untyped `obj.field` in the Pyret
    compiler and it wasn't totally clear that it was safe to project out the
    `.field` statically in TS, so casts resulted

- Aside from the two stack-management correlated issues, there were some
  agentic gotchas:
  - It stubbed out significant parts of the REPL tests, basically mocking out
    the runner entirely, which had to be added back

    https://github.com/jpolitz/pyret-lang/pull/5/changes/0f1d3d922a40a39f7971c37cfc57b9c93c4cab4f
  - Lots of little parity issues around what `to-repr` did in Pyret vs whatever
    stringification was happening in TS (this is just for internal compiler
    representations, and error messages, not actual Pyret runtime values).
    There were like 6 different implementations of "torepr"-y things:

    https://github.com/jpolitz/pyret-lang/pull/5/changes/043ceab4422ac5ad9479650ec1d47d23bd70b3d4


shriram:
Do you want to set up an experimental URL for this, and I can try it out in 0190?

jpolitz:
When does it start? My hope would be to just share e.g. `https://code.pyret.org/editor?compiler=ts` with them as the "experimental" URL (which works with shareurl fragments after it). That should be doable after step 1 of the plan, which would make it opt-in-by-url-param, opt-out by default.

shriram:
Classes begin Sep 9.

We are just about finished with placement (in 2-3 days). After we've sent out decisions, I can certainly ask students to bash at it. A handful might… We can also get @ds26gte to run all prior solutions from his private repo, which would be a very useful test.

jpolitz:
Question for @shriram and @blerner about git history and cleanup here.

Right now this all lives on a fork; we can do arbitrary rewrites before merging.

1. There are lots of "Joe + Claude" commits, which are a mix of automated Claude-work-in-progress commits and then actual co-authored commit messages with Joe says/Claude says where I wrote things out directly.
2. There are also a handful of markdown documents that are process-journal-ish. For example, `CONVENTIONS.md` is a doc about how Pyret mapped to TypeScript, and was *not* written by me, but by Claude as a way to do a multi-agent coordination. The `port-*.md` have notes about todo lists and things we/agents noticed during review that we checked off.

I torn between two history options:

a. Leave it as-is. We're learning how these processes work and we can keep this history in place. It's maybe useful for others who find the public repo and want to do archaeological stuff and figure out how agents worked (or didn't) for us.
b. Squash it all, write one big commit message for the whole campaign myself at the top, and leave the concatenation of all the squashed commit messages in the big message as the step-by-step record.

I don't think maintaining the (2)-style docs in-tree is is a good idea; **I'm** certainly not good at avoiding drift in documents like that and my instinct is that agents will drift faster if these docs proliferate. I also have some experience that these are *bad* things for agents to pick up in their context, because they skew towards overly strong factual claims that get too much weight. Code-as-truth is better for agent contexts (and, I'd argue, for people), with any summarizing/invariant communication done in the prompt rather than picked up implicitly from the codebase. I'm not sure if there's a *different* place we should store them for posterity, or if we just let them get GCed. **Do either of you care if I just prune them out on merge and don't keep them around?**

blerner:
Third option: we take those .md files and move them into the _docs_ folder,
in a subfolder called "code-archaeology", so that we have them later, but
they're not part of the in-src/ files that Claude manipulates?

jpolitz:
I'm not sure we want a AGENTS.md that has to exclude directories, and that agents won't pick things up in there, etc. All `grep`s have to exclude it (including over git history), and so on.

jpolitz:
@blerner if you're arguing that the history is valuable (which I'd buy), I might even suggest a different repo for it. `pyret-scrolls` or something (though maybe we should save that for our TUI). I'm worried about the accretion over time in the actual repo.

blerner:
_Maybe_ a different repo, but we've gone to such pains to make a mono-repo it feels odd to deliberately create a separate one...

Maybe we use GitHub Discussions for these design descriptions? (I know we don't use those very much, but at least they're still associated with the repo...?)

jpolitz:
Yeah that's not a bad idea. Then you could choose to point an agent at it (and have it update TODOs, etc), **or not** if you want a clean context that just considers the code and not a bunch of potentially out-of-date discussion. And it keeps it ~with the repo. It's tied to Github, but so much of what we do is tied to Github.

It's also explicitly dated (rather than implicitly by `git blame`)

jpolitz:
Oh I figured out something I think handles both well.

We can make it be an archival tag. Basically squash merge here and remove all the cruft/write a good overall comimt message, but *also* make a tag on the commit *before* any deletion and push that.

Then we/others can look it up later via tag (and we only need to remember "search the tags") and the tag serves as a root for reachability so things don't get GCed.

It never shows up on the drydock history because we squash, so that throughline of “the main line code” isn't polluted with it. All stays in-repo and would transfer across services, etc, with a full clone.

jpolitz:
@shriram @blerner  Here's what I want to put in the squash commit message, which is worth commenting on for posterity:


```
TypeScript compiler implementation and related test infrastructure

# What is in the commit

- .ts files that implement the same logic as the .arr parts of arr/compiler
- Flags for selecting this implementation in all the relevant components:
  - Via flag at the CLI
  - Via user config in the VScode extension
  - Via URL parameter in code.pyret.org
- Substantial new testing infrastructure spanning these components (some of
  which did not have automated tests before) and the TS port specifically

# Who wrote this

This is a squashed merge commit for a large, mostly agentic (Claude Opus 4.8
and 5; Fable 5) development process. The full commit chain is available at the
tag ts-port-archive for posterity and transparency.

The prose of this commit message is written entirely by me (Joe Politz) along
with Ben Lerner. Claude helped fill in some commit hashes, code snippets, and
ASCII diagrams.

# Why bother

TypeScript (and JavaScript) simply have benefits for running the compiler in a
browser compared to the performance of Pyret:

1. Dramatically faster load times/smaller deploy bundles (5.6MB → 3.4MB
gzipped, average time to editor ready in the headless CI 2.1s → 1.1s)

2. Dramatically faster compile times for user code (e.g. core.arr Bootstrap
library 9s → 2.5s, main Pyret test bundle 100s → 20s)

3. Seconds-long build times of the compiler itself (90s pyret compiler → 13s
tsc [really 3s tsc and 10s other build stuff])

There's a less objective argument that people and agents will be better at
contributing to a TypeScript codebase. I'm not sure that outweighs all the
benefits of being self-hosted. However, the objective benefits above are enough
of a shift *today* that it's questionable why we would *not* give our users
this experience (e.g. compile and load time on a Chromebook/old iPad matters in
the classroom).

# How we did it

The lineage of this commit comes from an overnight goal prompt given to Fable
on June 10-11, 2026:

  /goal Port the compiler in pyret-lang to TypeScript. Leave the runtime
  libraries (e.g. trove/) in Pyret, but make everything in the
  arr/compiler implementation match what's there, but be in clean
  TypeScript. Test by adding new Makefile targets that make sure the
  same compiler options can be used with the new typescript compiler.
  Organize files however you like, feel free to copy things, but be
  strictly additive so it's obvious that nothing existing breaks. Add
  new tests as necessary to check things. Existing tests should work,
  though if you find hard constraints (e.g. tests that test the compiler
  from Pyret) you can add new tests that demonstrate the same behavior
  on the new version instead of dealing with FFI crossover issues. Don't
  change the runtime representations of values, and the same runtime.js,
  stack management, etc should all work.

That produced this commit: ec1034ebf ("Add TypeScript port of the compiler
(lang/src/ts-compiler)", reachable via the ts-port-archive tag)

Not bad! This is better than Ben and I managed by hand with the time we had
available over some summers. We've definitely written better prompts since
then; in late spring it was compelling enough to work forward from.

The past few months have been a lot of verifier-writing,
test-infrastructure-checking, and maybe a little soul-searching. These will
certainly be the first lines of code in the codebase that Ben or I did not
write or read (we have not read all of this code). We did agentic review, wrote
a lot of tests (or caused Claude to), and read code where things that seemed
off.

We've considered going back and re-prompting to see if we can get a better
version from scratch, but at this point we've learned so much and invested
enough review that I think it's local-maximum-enough, so here we are.

# Where things stand

We're comfortable merging this. Here we report on:

- What's in the code
- The verification infrastructure
- Where we found significant issues with the port

## Ported code

The port is mostly boring, as it should be. There are translations of Pyret
patterns to TypeScript patterns. For example, a data declaration in Pyret's
compiler, used for different kinds of names that can appear in bindings:

  # src/arr/trove/ast.arr
  data Name:
    | s-underscore(l :: Loc) with: ...
    | s-name(l :: Loc, s :: String) with:
      method key(self): "name#" + self.s end
      ...
    | s-atom(base :: String, serial :: Number) with:
      ...
      method key(self): "atom#" + self.base + "#" + tostring(self.serial) end
  sharing:
    method _lessthan(self, other): self.key() < other.key() end,
    ...
    method _equals(self, other, eq): eq(self.key(), other.key()) end,
  end

Translates to TypeScript:

  // src/ts-compiler/src/ast.ts
  export abstract class NameBase {
    abstract get $name(): string;
    ...
    abstract key(): string;
    ...
    lessthan(other: Name): boolean { return this.key() < other.key(); }
    ...
    equals(other: Name): boolean { return this.key() === other.key(); }
  }

  export class SName extends NameBase {
    get $name(): 's-name' { return 's-name'; }
    constructor(public l: Loc, public s: string) { super(); }
    key(): string { return 'name#' + this.s; }
  }
  ...
  export class SAtom extends NameBase {
    get $name(): 's-atom' { return 's-atom'; }
    constructor(public base: string, public serial: number) { super(); }
    ...
    key(): string { return 'atom#' + this.base + '#' + String(this.serial); }
  }

  export type Name = SUnderscore | SName | SGlobal | SModuleGlobal
                   | STypeGlobal | SAtom;
  export function isSName(x: any): x is SName { return x instanceof SName; }

(A brief aside. You might say “Joe, why don't Pyret's data definitions just
*compile* to that? It seems like you ought to be able to trivially get the same
TypeScript-y output, just as fast and terse, with a straightforward compiler.
Hold that thought through the rest of this, but a short answer for now: Pyret
has many, many dynamic features that demand much more of the compiler than this
direct translation.)

The bulk of the port is like this – there's an obvious TypeScript analogue for
each Pyret form. This includes data with variants, tons of standard recursive
pattern matching using switch over the $name field, visitor-pattern
infrastructure for traversing and patching ASTs, unsurprising I/O, and so on.
This is reassuring on its own. There are nitpicks and cleanups we may get to
(it's not necessary for $name to be a getter, for instance).

Unsurprisingly for a pattern translation, the ported code is a bit longer
(31kloc .ts vs. 26kloc .arr), though it differs depending on file. TypeScript
is also a really good fit here: its type system easily expresses the type-based
parts of Pyret, with easy escape hatches for untyped code or
challenging-to-type-check patterns (which the Pyret compiler has).

## Testing and Verification Strategy

We're guided by a few main questions:

- Is the new compiler correct, in that it produces the same output as the
  existing one?
- What differences will users notice (other than the speedups), if any?

### Output Comparison

- Is the new compiler correct, in that it produces the same output as the
  existing one?

Up to the programs in this repository, yes! One of the most useful things about
a port like this is the presence of a strong oracle. The strongest result for
clear correctness we have is that the TS compiler and the .arr compiler can
both compile the .arr compiler, which can compile itself, and the two paths
result in *byte identical compilers*. So even if we ditched the existing
bootstrapped compiler blobs, we could still move forward with the TS
implementation confidently, at least for getting re-bootstrapped.

              src/arr/compiler/*.arr (compiler source)
                     |                        |
        compiled by phaseA         compiled by the TS compiler
        (.arr compiler, built by   (built by tsc alone -- no
        the committed phase0 blob)  Pyret used)
                     |                        |
                     v                        v
             phaseB/pyret.jarr        phaseB-ts/pyret.jarr
                     |                        |
        each compiles the same compiler source again
                     |                        |
                     v                        v
             phaseC/pyret.jarr        phaseC-ts/pyret.jarr

        phaseB == phaseC         (.arr self-fixpoint)
        phaseB-ts == phaseC-ts   (TS self-fixpoint)
        phaseB == phaseB-ts      (the headline: byte-exact drop-in)

        => all four standalones are one byte-identical fixpoint,
           reached from two independent seeds

This can be run with `make bootstrap-converge` (in lang/), and CI runs it on
every push as the "bootstrap convergence (.arr == TS, byte-identical)" job.

This took a little bit of work, mainly around canonicalizing a lot of
unspecified orderings in the compiler.  Two specific examples:

- String-keyed dictionaries abound in the compiler, as do iterations over them.
  Pyret has its own HAMT string dict implementation, while the TS
  implementation chose JS Map, which have different iteration orders. We added
  a lot of sorting to iteration points to make sure outputs would be identical.
- Gensymed identifiers in Pyret are eagerly created as needed. This means a
  subtle ordering difference in an AST traversal that would not otherwise
  affect *correctness* of the generated code *could* change the variable names
  in it. Agents did most of this work, and it put quite a tight constraint on
  the implementation to traverse ASTs and generate IDs in precisely the same
  way.

There are also a handful of other tests (see `ts-parity-test`, `ts-wf-parity`,
`ts-type-check-parity`) that check byte-for-byte parity of output for cases not
covered by the compiler bootstrap. These were mostly added later on (see the
issues section below).

### The other 80%

- What differences will users notice (other than the speedups), if any?

The bootstrap fixpoint gives a lot of confidence in the happy path of code
generation at the command line. That's the core compiler correctness result.
But there are many things Pyret does and many ways users interact with Pyret
that aren't tested at all by this path:

- Loading and running the compiler in the browser, in WebViews via the embed
  API and VScode
- Loading libraries and starter code over shared GDrive links and URLs
- Interaction with media like animations via reactors, and spreadsheets
  via tables
- Rich error renderings, which we've put significant effort into

We *do* have a pretty good history of writing headless browser tests for things
like this; we have assertions that check the DOM rendering of many errors.
Historically, these all worked on headless Chrom(e|ium) runs over a selenium
driver, pointed at a development server running CPO. The embedding API and
VScode mode were implicitly “tested” under the assumption that once they
loaded they'd behave more or less like the headless tests in-browser.

For a change of this magnitude we wanted a bit more confidence.  These are now
unified in a single test directory that can be pointed at each of the different
deployment contexts (vscode, browser, embed), in either mode, testing the cross
product:

{ vscode, browser, embedded iframe } ⨯ { typescript, pyret }

(There are actually more than this, one to test some details of how the Gitlab
browser-hosted VScode serves files and one distinguishing a static-file embed
vs a server embed, but logically these are the environments.)

This found real issues (detailed below).

We also added new tests in a few directions, which are unmerged due to us being
wary of dumping a ton of complexity, but both angles helped give us confidence
or directly found issues:

1. We ran *all* the public starter files from the Bootstrap curriculum through
all the environments with basic automated checks that spreadsheet data loaded
and animations opened
2. We ran all the tests through Chrome and an *old* Safari version (17) which
had previously had some regressions related to both engine issues and JS
feature issues

# Port issues found, and blind spots

The port, and the agentically-driven verifiers, had some *significant* issues
that were caught only by human review. These are useful stories, and also make
us wary that there still may be places we haven't looked that will surprise us.
We're pretty sure that the verification story above mitigates a lot of risk,
but we won't be surprised if more issues arise.

Here are a few of the categories of issues:

## Stack in the typescript compiler

The .arr compiler ran on the Pyret runtime, which has a managed stack with
heap-bounded stack space (and proper tail calls), interruptibility, and the
ability to capture and restart a continuation. Each of these interacted with
the port.

- Stack depth was an issue. Pyret's compiler has a somewhat naive recursive ANF
  pass. It is implemented for correctness, not space or speed, so it, for
  example, turns a lot of sequencing into nesting. A top-level

  x = 10 + a + b + c

  turns into a nested let t1 = 10 + a in let t2 = t1 + b in let x = t2 + c ...
  structure in ANF before being flattened out into generated code.

  This caused stack overflows for even modest programs (files below 10000
  lines), because *sequencing* could also turn into nesting. The
  Pyret-implemented compiler never noticed this because in terms of absolute
  space it was still the size of the input program. But in JavaScript, in the
  browser, the stack limit bit. It was actually critical to test in
  browser-like environments here: the original commit sidesteps this entirely
  by setting a large stack size on the node process. It wasn't until in-browser
  testing that it was clear that even reasonable curricular libraries could hit
  the shallow JS stack sizes in that environment. The fixes are straightforward
  algorithmic changes to turn recursion into iteration; further on we have some
  plans to change representations to flatten the AST out further.

- Interruptibility caused UI issues in browser contexts and in the server used
  to speed up the CLI. In the Pyret implementation, the whole compile + run
  happens on a few Pyret stacks, which are interruptible. The TypeScript
  compiler is (mostly, see below) synchronous, so there's no meaning to
  “stopping” it from within the same JS thread. This is fine from a UX
  perspective – the TS compiler is so much faster that compile times are
  negligible. However, it caused divergence issues in the state machine of the
  code.pyret.org UI for fast clicks of the "stop" button, which could leave the
  compiler running (and eventually triggering a run) while the interface
  thought it had stopped.

- The Pyret compiler simulates synchronous system calls on async JS APIs with a
  continuation-like interface. The TypeScript implementation has no such
  luxury. In fact, the arguably worst/most “following the letter not the
  intent” of the prompt code showed up here. Pyret modules can be loaded by
  URL, by file, by Google Drive API, and more, all fundamentally asynchronous.
  So the part of the compiler that fetches modules needs to be asynchronous in
  TypeScript, or so one would think.

  The code Fable wrote instead when it got to that point *started a new
  process, synchronously*, did the fetch in the new process waiting for its
  completion, and then returned the result. Simulated blocking syscalls in raw
  JS!

  https://github.com/jpolitz/pyret-lang/pull/5/commits/d27c1956c4641f43da71a334d74ed851dc74a921

  This was of course not desirable. Of course, execFileSync does not exist in
  the browser. In the browser it had papered over the issue by doing a
  pre-caching fetch of all the files *first*, partially duplicating the logic
  of the module chase in the compiler, and then having it work out that the
  exec branch was never taken because cached modules were always found first.

  This was surprising when we found it, and did the most violence to my
  confidence in the port – this wasn't a category of failure I had anticipated
  at all. We've looked and nothing this bad seems to persist, but it's quite a
  lesson. Worse, when we had an agent do review, it barely flagged it, because
  there was a detailed header comment explaining the *need* for the synchronous
  exec. My hypothesis is that this caused the strange code to be “trusted” by
  the review.

## False confidence from passing tests of the wrong artifact

Pyret's test suite has a few tests that directly import files from the compiler
with lines like:

  import file("../../../src/arr/compiler/...")

These test a *lot* of behavior, especially around compile error messages at the
CLI and the chained-environment behavior of the REPL.

These were actually what I was trying to refer to in the original prompt with
“Existing tests should work, though if you find hard constraints (e.g. tests
that test the compiler from Pyret) you can add new tests that demonstrate the
same behavior on the new version instead of dealing with FFI crossover
issues.” My intent with that sentence was to avoid issues like importing the
Pyret-implemented compiler from a TypeScript-compiled bundle of generated
code, which is just a confusing artifact to bother running.

The initial generated code did have a nod towards this: it wrote
`ts-parity-test`, which had some specific programs written to check correctness
against the Pyret implementation, but only 16 deliberately crafted programs.
As a result, for the large Pyret acceptance suite, many tests treated the .arr
compiler as a runtime library no matter who compiled them, and the agent
counted those as TS-port coverage despite the prompt warning about exactly
this.

The REPL tests were particularly problematic – the agent wrote new REPL tests
for the TypeScript implementation of the REPL, but stubbed out the part that
actually *ran* the code. So the REPL tests under the TS compiler were only
checking basic compile properties, not the full end-to-end testing a REPL suite
implies.

These were frustrating, and also contributed to mistrust. Human review was
crucial to make sure we were not resting decisions on false confidence from
running the wrong thing.

In fact, in this whole process, the bulk of human review went into writing,
designing, choosing, and checking *verifiers* rather than the code itself. This
is somewhat predictable, but these particular pitfalls seem, to me, to be a new
category of near-subversive false confidence to watch out for.

# Conclusions

We'll roll this out slowly, surely adding more tests, automated and manual, as
we go. I'm still skeptical of some of the code quality, especially around
comment blocks generated by agents. I'm not sure whether development
accelerates (more agentic contributions) or takes a reflective pause to tighten
things up from here to set up for more large efforts.

Overall, both the original compiler and the TS compiler are *more* tested than
they were before, and both certainly still have bugs. The whole exercise really
firmed up a lot of the build system and automation pipeline for running Pyret
across its now-many platforms and contexts.
```

PR #1920, opened 2026-08-26
Pause schedules: configurable initial GAS/RUNGAS for stack-capture te…

jpolitz:
…sting

Adds a --pause-schedule <file.js> compiler option to both the Pyret-hosted and TypeScript compilers. The file's contents are baked into the standalone's runtimeOptions as a module computing initialGas/initialRunGas getters; the runtime consults them at every gas-refill point. In the browser, a runtime is configured directly via setPauseSchedule -- no environment variables.

Six deterministic schedule profiles in lang/tests/pause-schedules/ (constant small/medium, seeded random small/wide, alternating, sawtooth), plus a make pause-schedule-test rule that runs the main2 suite under each profile requiring byte-equal output against the unscheduled baseline, and a byte- identity check of an .arr-compiled vs TS-compiled scheduled build.

Includes three stack-capture fixes in the stock runtime that stress-running under schedules exposed: refill on pause-resume, GAS/RUNGAS floors at pop time, and the fold entry-check re-capture path.

### DIFF
--- a/lang/.gitignore
+++ b/lang/.gitignore
@@ -35,6 +35,8 @@ build/docs
 build/phaseA
 build/phaseB
 build/phaseC
+build/phaseB-ts
+build/phaseC-ts
 build/phase1
 build/phase2
 build/phase3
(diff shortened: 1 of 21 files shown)


blerner (approved):
Codewalked together; approved.

PR #1927, opened 2026-09-02
Delete a bunch of old measurement tooling that is *very* bitrotted

jpolitz:
Shoutout to @awstlaur for making me laugh 10 years later with the contents of the required CSV file for jenkins:

  ▎ This file is intentionally left blank. Except for that. And this. And the following.
  ▎ ARE YOU HAPPY NOW, JENKINS? Look, this file exists. And it has stuff.


awstlaur:
Glad I could entertain you all these years later

PR #1929, opened 2026-09-02
Fix #1928 (error rendering issue in TS mode)

### LINKED ISSUE #1928: Inconsistent "internal errors" rendering in VScode vs CPO
The following program (reduced from a larger example that one of my TAs found):
```
use context empty-context # context is irrelevant

check:
  f(g
    is
    h)
end
```
correctly renders the expected error in CPO:

Image

but it renders an "Internal errors prevented this error from being shown" message in VSCode.

jpolitz:
A TS-internal error representation made it to the renderer. Fix it to reconstruct the right AST node at the cpo conversion layer, and add tests for related cases (errors that embed an AST node for rendering)

### DIFF
--- a/code.pyret.org/src/web/js/ts-compiler-lib.js
+++ b/code.pyret.org/src/web/js/ts-compiler-lib.js
@@ -27,11 +27,12 @@
   requires: [
     { "import-type": "builtin", name: "load-lib" },
     { "import-type": "builtin", name: "either" },
-    { "import-type": "builtin", name: "error-display" }
+    { "import-type": "builtin", name: "error-display" },
+    { "import-type": "builtin", name: "ast" }
   ],
   nativeRequires: [],
   provides: {},
-  theModule: function(runtime, namespace, uri, loadLib, eitherLib, errorDisplayLib) {
+  theModule: function(runtime, namespace, uri, loadLib, eitherLib, errorDisplayLib, astLib) {
     var gf = runtime.getField;
     var gmf = function(m, f) { return gf(gf(m, "values"), f); };
     // NOTE: runtime.ffi is not fully populated until the ffi builtin
@@ -183,6 +184,13 @@
       return e && e.name === "PyretParseError" && typeof e.kind === "string";
     }
 
+    function tsCheckOpToPyret(op) {
+      var mkOp = gmf(astLib, op.$name);
+      var l = tsLocToPyret(op.l);
+      if(typeof op.op === "string") { return mkOp.app(l, op.op); }
+      return mkOp.app(l);
+    }
+
     function throwPyretParseError(e) {
       var loc = tsLocToPyret(e.loc);
       switch(e.kind) {
@@ -197,7 +205,7 @@
         case "parse-error-bad-operator":
           return runtime.ffi.throwParseErrorBadOper(loc);
         case "parse-error-bad-check-operator":
-          return runtime.ffi.throwParseErrorBadCheckOper(tsLocToPyret(e.op && e.op.l ? e.op.l : e.loc));
+          return runtime.ffi.throwParseErrorBadCheckOper(tsCheckOpToPyret(e.op));
         case "parse-error-colon-colon":
           return runtime.ffi.throwParseErrorColonColon(loc, e.nextToken);
         case "parse-error-bad-app":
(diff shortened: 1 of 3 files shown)


code.pyret.org/src/web/js/ts-compiler-lib.js
```diff
@@ -183,6 +184,13 @@
       return e && e.name === "PyretParseError" && typeof e.kind === "string";
     }
 
+    function tsCheckOpToPyret(op) {
```
blerner, line 187: I'm not 100% sure what this is doing -- I think it's reflectively trying to find the constructor that goes with the existing `op` argument, and then constructing a new test op?  Is the actual underlying issue that line 200 below was actually a type error, that `throwParseErrorBadCheckOper` expected an AST node and was given a srcloc instead?  Then we should probably fix `https://github.com/brownplt/pyret-lang/blob/horizon/src/js/trove/ffi.js#L502 as well, to clarify its signature?

code.pyret.org/test/errors.js
```diff
@@ -249,6 +249,47 @@ describe("Rendering errors", function() {
      "load-table: h1, h2\n  source: src1\n  sanitize h1 using s1\n  sanitize h2 using s2\n  sanitize h1 using s1\nend",
      "is already sanitized by the sanitizer sanitize h1 using s1"],
 
+    ["bad-check-operator-outside-test",
+     "check:\n  f(g\n    is\n    h)\nend",
+     "must be used inside a"],
+
+    // well-formedness
+    ["single-branch-if", "if true: 5 end", "if-expression"],
```
blerner, line 257: These tests surprise me with respect to the error above -- do they _all_ trigger the same bug?  Also, is this bug unique to the new TS implementation, rather than the prior Pyret implementation?  Or is there a regression in the Pyret side as well?

jpolitz: These are new tests that don't fail, but cover similar cases, and were not tested before.

code.pyret.org/src/web/js/ts-compiler-lib.js
jpolitz: Yeah it's constructing a new test op from a bare JS value that the TS implementation had put in. Was “right” in the sense of having the right information, but not translated to the right form for the renderer.

AFAICT it's a one off (that's what all the other tests are for) – other forms that have this AST-in-error do it right.

> expected an AST node and was given a srcloc instead

No, it expected a `trove/ast.arr` AST node and was given a `ts-compiler/src/ast.ts` AST node instead.

code.pyret.org/test/errors.js
jpolitz: Sorry that answered the first question.

Is it unique to TS? **yes**, it's about where TS representations cross over to the (single copy, same-as-always) renderer in CPO that expects exactly .arr-shaped errors

There is no regression on the Pyret side.

code.pyret.org/src/web/js/ts-compiler-lib.js
blerner: ...mmm, no?  It used to call `tsLocToPyret`, which (https://github.com/jpolitz/pyret-lang/blob/058e694527ccf68d2ad52a704b5ba581b15f34be/code.pyret.org/src/web/js/ts-compiler-lib.js#L53) produced a Pyret srcloc from a TS srcloc.  But according to https://github.com/jpolitz/pyret-lang/blob/058e694527ccf68d2ad52a704b5ba581b15f34be/lang/src/arr/trove/error.arr#L2773, _this one and only one error_ expects a `s-op-*` instead of a `srcloc`.  So I think ffi.js needs to be updated to match.

jpolitz: 🤦 yeah you're right, it was passing a loc *that it extracted from the TS ast*. Two things going on, the conversion from TS rep to .arr rep *and* the confusing single case of an error in this union that takes a node not a loc.

I think ffi.js is misleading as well, yup – the Pyret side is passing a CheckOp node into a variable called `loc`

blerner: so let's fix the misleading signature in ffi.js, and then I'm on board with this fix and its tests

blerner:
should this merge to horizon, so it deploys the vscode build?

PR #1939, opened 2026-09-09
Autograder patches

jpolitz:
CC @ironm00n 

This cleans up the weird stuff I saw in my review. My comments
here on the last few commits. This adds much of @ironm00n's work from
the pyret-autograder patches, plus makes pyret-npm expose enough stuff
to run the autograder from there (modulo contexts from code.pyret.org,
which can be easily fetched separately).

- Changes caches to use "logical uris" that start from a basedir rather
  than from the absolute path, if LogicalRoot is provided, which
  currently it only is for npm-style imports. This gives future
  opportunities for shipping around cached copies of files not tied to
  e.g. /Users/joe.
- A CLI import flavor called `file-reset-load-path`, which resets the
  load path to "." (the working directory *in which the compiler is
  running*). This makes it so you can reliably "pop out" of a nested
  module to the base of the project without variable "../../..", which
  can be a huge pain to rewrite when moving student submissions around.
- An mtime logic change that trusts read-only-compiled-dirs no matter
  what their mtime is. They may not *have* associated sources, or their
  mtime may be set by a tool that installs, unzips, etc. but they are
  read-only and trusted as built, so their mtime should not cause module
  loading to look for their source. Downstream modules are still
  invalidated by source mtimes computed from the worklist, so updating a
  package's *sources* recompiles whatever imports it. Modules *without*
  sources, like builtins in lib-compiled, report mtime 0 and never
  invalidate dependents; that's been true and is still a problem (e.g.
  npm updating pyret-npm can change the statically provided information
  about lists out from under you, needs a content-hash-invalidation
  check or a version number, not for this commit)
- Improvements to npm imports to calculate paths/URIs relative to the
  npm package root, also needed for compiled caches to be reasonably
  independent of absolute paths. In particular this lets the autograder
  easily import (and reliably cache) files from the compiler bundle
  itself, for steps like checking well-formedness.

Then adds related tests for all of these, including a parity check of
the way the new URIs appear, checks of npm imports including that they
get cached correctly including if the npm package moves, and generally
that the new path calculation for `file-reset-load-path` works.


jpolitz:
(And @ironm00n I'll squash this on merge with you as author credit.)

ironm00n (approved):
Thanks! this is great!

lang/src/arr/compiler/cli-module-loader.arr
```diff
@@ -229,14 +229,18 @@ end
 
 fun get-loadable(basedir, read-only-basedirs, l, max-dep-times) -> Option<Loadable>:
   locuri = l.locator.uri()
+  # read-only dirs are immutable caches
+  fun staleness-floor(rob):
```
blerner: Who's `rob`?  (I joke, slightly; I saw "read-only _dirs_" in the comment above and expected to see `rod`, maybe, and had to read farther away to find `read-only-basedirs` to figure out `rob`.

jpolitz: Haha OK we both had the same experience, should rename.

lang/src/arr/compiler/cli-module-loader.arr
```diff
@@ -298,9 +313,24 @@ end
...
+  logical :: Option<LogicalRoot>
 }
 
+fun logical-uri-for(ctxt :: CLIContext, real-path :: String) -> Option<String>:
+  cases(Option) ctxt.logical:
+    | none => none
+    | some(lr) =>
+      prefix = lr.real + "/"
+      if string-starts-with(real-path, prefix):
+        rest = string-substring(real-path, string-length(lr.real), string-length(real-path))
```
jpolitz: Need to check if this should use path math and a relative-resolve. Strings have bitten us before here (esp. in vscode, though the VScode tests pass IDK if they hit this logic)

lang/src/arr/compiler/cli-module-loader.arr
```diff
@@ -383,11 +426,18 @@ fun module-finder(ctxt :: CLIContext, dep :: CS.Dependency):
         else:
           raise("Cannot find import " + torepr(dep))
         end
+      else if protocol == "file-reset-load-path":
```
jpolitz: Maybe this should be "project-path"? What other names might make sense. We don't love `file-reset-load-path`; it is meaningful but "reset" implies stateful when this is really just a different base.

lang/src/arr/compiler/cli-module-loader.arr
```diff
@@ -383,11 +426,18 @@ fun module-finder(ctxt :: CLIContext, dep :: CS.Dependency):
...
       else if protocol == "js-file":
         clp = ctxt.current-load-path
         real-path = get-real-path(clp, args.get(0))
         new-context = ctxt.{current-load-path: Filesystem.dirname(real-path)}
-        locator = JSF.make-jsfile-locator(real-path)
+        base = JSF.make-jsfile-locator(real-path)
+        locator = cases(Option) logical-uri-for(ctxt, real-path):
+          | none => base
+          | some(u) => base.{ method uri(self): u end }
+        end
```
jpolitz: This seems like a repeated pattern of cases on the logical-uri-for. Almost want like `wrap-in-logical-uri(locator, ctxt, real-path)`?

jpolitz:
Ben and I reviewed this today @ironm00n and we want to change `file-reset-load-path` to `project-path`. Any objection to that name?

ironm00n:
> Ben and I reviewed this today @ironm00n and we want to change `file-reset-load-path` to `project-path`. Any objection to that name?

Sure that sounds good, I didn’t put any thought into the name since I hadn’t originally intended to upstream it.

PR #1940, opened 2026-09-12
Follow v-alias when binding an included var, so `:=` works through re…

jpolitz:
…-exports

This is a clear bug, because it works at one level of import but fails when chained (because of a mistake handling `v-alias`).

Fix is just to resolve better and not fall through to “don't know, must be `vb-let`”.

Note that this *has* to work for one level for the REPL to work at all with variables, so it makes sense for it to work in general.

The dynamic semantics is fine with it – the variable is boxed.

@schanzer ran into this in a Bootstrap starter file (in production) when he refactored a variable to be one more import hop away from an assigning module.

Most PLs would use a getter/setter pair to do this, but in Pyret it's allowed, so we should support it correctly.

Claude said:

`include from M: x end` (and `use context M`, which desugars to an include) chose the binder kind straight off M's own export. When M re-exports x with `provide from D: ... end`, that export is a v-alias, which fell into the vb-let arm -- so a `var` defined in D became unassignable once reached through any re-export, and the error blamed the `var` line for being an "identifier definition".

Resolve the alias to the defining module first (value-by-uri-value already follows alias chains; `s-dot` on a module already did this). Same one-line change in the TS compiler.

Test: reexports-vars.arr re-exports defines-vars.arr's vars (via `*` and a rename); test-import-variable-reexport.arr assigns through both and checks the value from the importer, the definer, and the re-exporter. 8/8 pass with the fix; the existing one-hop test-import-variable.arr still passes 5/5. Without the fix the new test fails at name resolution.

### DIFF
--- a/lang/src/arr/compiler/resolve-scope.arr
+++ b/lang/src/arr/compiler/resolve-scope.arr
@@ -887,10 +887,17 @@ fun resolve-names(p :: A.Program, thismodule-uri :: String, initial-env :: C.Com
       | some(value-export) =>
         # A re-exported name (`provide from M: x end`) arrives here as a
         # v-alias; follow it to the defining module so a `var` stays
-        # assignable through any number of re-exports.
+        # assignable through any number of re-exports. If the target isn't
+        # in the environment (e.g. builtin://global in CPO, whose values
+        # aren't listed as provides), keep the alias and bind it as a let,
+        # which is what happened before aliases were followed at all.
         resolved-export = cases(C.ValueExport) value-export:
           | v-alias(origin, original-name) =>
-            initial-env.value-by-uri-value(origin.uri-of-definition, original-name)
+            if initial-env.all-modules.has-key-now(origin.uri-of-definition):
+              initial-env.value-by-uri(origin.uri-of-definition, original-name).or-else(value-export)
+            else:
+              value-export
+            end
           | else => value-export
         end
         vbinder = cases(C.ValueExport) resolved-export block:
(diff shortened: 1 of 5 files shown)


blerner:
This is definitely a design change -- I _know_ we thought about this a while ago and had decided not to support exporting mutable vars from one module to another.  (I think, IIRC, there was a problem with ensuring the annotations in the providing module would fire upon setting the variable from the client module?)

blerner:
(And I don't think I see tests in this PR that include trying to set `x` to some non-number value)

jpolitz:
That's a good point, and that has been a bug for a long time, and is unaffected by this commit.

image

FWIW, I absolutely disagree that we decided this – the REPL relies on using the module infrastructure for variables to be assignable in the REPL, and has for a long, long time. We can't go back on that.

blerner:
I think/thought the REPL case was slightly magical here?  I agree we're not going to deliberately break that; but I did think we'd decided against supporting it via `provide`/`import` in modules.

blerner:
That said, I'm not _against_ this change, if it isn't a regression in the semantics; I just don't want to preclude us from ever fixing that type-annotation issue

jpolitz:
The thing is, it works (and has ~always worked AFAICT for one module hop). So *not* making this change keeps the status quo badness of “when you refactor to one more module hop away, the var breaks”.

So we either regress (*no* module assignment across module boundaries), or we make it uniform. I think either way probing the annotations around `vb-var` matters. We can attach an annotation to the box at construction time, for sure, and maybe we need to look into doing that.

